PR #13, opened 2018-09-05
[models.py] Removed signal #9

### LINKED ISSUE #9: Remove signal
[Connecting signals in this way](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/models.py#L91) is not as good as doing it in the `ready` method of the `AppConfig`.

[This commit in django-freeradius](https://github.com/openwisp/django-freeradius/commit/a65f62d6139ad6b01b84215ef968f17a93b0db22) makes that line unnecessary and we can remove it.

Mah1ndra:
Connecting signals in this way is not as good as doing it in ready
method of the AppConfig.
Commits made in django-freeradius make this line unnecessary and we can
remove this.

Fixes #9

### DIFF
--- a/openwisp_radius/models.py
+++ b/openwisp_radius/models.py
@@ -1,4 +1,3 @@
-from django.contrib.auth import get_user_model
 from django.db import models
 from django.db.models import signals
 from django_freeradius.base.models import (AbstractNas, AbstractRadiusAccounting, AbstractRadiusBatch,
@@ -86,6 +85,3 @@ class RadiusUserProfile(OrgMixin, AbstractRadiusUserProfile):
     class Meta(AbstractRadiusUserProfile.Meta):
         abstract = False
         swappable = swappable_setting('openwisp_radius', 'RadiusUserProfile')
-
-
-signals.post_save.connect(set_default_limits, sender=get_user_model())


openwisp_radius/models.py
```diff
@@ -88,4 +88,4 @@ class Meta(AbstractRadiusUserProfile.Meta):
         swappable = swappable_setting('openwisp_radius', 'RadiusUserProfile')
```
rohithasrk: Unnecessary lines @Mah1ndra?

rohithasrk: Also remove the first line importing get_user_model

Mah1ndra: okay. Thank you for the review @rohithasrk

openwisp_radius/models.py
```diff
@@ -85,7 +84,3 @@ def _get_instance(self, **kwargs):
 
     class Meta(AbstractRadiusUserProfile.Meta):
         abstract = False
-        swappable = swappable_setting('openwisp_radius', 'RadiusUserProfile')
```
rohithasrk: This line was not supposed to be deleted @Mah1ndra

Mah1ndra: okay @rohithasrk . i'll update accordingly :+1:

PR #15, opened 2018-09-11
[models] Smart Migrations

### LINKED ISSUE #10: Smart migrations
Same as https://github.com/openwisp/django-freeradius/issues/41.

It would be cool to make it possible to reuse this migration here:
https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/migrations/0001_initial_squashed.py

atb00ker:
Blocked by: https://github.com/openwisp/openwisp-users/pull/30

- [x] Ensure our models reflect the default freeradius schema very closely
- [x] The first django migration should be dedicated to the default freeradius schema
- [x] Every modification and addition to the default schema, must be done in subsequent migrations, starting from 0002
- [ ] We then need to document how to skip the first migration in case users want to use django-freeradius on an existing freeradius 3 database

Closes #10


atb00ker:
@nemesisdesign 
Please review this;
Since `organization` is moved to `0002`, it was necessary to give it a default value, i gave it a default value of `uuid(00000000-0000-0000-0000-000000000000)`
Is this the correct way to go about it?
If yes,
Should i add an organisation fixture with that uuid?
If No,
Please suggest how should i proceed! :smile:

nemesisdesign:
I recognize having a default is necessary to add a field after the tables are created, but leaving such a default is not only dirty but can lead to unexpected pain in the future.

A possible cleaner approach would be:

1. create the organization table and a default organization - we need a real id though and we need to generate it in the beginning of the migration file (no hardcoding) for security reasons because we'll be using the `uuid` field as one of the fields used in the authentication
2. add the organization field which sets the default to the uuid that was generated in the previous step
3. alter all the organization fields to remove the `default`

atb00ker:
No, I'd love to continue with this issue. :blush:

nemesisdesign:
@atb00ker I'm working on integrating this PR into openwisp-radius: https://github.com/openwisp/django-freeradius/pull/184

That's going to bring many changes here so I suggest you to pause any work you're doing here.

nemesisdesign:
@atb00ker https://github.com/openwisp/django-freeradius/pull/184 is merged, rebase your PR on the current master when you can

openwisp_radius/migrations/0001_initial_freeradius.py
```diff
@@ -32,21 +18,19 @@ class Migration(migrations.Migration):
...
             options={
+                'swappable': 'OPENWISP_RADIUS_NAS_MODEL',
+                'db_table': 'nas',
                 'verbose_name_plural': 'NAS',
                 'abstract': False,
                 'verbose_name': 'NAS',
-                'swappable': 'OPENWISP_RADIUS_NAS_MODEL',
-                'db_table': 'nas',
             },
-            bases=(openwisp_users.mixins.ValidateOrgMixin, models.Model),
```
atb00ker: **Note:** `bases` is missing in all the classes and `makemigrations` didn't add it in the next file either! Should i add them manually? :blush:

openwisp_radius/migrations/0001_initial_freeradius.py
```diff
@@ -1,26 +1,12 @@
...
-import openwisp_radius.models
-import openwisp_users.mixins
-import re
-import uuid
 
 
 class Migration(migrations.Migration):
 
     initial = True
-
```
nemesisdesign: I'd add some comments here (or why not a docstring to the class) explaining this is the default freeradius schema

openwisp_radius/migrations/0002_initial_openwisp_radius.py
```diff
@@ -0,0 +1,241 @@
...
+from django.conf import settings
+from django.db import migrations, models
+from django_freeradius.migrations import add_default_group_to_existing_users, add_default_groups
+from openwisp_radius.migrations import add_default_organization
+
+import openwisp_radius.models
+import openwisp_users.mixins
+
+class Migration(migrations.Migration):
+
```
nemesisdesign: i'd add a docstring here explaining this adds the custom logic of openwisp-radius

openwisp_radius/migrations/0002_initial_openwisp_radius.py
```diff
@@ -0,0 +1,241 @@
...
+    dependencies = [
+        ('openwisp_users', '0003_default_organization'),
```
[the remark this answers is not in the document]
atb00ker: Yes,it's the 0003 migration of **openwisp_users**.
1. It was automatically generated.
2. It is required because before running this file i want the default_organisation set from the openwisp_users migrations.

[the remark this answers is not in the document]
nemesisdesign: oh right, I'm so sorry I was quick to read that as `openwisp_radius`, you should install the master of openwisp-users in `.travis.yml` then, [as we are doing for django-freeradius](https://github.com/openwisp/openwisp-radius/blob/master/.travis.yml#L16-L17)

atb00ker:
1. **Noted:** Adding more comments in general.
2. Made all the suggested changes.
3. **Note:** Tests will fail until https://github.com/openwisp/openwisp-users/pull/30 is merged.
4. **Important:** Since README is currently empty, i have not added:
For users upgrading from freeradius 3, please run:
```
./manage.py migrate --fake openwisp_radius 0001_initial_freeradius
./manage.py migrate
```

nemesisdesign:
I'll take care of the docs.

nemesisdesign:
Documentation added to django-freeradius: https://github.com/openwisp/django-freeradius/commit/8260b04ec103955193f5614ecd810050067c9a2f
We'll document the differences with openwisp-radius there directly.

PR #16, opened 2018-09-15
[api] Understand organization through automatically through token.

### LINKED ISSUE #12: API should understand organization automatically
At the moment it is necessary to change the configuration of freeradius to send the `organization`, but this doesn't scale well when using multiple organizations.

We can make openwisp-radius automatically figure out for which organization the request is coming from by doing something similar to what `openwisp-controller` does.

Here is a rough estimate of the things that need to be changed/added:

- a model similar to to [OrganizationConfigSettings](https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/models.py#L199-L227) (eg: `OrganizationRadiusSettings`) of openwisp-controller should be added, this model must store the `token` for the freeradius API and a boolean value that indicates whether the freeradius module is active for that organization (some organization may not use it and hence it doesn't make sense to keep it enabled, it also reduces the attack surface) 
- `TokenAuthentication` needs to be extended and instead of checking the settings it should query the database, should look for the uuid and token of the organization, if not found should return `403`
- since freeradius may send many requests to the API, performing the same DB query each time for authorization would slow the system down, so we should use caching to avoid this
- the cache needs to be invalidated (partially or totally, depends on how we decide to implement it) each time `OrganizationRadiusSettings` emits the signal of `post_save`

This change may require changes in [django-freeradius](https://github.com/openwisp/django-freeradius) too.

R9295:
Implements and closes #12


R9295:
@nemesisdesign any idea why that particular test in the build is failing?

nemesisdesign:
Correct, `test_save_method` is failing because of another issue, don't worry about that. If other tests would fail, that should be corrected but this is not the case now.

openwisp_radius/models.py
```diff
@@ -103,3 +106,13 @@ def _get_instance(self, **kwargs):
     class Meta(AbstractRadiusUserProfile.Meta):
         abstract = False
         swappable = swappable_setting('openwisp_radius', 'RadiusUserProfile')
+
+
+class OrganizationRadiusSettings(OrgMixin, models.Model):
+    id = models.UUIDField(default=uuid.uuid4, primary_key=True)
+    token = models.CharField(max_length=15)
+    active = models.BooleanField(default=True)
```
nemesisdesign: is this field being used? Doesn't seem so. In that case please remove it.

openwisp_radius/models.py
```diff
@@ -103,3 +106,13 @@ def _get_instance(self, **kwargs):
     class Meta(AbstractRadiusUserProfile.Meta):
         abstract = False
         swappable = swappable_setting('openwisp_radius', 'RadiusUserProfile')
+
+
+class OrganizationRadiusSettings(OrgMixin, models.Model):
+    id = models.UUIDField(default=uuid.uuid4, primary_key=True)
+    token = models.CharField(max_length=15)
```
nemesisdesign: can you add a default value which is automatically generated and a validator like we do in openwisp-controller?
https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/models.py#L218-L219
You can copy that code, then we can create some GCI dedicated issues to move that logic into openwisp-utils.

openwisp_radius/api/views.py
```diff
@@ -14,3 +55,24 @@ def _create_batch(self, serializer, **kwargs):
 
 
 batch = BatchView.as_view()
+
```
nemesisdesign: why are we not adding it to `BatchView`?

Let's try to avoid repeating the same statmements, try something like:

```python
class TokenAuthorizationMixin(object):
    authentication_classes = (TokenAuthentication,)


class AuthorizeView(TokenAuthorizationMixin, BaseAuthorizeView):
    pass

# etc.
```

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+from rest_framework.exceptions import AuthenticationFailed
 
 from openwisp_users.models import Organization
 
+from ..models import OrganizationRadiusSettings
+
+
+class TokenAuthentication(BaseTokenAuthentication):
+    def authenticate(self, request):
+        id = None
```
nemesisdesign: do you see github highlights `id`? That's because it's a built-in function.
Sometimes we are forced to use `id`, like when defining `id = UUIDField()`, but in this case we are not, so I think it's cleaner to use `uuid`, the syntax highlighting doesn't distract our mind when reading the code (remember code is read a lot more than it's written).

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+        token = None
+        if request.META.get('HTTP_AUTHORIZATION'):
+            headers = request.META.get('HTTP_AUTHORIZATION').split(',')
+            for header in headers:
+                try:
+                    id = header.split(' ')[1]
+                    token = header.split(' ')[2]
+                except IndexError:
+                    raise AuthenticationFailed('Token authentication failed')
+        elif request.GET.get('token') and request.GET.get('id'):
```
nemesisdesign: try the following: `elif 'token' in request.GET and 'id' in request.GET`

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+        if id and token:
+            if cache.get(id):
+                if cache.get(id).token != token:
+                    raise AuthenticationFailed('Token authentication failed')
+                return (AnonymousUser(), None)
+            else:
+                try:
+                    instance = OrganizationRadiusSettings.objects.get(organization=id,
+                                                                      token=token)
+                    cache.set(str(instance.pk), instance)
```
nemesisdesign: why cache the instance if we need only the token? I'd cache just the token.

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+                    raise AuthenticationFailed('Token authentication failed')
+        elif request.GET.get('token') and request.GET.get('id'):
+            id = request.GET.get('id')
+            token = request.GET.get('token')
+
+        if id and token:
+            if cache.get(id):
+                if cache.get(id).token != token:
+                    raise AuthenticationFailed('Token authentication failed')
+                return (AnonymousUser(), None)
```
nemesisdesign: `return (AnonymousUser(), None)` it's duplicated, see below

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+            else:
+                try:
+                    instance = OrganizationRadiusSettings.objects.get(organization=id,
+                                                                      token=token)
+                    cache.set(str(instance.pk), instance)
+                    return (AnonymousUser(), None)
```
nemesisdesign: `return (AnonymousUser(), None)` it's duplicated, see below

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+                except OrganizationRadiusSettings.DoesNotExist:
+                    raise AuthenticationFailed('Token authentication failed')
+
+        raise AuthenticationFailed('Token authentication failed')
```
nemesisdesign: Here I think you can do:

```python
    return (AnonymousUser(), None)
raise AuthenticationFailed('Token authentication failed')
```
This way `return (AnonymousUser(), None)` is written only once and the code will get there only if `AuthenticationFailed` was not raised so it should work.
If `uuid` and `token` are `None` the `if` block won't executed and the execution should jump to the last raise, right?

openwisp_radius/api/views.py
```diff
@@ -1,8 +1,49 @@
...
+                try:
+                    id = header.split(' ')[1]
+                    token = header.split(' ')[2]
+                except IndexError:
+                    raise AuthenticationFailed('Token authentication failed')
+        elif request.GET.get('token') and request.GET.get('id'):
+            id = request.GET.get('id')
+            token = request.GET.get('token')
+
+        if id and token:
```
nemesisdesign: alternatively here you could do:

```python
if not uuid or not token:
    raise AuthenticationFailed('Token authentication failed')

# etc.
```
This keeps the indentation level low so the code is easier to read (too many nested levels make it hard to read).

openwisp_radius/models.py
```diff
@@ -103,3 +106,13 @@ def _get_instance(self, **kwargs):
...
+    active = models.BooleanField(default=True)
+
+    def save(self, *args, **kwargs):
+        cache.delete(str(self.id))
```
nemesisdesign: you are using `instance.pk` in the rest of the code, you should do the same here for consistency.

Some points:

1. wouldn't it be better to do this after the save operation is done? eg:

```python
result = super(OrganizationRadiusSettings, self).save(*args, **kwargs)
# update cache
```
If for some reason the query fails (think about a database failure), we won't update the cache, which sound more correct to me.

2. instead of deleting, can we simply update the cache?

3. I think we need also to override the `delete` method and use `cache.delete` there

openwisp_radius/models.py
R9295: @nemesisdesign the default length of netjsonconfig's ``get_random_key`` is 32. Should we use https://github.com/django/django/blob/master/django/utils/crypto.py#L48 this instead? We can set it to 15, and I can write a new validator, or it can be set as a GCI task.

openwisp_radius/api/views.py
R9295: Some tests are failing if I add the ``Mixin`` to ``BatchView``. It seems that in django-freeradius, the View is without any ``authentication_classes`` https://github.com/openwisp/django-freeradius/blob/master/django_freeradius/api/views.py#L170

nemesisdesign: Damn.. I'm going to fix that

PR #26, opened 2018-10-03
[batch] Fix batch users not being assigned an organization.

### LINKED ISSUE #24: [batch] Batch users do not get the organization
Users created with batch operations from the admin must be created as members of the same organization  which was used to create the batch operation object, at the moment they get no organization.

It seems that this is being done for the API but the feature has to be unit tested.

R9295:
Fixes and closes #24


nemesisdesign:
Thanks @R9295 :+1:

PR #28, opened 2018-10-08
[api] Accounting API now restricts records by organization id in token

### LINKED ISSUE #19: [api] Accounting API should show only records of current organization
How to reproduce the problem:

1. set `DJANGO_FREERADIUS_EDITABLE_ACCOUNTING = True` in your `local_settings.py` (or `settings.py` if you like quick & dirty hacks)
2. ensure you have at least 2 organizations created in the system
3. create a few fake accounting records for both organizations
4. now open `http://localhost:8000/api/v1/accounting/?uuid=<uuid-of-org>&token=<token-of-org>` substituting <uuid-of-org1> and <token-of-org1> with the uuid and token of one of the two orgs

We expect to be able to see only accounting records of that org, but we see all of them.

How to fix:

- write a failing test case that replicates what you just did manually in the automated test suite (Test Driven Development is the best approach when fixing bugs)
- now edit `AccountingView`, probably defining a custom `get_queryset` method which filters by organization id would do, you can get the organization id from `self.request.auth` (read the code of [TokenAuthentication](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/views.py#L18-L55) and see also [get_queryset on DRF docs](http://www.django-rest-framework.org/api-guide/generic-views/#get_querysetself))

R9295:
Note: the ```start time``` filter is from the inherited class(tests fail otherwise as they're expected to  be sorted). Should I call ```super()``` and then filter over it instead? 
Implements and closes #19

### DIFF
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -93,7 +93,8 @@ postauth = PostAuthView.as_view()
 
 
 class AccountingView(TokenAuthorizationMixin, BaseAccountingView):
-    pass
+    def get_queryset(self):
+        return super().get_queryset().filter(organization=self.request.auth)
 
 
 accounting = AccountingView.as_view()
--- a/openwisp_radius/tests.py
+++ b/openwisp_radius/tests.py
@@ -16,7 +16,7 @@ from django_freeradius.tests.base.test_models import (BaseTestNas, BaseTestRadiu
 from django_freeradius.tests.base.test_social import BaseTestSocial
 from django_freeradius.tests.base.test_utils import BaseTestUtils
 
-from openwisp_users.models import OrganizationUser
+from openwisp_users.models import Organization, OrganizationUser
 
 from .mixins import CallCommandMixin, CreateRadiusObjectsMixin, PostParamsMixin
 from .models import (Nas, OrganizationRadiusSettings, RadiusAccounting, RadiusBatch, RadiusCheck, RadiusGroup,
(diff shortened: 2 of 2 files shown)


openwisp_radius/api/views.py
```diff
@@ -93,7 +94,8 @@ class PostAuthView(TokenAuthorizationMixin, BasePostAuthView):
 
 
 class AccountingView(TokenAuthorizationMixin, BaseAccountingView):
-    pass
+    def get_queryset(self):
+        return RadiusAccounting.objects.filter(organization=self.request.auth).order_by('-start_time')
```
nemesisdesign: yes you should be able to call super here and chain `filter` on it, eg:

```python
return super().get_queryset().filter(organization=self.request.auth)
```

openwisp_radius/tests.py
```diff
@@ -169,6 +169,30 @@ def test_register_404(self):
...
+        test1.full_clean()
+        test1.save()
+        test2 = self.radius_accounting_model(session_id='asd2',
+                                             organization=self.default_org,
+                                             nas_ip_address=nas_ip,
+                                             unique_id='1234')
+        test2.full_clean()
+        test2.save()
+        data = self.client.get(self._acct_url,
+                               HTTP_AUTHORIZATION='Bearer {0} {1}'.format(org.pk,
```
nemesisdesign: shouldn't this authorization header already be present in the test class since it's used in all other tests?

PR #30, opened 2018-10-17
[runtests.py] Added indentation

OltarzewskiK:
I have added tabulation before [this line](https://github.com/openwisp/openwisp-radius/blob/master/runtests.py#L15), because in the previous version `execute_from_command_line(args)` line wasn't in if statement.


OltarzewskiK:
@nemesisdesign Fixed

rohithasrk:
Thank you @OltarzewskiK!

PR #32, opened 2018-11-18
[ci] add ci check for migrations

OltarzewskiK:


requirements-test.txt
```diff
@@ -4,3 +4,4 @@ isort
 flake8
 responses
 freezegun
+openwisp-utils
```
nemesisdesign: `openwisp-utils>=0.2.1` as in https://github.com/openwisp/django-freeradius/commit/9647229d324c639d2449e1ead7981caf618e2fed

OltarzewskiK: Done

PR #41, opened 2018-12-09
[admin] Updated AlwaysHasChangedForm #18

### LINKED ISSUE #18: [admin] Use AlwaysHasChangedMixin from openwisp-utils
**Reserved for GCI**

This issue can be done only once https://github.com/openwisp/openwisp-controller/issues/44 is completed.

Remove  `openwisp_radius.admin.AlwaysHasChangedForm` logic in favour of `AlwaysHasChangedMixin` in openwisp-utils (instead of inheriting from `object`, inherith from `AlwaysHasChangedMixin` and remove the logic of the class).

This is done to avoid duplication with other modules that are implementing similar logic.

OltarzewskiK:
Fixes #18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
   - pip install $DJANGO
   - python setup.py -q develop
   - pip install -r requirements-test.txt
+  - pip install -e git+git://github.com/openwisp/openwisp-utils#egg=openwisp_utils
 
 before_script:
   - flake8
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 django-freeradius>=0.1.0a
 openwisp-users>=0.1.8,<0.2
-openwisp-utils>=0.2,<0.3
 django-rest-auth>=0.9.3,<0.12.0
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
 django-freeradius>=0.1.0a
-openwisp-users>=0.1.8,<0.2
-openwisp-utils>=0.2,<0.3
+openwisp-users>=0.1.8,< 0.2
```
nemesisdesign: do not add that space please

requirements.txt
```diff
@@ -1,4 +1,4 @@
...
+openwisp-utils>=0.2,< 0.3.1
```
nemesisdesign: try to remove this because it's already defined in `django-freeradius` so having it here is redundant

.travis.yml
```diff
@@ -17,6 +17,7 @@ install:
   - pip install https://github.com/openwisp/django-freeradius/tarball/master
   - # TODO: this is temporary, remove it when openwisp-users > 0.1.11 is released
   - pip install https://github.com/openwisp/openwisp-users/tarball/master
+  - pip install -e git+git://github.com/openwisp/openwisp-utils#egg=openwisp_utils
```
nemesisdesign: try moving the previous two lines after `pip install -r requirements-test.txt`

nemesisdesign: actually, the previous 3 (move the comment as well)

PR #45, opened 2019-03-27
[api] Add reset and change password endpoints #44

### LINKED ISSUE #44: [api] Add reset and change password endpoints
Add two endpoints:

- one for users who forgot their password and need to reset it
- one for users which are authenticated and need to change their password

[django-rest-auth](https://github.com/Tivix/django-rest-auth) may already provide these endpoints, in that case we can mostly reuse their base classes as we do for the registration API endpoint.

Vivekrajput20:
Closes #44

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -1,0 +1,21 @@
+from django.conf import settings
+from rest_auth.serializers import PasswordResetSerializer as BasePasswordResetSerializer
+
+
+class PasswordResetSerializer(BasePasswordResetSerializer):
+    def save(self):
+        request = self.context.get('request')
+        password_reset_url = self.context.get('password_reset_url')
+        # Set some values to trigger the send_email method.
+        opts = {
+            'use_https': request.is_secure(),
+            'from_email': getattr(settings, 'DEFAULT_FROM_EMAIL'),
+            'email_template_name': ('custom_password_reset_email.html'),
+            'request': request,
+            'extra_email_context': {
+                'password_reset_url': password_reset_url
+            }
+        }
+
+        opts.update(self.get_email_options())
+        self.reset_form.save(**opts)
--- a/openwisp_radius/api/urls.py
+++ b/openwisp_radius/api/urls.py
@@ -11,6 +11,16 @@ urlpatterns = [
     url(r'^registration/(?P<slug>[\w-]+)/$',
         views.register,
         name='rest_register'),
+    # password reset
+    url(r'^password/reset/confirm/(?P<slug>[\w-]+)/$',
+        views.password_reset_confirm,
+        name='rest_password_reset_confirm'),
+    url(r'^password/reset/(?P<slug>[\w-]+)/$',
+        views.password_reset,
+        name='rest_password_reset'),
+    url(r'^password/change/(?P<slug>[\w-]+)/$',
+        views.password_change,
+        name='rest_password_change'),
     # obtaining the user token is also different for every org
     url(r'^user-token/(?P<slug>[\w-]+)/$',
         views.obtain_auth_token,
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -1,10 +1,15 @@
 import logging
+from uuid import UUID
 
 from allauth.account import app_settings as allauth_settings
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser
+from django.contrib.auth.tokens import default_token_generator
 from django.core.cache import cache
 from django.http import Http404
+from django.utils.encoding import force_bytes, force_text
+from django.utils.http import urlsafe_base64_decode, urlsafe_base64_encode
 from django.utils.translation import ugettext_lazy as _
 from django.views.decorators.csrf import csrf_exempt
 from django_freeradius.api.views import AccountingView as BaseAccountingView
@@ -14,7 +19,11 @@ from django_freeradius.api.views import ObtainAuthTokenView as BaseObtainAuthTok
 from django_freeradius.api.views import PostAuthView as BasePostAuthView
 from rest_auth.app_settings import JWTSerializer, TokenSerializer
 from rest_auth.registration.views import RegisterView as BaseRegisterView
+from rest_auth.views import PasswordChangeView as BasePasswordChangeView
+from rest_auth.views import PasswordResetConfirmView as BasePasswordResetConfirmView
+from rest_auth.views import PasswordResetView as BasePasswordResetView
 from rest_framework.authentication import BaseAuthentication
+from rest_framework.authentication import TokenAuthentication as BaseTokenAuthentication
 from rest_framework.exceptions import AuthenticationFailed, ParseError, ValidationError
 from rest_framework.serializers import Serializer
 
@@ -24,6 +33,7 @@ from ..models import OrganizationRadiusSettings
 
 logger = logging.getLogger(__name__)
 _TOKEN_AUTH_FAILED = _('Token authentication failed')
+User = get_user_model()
 
 
 class TokenAuthentication(BaseAuthentication):
@@ -133,6 +143,13 @@ class DispatchOrgMixin(object):
             raise Http404()
         return super().dispatch(*args, **kwargs)
 
+    def validate_membership(self, user):
+        if (self.organization.pk,) not in user.organizations_pk:
+            message = _('User "{}" is not member '
+                        'of "{}"').format(user.username, self.organization.slug)
+            logger.warning(message)
+            raise ValidationError({'non_field_errors': [message]})
+
 
 class RegisterView(DispatchOrgMixin, BaseRegisterView):
     def perform_create(self, serializer):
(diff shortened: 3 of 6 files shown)


tests/settings.py
```diff
@@ -68,7 +68,9 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [],
+        'DIRS': [
+            os.path.join(os.path.dirname(BASE_DIR), 'openwisp_radius', 'templates')
```
nemesisdesign: this should not be necessary, if you have to add this here, something is wrong.
This configuration here is used to allow to customize templates at project level, while we are working at reusable-app level.

Vivekrajput20: I tried doing it without writing this here but the default template is not overwritten by any of the directory structures I tried at app level. What I think the bug is the template loaders search for the app level directories in the order INSTALLED APP. And 'django.contrib.admin' precedes openwisp_radius so it's using that template. Though I'm not so sure of this.
What should I do?

Vivekrajput20:
@nemesisdesign  I have made the requested changes. Please review it.
Thanks

tests/settings.py
```diff
@@ -161,3 +161,8 @@
     from local_settings import *
 except ImportError:
     pass
+
+REST_AUTH_SERIALIZERS = {
+    'PASSWORD_RESET_SERIALIZER':
+        'openwisp_radius.api.serializers.PasswordResetSerializer',
```
nemesisdesign: I'd write this on one line for consistency with the rest of the settings, we allow `110` characters line length if I'm not wrong

openwisp_radius/api/serializers.py
```diff
@@ -0,0 +1,17 @@
+from django.conf import settings
+from rest_auth.serializers import PasswordResetSerializer as BasePasswordResetSerializer
+
+
+class PasswordResetSerializer(BasePasswordResetSerializer):
+    def save(self):
```
nemesisdesign: can you add a test that covers this code so to avoid the coverage to decrease?

Vivekrajput20:
@nemesisdesign I have written the tests. Please review them.

openwisp_radius/api/urls.py
```diff
@@ -11,6 +11,16 @@
     url(r'^registration/(?P<slug>[\w-]+)/$',
         views.register,
         name='rest_register'),
+    # password reset
+    url(r'^password/reset/$',
```
nemesisdesign: here's the first challenge of this project, which I didn't notice before when reviewing, sorry for that.

Do you see how the other URLs have the `(?P<slug>[\w-]+)` parameter?
that's because each of these URLs has to be called by different organizations.

Since the password change and reset operations will be performed from the captive pages of different organizations, we need to include this parameter in the URL and have each view use this parameter to check whether the user belongs to the organizations or not, if not, return something like `403` HTTP status code.

openwisp_radius/templates/custom_password_reset_email.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}{% autoescape off %}
+{% blocktrans %}You're receiving this email because you requested a password reset for your user account at
+{{ site_name }}.{% endblocktrans %}
+
+{% trans "Please go to the following page and choose a new password:" %}
+{% block reset_link %}
+{# Password reset confirm url #}
+http://example.com/password/reset/{{uid}}/{{token}}
```
nemesisdesign: I didn't notice before, but this URL is hardcoded, which is not good because it cannot be edited by users and it won't work.

I guess this will be the URL of the captive page where the user will complete the password reset right?

I think this URL should be configurable with a django setting but should be different for each organization, because the captive page changes for each organization.

I would use a setting like the following:

```python
PASSWORD_RESET_URLS = {
    'default': 'https://example.com/{slug}/password/reset/{uuid}/{token}'  # fallback in case the specific org page is not defined
    # use the uuid because the slug can change
    'a8697849-6f9c-443a-b530-9dde3a910214': ''https://organization1.com/{slug}/password/reset/{uuid}/{token}',
    '463bdcc3-7bb8-479e-86dc-eb3f0c7eb50d: ''https://organization2.com/{slug}/password/reset/{uuid}/{token}'
}
```

You then retrieve the `Organization` in the view from the slug and use it to get the URL:

```python
organization = Organization.objects.get(slug=slug)
password_reset_url = settings.PASSWORD_RESET_URLS.get(str(organization.id))
if not password_reset_url:
    # always provide a default value
    password_reset_url = PASSWORD_RESET_URLS.get('default', DEFAULT_PASSWORD_RESET_URL)
```

Keep in mind this is almost pseudo-code.

openwisp_radius/api/urls.py
Vivekrajput20: Since email is unique to every user, I thought we can let users change the password without verifying the organisation. I don't think there is a need to verify the organisation for resetting the password.

Also, the email field for the users added with batch API view is empty and we don't have any API view to let users edit email or any other information. Those users can't change their password if forgotten. Should I add it to the issues?

openwisp_radius/templates/custom_password_reset_email.html
Vivekrajput20: Can't we modify the organisations model and add a password_reset_url field or some front-end url field to it. I think it will be a better way of doing it.

nemesisdesign: no, because the module that defines the `Organization` model, openwisp-users, does not know anything about this feature that is specific to the radius module and it should not know.
In the near future we can add something easier to configure like we have for the organization shared secret for openwisp controller or the organization radius API token for the radius module, we could have a model that is related to `Organization`, which stores all the configuration of a captive page for a specific org, but it's too early to do that now.

However, once we complete this, we should create the issue to do this improvement, so remind me please if I forget.

openwisp_radius/api/urls.py
nemesisdesign: you are missing the point, these URLs will be called from a captive page that is specific to an organization, so if the user is not member of that organization it makes no sense to let that user reset their password with that organization

Vivekrajput20:
@nemesisdesign  I have implemented the requested changes and have updated the tests.

tests/settings.py
```diff
@@ -161,3 +161,13 @@
     from local_settings import *
 except ImportError:
     pass
+
+REST_AUTH_SERIALIZERS = {
+    'PASSWORD_RESET_SERIALIZER': 'openwisp_radius.api.serializers.PasswordResetSerializer',
+}
+
+PASSWORD_RESET_URLS = {
+    'default': 'https://example.com',  # fallback in case the specific org page is not defined
```
nemesisdesign: The entire password reset URL of the frontend must go here, someone may want to use a custom application which has a URL that is different than the openwisp captive page app will implement.

you should use:

```python
'https://example.com/{organization}/password/reset/{id}/{token}'
```

openwisp_radius/templates/custom_password_reset_email.html
```diff
@@ -0,0 +1,16 @@
+{% load i18n %}{% autoescape off %}
+{% blocktrans %}You're receiving this email because you requested a password reset for your user account at
+{{ site_name }}.{% endblocktrans %}
+
+{% trans "Please go to the following page and choose a new password:" %}
+{% block reset_link %}
+{# Password reset confirm url #}
+{{password_reset_url}}/{{organization}}/password/reset/{{uid}}/{{token}}
```
nemesisdesign: only `{{password_reset_url}}` must go here, and it must be calculated in the view

openwisp_radius/api/views.py
```diff
@@ -172,3 +178,52 @@ def get_user(self, serializer, *args, **kwargs):
...
+
+password_change = PasswordChangeView.as_view()
+
+
+class PasswordResetView(BasePasswordResetView):
+
+    def get_serializer_context(self):
+        organization = Organization.objects.get(slug=self.kwargs['slug'])
+        default_url = settings.PASSWORD_RESET_URLS.get('default')
+        password_reset_url = settings.PASSWORD_RESET_URLS.get(str(organization.pk), default_url)
```
nemesisdesign: Here we need something like:
```python
password_reset_url = password_reset_url.format(organization=organization.slug,
                                               uid=<uid here>,
                                               token=<token here>)
```

openwisp_radius/templates/custom_password_reset_email.html
```diff
@@ -0,0 +1,16 @@
...
+{% endblock %}
+{% trans "Your username, in case you've forgotten:" %} {{ user.get_username }}
+
+{% trans "Thanks for using our site!" %}
+
+{% blocktrans %}The {{ site_name }} team{% endblocktrans %}
+
+{% endautoescape %}
```
nemesisdesign: add a final new line here please

openwisp_radius/api/urls.py
```diff
@@ -11,6 +11,16 @@
     url(r'^registration/(?P<slug>[\w-]+)/$',
         views.register,
         name='rest_register'),
+    # password reset
+    url(r'^password/reset/confirm/$',
```
nemesisdesign: we need to implement the same functionality which checks the org here, remember: these URLs are used by captive pages that are specific to each organization, it's like they are totally separate instances.

In some other URLs of the API we don't have the `slug` parameter because the organization data is sent in the authorization header with a Bearer token, eg: `Authorization: Bearer {uuid} {token}`, all API URLs have a way to separate the different organizations using the system.

openwisp_radius/api/urls.py
```diff
@@ -11,6 +11,16 @@
...
+        views.password_reset_confirm,
+        name='rest_password_reset_confirm'),
+    url(r'^password/reset/(?P<slug>[\w-]+)/$',
+        views.password_reset,
+        name='rest_password_reset'),
+    url(r'^password/change/$',
```
nemesisdesign: we need the same here

openwisp_radius/api/views.py
```diff
@@ -172,3 +178,52 @@ def get_user(self, serializer, *args, **kwargs):
...
+                try:
+                    user = User.objects.get(email=email)
+                    if (organization.pk,) not in user.organizations_pk:
+                        message = _('Email "{}" is not member '
+                                    'of "{}"').format(email, kwargs['slug'])
+                        logger.warning(message)
+                        return HttpResponseForbidden(message)
+                except User.DoesNotExist:
+                    raise Http404()
+        except Organization.DoesNotExist:
```
nemesisdesign: there's some duplication with `DispatchOrgMixin`, try to reuse that as much as possible,
also try to avoid duplication of:

```python
if (self.organization.pk,) not in user.organizations_pk:
    message = _('User "{}" is not member '
                'of "{}"').format(user.username, kwargs['slug'])
    logger.warning(message)
```

Better create a method in `DispatchOrgMixin` that does that operation and reuse it.
Duplication is evil because it adds maintenance overhead on the project.
Remember that software has to be maintained and updated once is written, the more duplication we have, the more code we will have to maintain and update.

Vivekrajput20:
@nemesisdesign I have resolved all the mentioned issues. Please review it now.

PR #47, opened 2019-04-05
[settings] Enable CORS in django rest framework for cross-domain ajax requests

### LINKED ISSUE #46: Enable CORS for cross-domain ajax requests
Enable CORS in django rest framework for cross-domain ajax requests from the frontend Apps

Vivekrajput20:
closes #46

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -172,3 +172,16 @@ REST_AUTH_SERIALIZERS = {
 #     # use the uuid because the slug can change
 #     # 'dabbd57a-11ca-4277-8dbb-ad21057b5ecd': 'https://organization.com/{organization}/password/reset/{uid}/{token}',
 # }
+
+try:
+    import corsheaders
+
+    INSTALLED_APPS[:0] = ['corsheaders']
+
+    MIDDLEWARE[:0] = ['corsheaders.middleware.CorsMiddleware', 'django.middleware.common.CommonMiddleware']
+
+    # WARNING: for development only!
+    CORS_ORIGIN_ALLOW_ALL = True
+
+except ImportError:
+    pass
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -4,3 +4,4 @@ responses
 freezegun
 openwisp-utils[qa]>=0.2.1
 djangorestframework-jwt
+django-cors-headers>=2.5.2


requirements.txt
```diff
@@ -1,3 +1,4 @@
 django-freeradius>=0.1.0a
 openwisp-users>=0.1.8,<0.2
 django-rest-auth>=0.9.3,<0.12.0
+django-cors-headers>=2.5.2
```
nemesisdesign: this line makes `django-cors-headers` a dependency of `openwisp-radius` and will automatically install it when a developer using `openwisp-radius` will install this openwisp module.

But the user could set CORS in other ways, for example using Nginx. This is a configuration that is performed on project level, not on app-level (openwisp-radius is a django app, see the concept of `INSTALLED_APPS`, not a django project).

Instead of adding this here, install it in `.travis.yml` after `python setup.py install` is called.

tests/settings.py
```diff
@@ -172,3 +175,8 @@
 #     # use the uuid because the slug can change
 #     # 'dabbd57a-11ca-4277-8dbb-ad21057b5ecd': 'https://organization.com/{organization}/password/reset/{uid}/{token}',
 # }
+
```
nemesisdesign: remove the previous lines and do something like:

```
try:
    import corsheaders

    INSTALLED_APPS.append('corsheaders')

    MIDDLEWARE += ['......etc.....', '..........']

    CORS_ORIGIN_WHITELIST = () # whitelist everything here, will be valid only for development, but leave a comment explaining to use that kind of setting only during development
except ImportError:
    pass

**Comment by Vivekrajput20** [comment]
@nemesisdesign please review it.
thanks

File: tests/settings.py [hunk 3]
```diff
@@ -166,9 +166,17 @@
...
+    import corsheaders
+
+    INSTALLED_APPS[:0] = ['corsheaders']
+
+    MIDDLEWARE[:0] = ['corsheaders.middleware.CorsMiddleware', 'django.middleware.common.CommonMiddleware']
+
+    # WARNING: for development only!
+    CORS_ORIGIN_WHITELIST = (
+        # 'example.com',
+        # 'localhost:3000',
```
nemesisdesign: whitelist everything during development using `CORS_ORIGIN_ALLOW_ALL = True`

tests/settings.py
```diff
@@ -166,9 +166,17 @@
     'PASSWORD_RESET_SERIALIZER': 'openwisp_radius.api.serializers.PasswordResetSerializer',
 }
 
-# OPENWISP_RADIUS_PASSWORD_RESET_URLS = {
-#     # fallback in case the specific org page is not defined
-#     'default': 'https://example.com/{organization}/password/reset/{uid}/{token}',
-#     # use the uuid because the slug can change
-#     # 'dabbd57a-11ca-4277-8dbb-ad21057b5ecd': 'https://organization.com/{organization}/password/reset/{uid}/{token}',
-# }
```
nemesisdesign: do not remove these comments please

Vivekrajput20:
@nemesisdesign  please review it.

nemesisdesign:
Thanks :+1:

PR #56, opened 2019-07-14
[bug] Fix broken tests on current master

Vivekrajput20:


Vivekrajput20:
@nemesisdesign  Now the build most probably won't fail after the changes in the django_freeradius are merged. I also added the runflake8 and runisort file to ease the testing during development.
Vivek

PR #58, opened 2019-07-19
[api] add support for validate auth token api #57

### LINKED ISSUE #57: [api] Add support for validate auth token api.

Vivekrajput20:
Closes #57

### DIFF
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -5,7 +5,8 @@ from django.test import override_settings
 from django.urls import reverse
 from django.utils.encoding import force_bytes, force_text
 from django.utils.http import urlsafe_base64_encode
-from django_freeradius.tests.base.test_api import BaseTestApi, BaseTestApiReject, BaseTestApiUserToken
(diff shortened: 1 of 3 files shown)


openwisp_radius/tests/test_api.py
```diff
@@ -335,3 +336,22 @@ def test_user_auth_token_404(self):
...
+
+    def _get_url(self):
+        return reverse('freeradius:validate_auth_token',
+                       args=[self.default_org.slug])
+
+    def get_user(self):
+        user = self.user_model.objects.create_user(
+            username='test_name',
+            password='test_password',
+        )
```
nemesisdesign: since `get_user` is already defined in `BaseTestApiValidateToken`, you should be able to avoid redefining this call to create the user by calling:
```python
user = super().get_user()
```

Please try this and let me know.

Vivekrajput20: @nemesisdesign it throwing this error:
```
UNIQUE constraint failed: openwisp_users_organizationuser.user_id, openwisp_users_organizationuser.organization_id
```

nemesisdesign: ok

PR #60, opened 2019-12-03
[Travis] Fixed warnings

kidzmyujikku:
adding os : linux
delete sudo : false
delete contains whitespace in line 5


kidzmyujikku:
I have updated it, is it correct?

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
-sudo: false
 cache: pip
-
+os: linux
```
atb00ker: I checked this is only appearing as `info` not as `warning` so I think we can reverse it for now. 
(The thing is that these things keep changing, so it's not worth our time to keep the `info` part changing, only when it appears as a `warning` that we need to do something about it! :smile: )

atb00ker:
I checked this is only appearing as info not as warning so I think we can reverse it for now.
(The thing is that these things keep changing, so it's not worth our time to keep the info part changing, only when it appears as a warning that we need to do something about it! smile )

atb00ker:
Please avoid opening multiple PRs for one task. Fixing the old PR is a good way to learn git, ofcourse if fixing the PR by yourself is hard, you can ask for help!

PR #63, opened 2019-12-04
[Travis] Fixed Warning

kidzmyujikku:
fixed this warning
[warn] on root: deprecated key: "sudo" (The key `sudo` has no effect anymore.)


atb00ker:
Fixed with #68

PR #66, opened 2019-12-09
[Travis] Fix Warning on travis.yml

rafiudd:
Fixing this  : 
- [warn] on root: deprecated key: "sudo" (The key `sudo` has no effect anymore.)
- [info] on root: missing os, using the default "linux"

### DIFF
--- a/openwisp_radius/api/urls.py
+++ b/openwisp_radius/api/urls.py
@@ -25,4 +25,7 @@ urlpatterns = [
     url(r'^(?P<slug>[\w-]+)/account/token/$',
         views.obtain_auth_token,
         name='user_auth_token'),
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
-sudo: false
 cache: pip
+os : linux
```
atb00ker: This change is not required. Please reverse it.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: python
-sudo: false
 cache: pip
-
+
```
atb00ker: There is still this white space that's not required. Please remove this white space change as well.
(Usually don't make whitespace changes!)

rafiudd: Okay

PR #67, opened 2019-12-10
Fix Travis Warning on travis.yml

ibnufachrizal:
Fixing this :

[warn] on root: deprecated key: "sudo" (The key sudo has no effect anymore.)
[info] on root: missing os, using the default "linux"


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
-sudo: false
 cache: pip
+os : linux
```
atb00ker: This change is not required, please reverse it!

PR #78, opened 2020-01-19
[bug] Radius token for initial default org not created #71

### LINKED ISSUE #71: Radius token for initial default org not created
It seems to me that the `OrganizationRadiusSettings` object (and relative secret token) is not being created for the default organizaiton.

Write a test which fails if the default org does not have a related `OrganizationRadiusSettings` created.

Let the test fail, then add the logic to create it in the `add_default_organization` function (which is called during the first DB migration).

Weirdo914:
Auto create OrganizationRadiusSettings object for the default organizaiton.
Fixes #71

### DIFF
--- a/openwisp_radius/migrations/__init__.py
+++ b/openwisp_radius/migrations/__init__.py
@@ -18,6 +18,8 @@ def add_default_organization(apps, schema_editor):
         for record in Model.objects.all().iterator():
             record.organization_id = settings._OPENWISP_DEFAULT_ORG_UUID
             record.save()
+    OrganizationRadiusSettings = apps.get_model('openwisp_radius', 'organizationradiussettings')
+    OrganizationRadiusSettings.objects.create(organization_id=settings._OPENWISP_DEFAULT_ORG_UUID)
 
 
 def add_default_groups(apps, schema_editor):
--- a/openwisp_radius/tests/tests.py
+++ b/openwisp_radius/tests/tests.py
@@ -618,7 +618,7 @@ class TestOgranizationRadiusSettings(ApiTokenMixin, BaseTestCase):
     user_model = User
 
     def setUp(self):
-        self.org = self._create_org()
+        self.org = self._create_org(name='test', slug='test')
 
     def test_string_representation(self):
         rad = OrganizationRadiusSettings.objects.create(organization=self.org)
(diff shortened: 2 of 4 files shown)


openwisp_radius/tests/tests.py
```diff
@@ -618,7 +618,7 @@ class TestOgranizationRadiusSettings(ApiTokenMixin, BaseTestCase):
     user_model = User
 
     def setUp(self):
-        self.org = self._create_org()
+        self.org = self._create_org(name='test', slug='test')
```
atb00ker: Why do we need to pass these arguements now ? :confused:

openwisp_radius/migrations/__init__.py
```diff
@@ -18,6 +18,10 @@ def add_default_organization(apps, schema_editor):
         for record in Model.objects.all().iterator():
             record.organization_id = settings._OPENWISP_DEFAULT_ORG_UUID
             record.save()
+    Organization = apps.get_model('openwisp_users', 'Organization')
+    OrganizationRadiusSettings = apps.get_model('openwisp_radius', 'organizationradiussettings')
+    org = Organization.objects.get(id=settings._OPENWISP_DEFAULT_ORG_UUID)
```
atb00ker: Instead of using `UUID` can you please generate and use a random token, I just think exposing `UUID` to freeradius may come back to bite us. What do you think? :smile:

openwisp_radius/tests/tests.py
Weirdo914: The function returns the default organisation if no arguments are passed. In the test, new `OrganizationRadiusSettings` are created for this organisation. As now the settings will be already created, it would cause errors.

openwisp_radius/migrations/__init__.py
Weirdo914: I think you got something wrong there, this line is used to get the default organisation by the uuid. A random token is generated as a token is not provided to the  `OrganizationRadiusSettings` object below this line and it by default generates a random token.

atb00ker: Yes, you are correct. Sorry about that.

openwisp_radius/tests/tests.py
```diff
@@ -675,6 +675,10 @@ def test_uuid_in_cache(self):
         self.assertEqual(r.status_code, 200)
         cache.clear()
 
+    def test_default_organisation_radius_settings(self):
+        org = Organization.objects.get(slug='default')
+        self.assertTrue(hasattr(org, 'radius_settings'))
```
nemesisdesign: Could you please add the following line?

```python
self.assertTrue(isinstance(org.radius_settings, OrganizationRadiusSettings))
```

PR #80, opened 2020-02-05
[deps] Added support for Django 3.0

### LINKED ISSUE #79: [deps] Update dependencies (django 3.0)
Add support for django 3.0 and update dependencies.

Drop support for python 3.5, add 3.6 and 3.7 to the travis build.

TheOneAboveAllTitan:
Added support for Django 3.0
Bumped openwsip-user to 0.2.0
Fixes #79

### DIFF
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -41,7 +41,7 @@ class TestApi(ApiTokenMixin, BaseTestApi, BaseTestCase):
         data = self._get_accounting_params(**data)
         response = self.post_json(data)
         self.assertEqual(response.status_code, 403)
-        self.assertIn('setting the organization', response.data['detail'])
+        self.assertIn('setting the organization', str(response.data['detail']))
 
     def test_register_201(self):
         self.assertEqual(User.objects.count(), 0)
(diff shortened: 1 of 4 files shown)


openwisp_radius/tests/test_api.py
```diff
@@ -180,7 +180,7 @@ def test_api_password_change(self):
         }
         response = client.post(password_change_url, data=new_password_payload)
         self.assertEqual(response.status_code, 400)
-        self.assertIn("The two password fields didn't match.", response.data['new_password2'])
+        self.assertIn("The two password fields didn’t match.", response.data['new_password2'])
```
nemesisdesign: try `str(response.data['new_password2'])`

TheOneAboveAllTitan: This one is passing on my machine 
`test_radius_group_delete_default_by_non_superuser` is the one which is failing after installing `django-freeradius` again. Local environment was not updated with GitHub repository.

TheOneAboveAllTitan: In last line, shouldn't it be `confirm_response` instead of `response` ?

TheOneAboveAllTitan: Django 2.2 uses `'` for contractions while Django 3 uses `’` for same. This is causing test to fail.

TheOneAboveAllTitan:
Have a look here https://github.com/openwisp/django-freeradius/issues/271

requirements.txt
```diff
@@ -1,3 +1,3 @@
 django-freeradius<0.2.0
-openwisp-users>=0.1.12,<0.2
+openwisp-users>=0.1.12,<0.2.1
```
nemesisdesign: This should have been `openwisp-users>=0.1.12,<0.2.1` but I will fix it.

PR #84, opened 2020-04-30
[qa] Reformatted code with black

atb00ker:


requirements.txt
```diff
@@ -2,4 +2,5 @@ django-freeradius<0.2.0
 openwisp-users>=0.1.12,<0.3.0
 django-rest-auth>=0.9.3,<0.12.0
 django-sendsms>=0.4.0,<0.5.0
+mock>=2.0.0
```
nemesisdesign, line 5: why this?

atb00ker: Unintended change, Need to investigate how it got there.

atb00ker: Fixed!

PR #100, opened 2020-06-05
[chores/gsoc2020] clean-up #90

### LINKED ISSUE #90: [gsoc20] Clean up
- https://github.com/openwisp/openwisp-radius/pull/89#pullrequestreview-424041876
- https://github.com/openwisp/openwisp-radius/pull/89#pullrequestreview-424057621

atb00ker:
Closes #90

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,12 +29,15 @@ openwisp-radius
    :target: https://pypi.org/project/black/
    :alt: code style: black
 
-.. image:: docs/source/images/demo_radius.gif
+.. image:: https://github.com/openwisp/openwisp-radius/raw/master/docs/source/images/demo_radius.gif
    :alt: Feature Highlights
 
-------------
 
-**openwisp-radius** is Django reusable app that provides an admin interface to a `freeradius <http://freeradius.org/>`_ database.
+**openwisp-radius** is Django reusable app that provides an admin interface to a
+`freeradius <http://freeradius.org/>`_ database.
+
+.. image:: https://raw.githubusercontent.com/openwisp/openwisp2-docs/master/assets/design/openwisp-logo-black.svg
+   :target: http://openwisp.org
 
 `Documentation <http://openwisp-radius.readthedocs.io/en/latest/>`_ |
 `Change log <https://github.com/openwisp/openwisp-radius/blob/master/CHANGES.rst>`_ |
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -2,25 +2,42 @@
 openwisp-radius
 ===============
 
-.. image:: https://travis-ci.org/openwisp/openwisp-radius.svg
+.. image:: https://travis-ci.org/openwisp/openwisp-radius.svg?branch=master
    :target: https://travis-ci.org/openwisp/openwisp-radius
-   :alt: Automated Build
+   :alt: CI build status
 
-.. image:: https://coveralls.io/repos/openwisp/openwisp-radius/badge.svg
-   :target: https://coveralls.io/r/openwisp/openwisp-radius
-   :alt: Coverage
+.. image:: https://coveralls.io/repos/github/openwisp/openwisp-radius/badge.svg?branch=master
+   :target: https://coveralls.io/github/openwisp/openwisp-radius?branch=master
+   :alt: Test Coverage
 
 .. image:: https://requires.io/github/openwisp/openwisp-radius/requirements.svg?branch=master
    :target: https://requires.io/github/openwisp/openwisp-radius/requirements/?branch=master
    :alt: Requirements Status
 
-.. image:: https://img.shields.io/pypi/v/openwisp-radius
-   :target: https://pypi.org/project/openwisp-radius
-   :alt: PyPI
+.. image:: https://img.shields.io/gitter/room/nwjs/nw.js.svg
+   :target: https://gitter.im/openwisp/general
+   :alt: Chat
 
-openwisp-radius is part of the `OpenWISP project <http://openwisp.org>`_.
+.. image:: https://badge.fury.io/py/openwisp-radius.svg
+   :target: http://badge.fury.io/py/openwisp-radius
+   :alt: Pypi Version
 
-.. image:: http://netjsonconfig.openwisp.org/en/latest/_images/openwisp.org.svg
+.. image:: https://pepy.tech/badge/openwisp-radius
+   :target: https://pepy.tech/project/openwisp-radius
+   :alt: Downloads
+
+.. image:: https://img.shields.io/badge/code%20style-black-000000.svg
+   :target: https://pypi.org/project/black/
+   :alt: code style: black
+
+.. image:: images/demo_radius.gif
+   :alt: Feature Highlights
+
+
+**openwisp-radius** is Django reusable app that provides an admin interface to a
+`freeradius <http://freeradius.org/>`_ database.
+
+.. image:: https://raw.githubusercontent.com/openwisp/openwisp2-docs/master/assets/design/openwisp-logo-black.svg
    :target: http://openwisp.org
 
 .. toctree::
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -860,7 +860,7 @@ class AbstractRadiusBatch(OrgMixin, TimeStampedEditableModel):
             )
         if self.strategy == 'csv':
             validate_csvfile(self.csvfile.file)
-        super(AbstractRadiusBatch, self).clean()
+        super().clean()
 
     def add(self, reader, password_length=BATCH_DEFAULT_PASSWORD_LENGTH):
         users_list = []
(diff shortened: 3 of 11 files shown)


openwisp_radius/tests/test_admin.py
```diff
@@ -220,7 +220,10 @@ def test_radiuscheck_change(self):
 
     def test_radiusbatch_change(self):
         obj = self._create_radius_batch(
-            name='test', strategy='prefix', prefix='test', expiration_date='1998-01-28'
+            name='test',
+            strategy='prefix',
+            prefix='test-prefix4',
```
atb00ker, line 225: Change 1: I had to change the `prefix` for all the tests so they have unique prefixes and unique files created because for finding which file was created from which testcase, I could only perform on hit-and-trail which was a waste of time.

openwisp_radius/tests/test_admin.py
```diff
@@ -363,9 +366,12 @@ def test_radius_batch_save_model(self):
         response = self.client.post(add_url, data, follow=True)
         error_message = 'Ensure this value is greater than or equal to 1'
         self.assertTrue(error_message in str(response.content))
+        self._delete_file('test-prefix12.pdf')
```
atb00ker, line 369: I had to manually delete the file `test-prefix12.pdf` because `radbatch.pdf.path` contains `test-prefix12_<random-string>.pdf` yet this file exists as well as the one with the random string, is this intended behavior or a bug?

openwisp_radius/tests/test_commands.py
```diff
@@ -140,14 +140,21 @@ def test_prefix_add_users_commnad(self):
...
+        self._delete_file('test-prefix8.pdf')
+        options = dict(
+            organization=self.default_org, prefix='test-prefix7', n=-5, name='test2'
+        )
+        try:
             self._call_command('prefix_add_users', **options)
+        except SystemExit:
+            self._delete_file('test-prefix7.pdf')
+        else:
+            self.fail('SystemError not raised!')
```
atb00ker, line 160: Even when _call_command fails and the radiusbatch is not created, the file is created. Is this intended behavior?

openwisp_radius/tests/mixins.py
```diff
@@ -65,6 +70,24 @@ def setUp(self):
...
+            os.mkdir(settings.MEDIA_ROOT)
+
+    def tearDown(self):
+        for radbatch in RadiusBatch.objects.all():
+            # Delete associated pdf
+            if radbatch.pdf and os.path.isfile(radbatch.pdf.path):
+                os.remove(radbatch.pdf.path)
+            # Delete associated csv
+            if radbatch.csvfile and os.path.isfile(radbatch.csvfile.path):
+                os.remove(radbatch.csvfile.path)
```
nemesisdesign, line 86: why not do this in the delete method of the model and have tearDown call radbatch.delete()?

docs/source/index.rst
```diff
@@ -2,23 +2,39 @@
...
+   :target: https://pypi.org/project/black/
+   :alt: code style: black
+
+.. image:: images/demo_radius.gif
+   :alt: Feature Highlights
+
+**openwisp-radius** is Django reusable app that provides an admin interface to a
+`freeradius <http://freeradius.org/>`_ database.
 
 .. image:: http://netjsonconfig.openwisp.org/en/latest/_images/openwisp.org.svg
```
nemesisdesign, line 39: can you switch to this logo please? https://raw.githubusercontent.com/openwisp/openwisp2-docs/master/assets/design/openwisp-logo-black.svg

README.rst
```diff
@@ -29,12 +29,14 @@ openwisp-radius
...
+.. image:: https://github.com/openwisp/openwisp-radius/raw/master/docs/source/images/demo_radius.gif
    :alt: Feature Highlights
 
-------------
+**openwisp-radius** is Django reusable app that provides an admin interface to a
+`freeradius <http://freeradius.org/>`_ database.
 
-**openwisp-radius** is Django reusable app that provides an admin interface to a `freeradius <http://freeradius.org/>`_ database.
+.. image:: http://netjsonconfig.openwisp.org/en/latest/_images/openwisp.org.svg
+   :target: http://openwisp.org
```
nemesisdesign, line 39: can you switch to this logo please? https://raw.githubusercontent.com/openwisp/openwisp2-docs/master/assets/design/openwisp-logo-black.svg

README.rst
```diff
@@ -29,12 +29,14 @@ openwisp-radius
    :target: https://pypi.org/project/black/
    :alt: code style: black
 
-.. image:: docs/source/images/demo_radius.gif
+.. image:: https://github.com/openwisp/openwisp-radius/raw/master/docs/source/images/demo_radius.gif
    :alt: Feature Highlights
 
-------------
```
nemesisdesign, line 35: Leave blank line for consistency pls

openwisp_radius/tests/mixins.py
atb00ker: Great, Idea, I didn't have to do much to implement it! :smile:

openwisp_radius/tests/test_commands.py
nemesisdesign: I don't think so, looks like a bug to me.

nemesisdesign: Look here: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/base/models.py#L895-L905

`generate_pdf` writes to disk before validation, so if the validation fails, the file is left there anyway. Can you help to fix this please? We could do something like:

```python
try:
    batch.full_clean()
except ValidationError:
    delete_file()  # (pseudo code)
    raise e
```
I also created an issue for the GCI to address the broader issue that we should not save this info in a file on disk, I thought about a way to avoid this that should work: #102. 
Shame it will be a backward incompatible change, but I don't think it's a huge issue.

atb00ker: @nemesisdesign I've taken a different approach to solve it than in the comment:

In [`generate_pdf` I changed the path to `/tmp/{prefix}.pdf`](https://github.com/openwisp/openwisp-radius/pull/100/files#diff-d4dfe472aa734b022012d64055559459R162 ).

So that the file created before save goes to `/tmp/` and os cleans it later while the [final file saved by django goes in `{settings.MEDIA_ROOT}/{prefix}.pdf`](https://github.com/openwisp/openwisp-radius/pull/100/files#diff-141584df2cf4ec556f1b9fe45e78d246R902)

This also takes care of the problem that 2 files are create: one by `generate_pdf()` and another by django with a random name which is saved with the model.
Now, the name of the pdf file and radiusbatch name can be the same.

What do you think about this approach?

openwisp_radius/base/models.py
```diff
@@ -860,7 +860,7 @@ def clean(self):
             )
         if self.strategy == 'csv':
             validate_csvfile(self.csvfile.file)
-        super(AbstractRadiusBatch, self).clean()
+        super().clean()
```
atb00ker, line 863: This is a random chore, cleanup, not related to the issue but just saw it!

openwisp_radius/tests/test_commands.py
atb00ker: P.S: I'll work on #103 as soon as possible! :smile:

openwisp_radius/base/models.py
nemesisdesign: thanks :+1:

openwisp_radius/tests/test_commands.py
nemesisdesign: > This also takes care of the problem that 2 files are create: one by `generate_pdf()` and another by django with a random name which is saved with the model.
> Now, the name of the pdf file and radiusbatch name can be the same.
> 
> What do you think about this approach?

Is good thanks!

> P.S: I'll work on #103 as soon as possible!

Did you mistype? #103 is related to openwisp-notifications.

openwisp_radius/base/models.py
```diff
@@ -899,7 +899,7 @@ def prefix_add(self, prefix, n, password_length=BATCH_DEFAULT_PASSWORD_LENGTH):
             user.full_clean()
             self.save_user(user)
         pdf_file = generate_pdf(prefix, {'users': user_password})
-        pdf_file.name = pdf_file.name.split('/')[-1]
+        pdf_file.name = f'{settings.MEDIA_ROOT}/{prefix}.pdf'
```
nemesisdesign, line 902: I have tested this now, I get a 404 if I try to open the link to the file in the admin, so there's more work to do here.
I would go back to the idea of deleting the file if validation goes wrong.

Another issue that I found is that creating the file in `/tmp` may not spot if there's already a file with the same name.

openwisp_radius/tests/test_commands.py
atb00ker: Yeah, I meant #102

openwisp_radius/base/models.py
atb00ker: > I have tested this now, I get a 404 if I try to open the link to the file in the admin, so there's more work to do here.

Fixed it.

> Another issue that I found is that creating the file in /tmp may not spot if there's already a file with the same name.

It doesn't matter if the old file is overridden does it?
In anycase, I am already working on #102 so I suppose it shouldn't matter soon! :smile:

nemesisdesign: Ok let's move on, #102 is the best solution

PR #106, opened 2020-06-07
[radius] Allow customizing User model #87

### LINKED ISSUE #87: [radius] Allow customizing User model
Depends on https://github.com/openwisp/openwisp-users/issues/122.

Once swappable models are available in openwisp-users, we should update openwisp-radius to make sure it imports the models of openwisp-radius using the swappable methods and we should provide a documentation that explains how to extend the user model.

We should make sure that when the user model is extended, the new fields are automatically included in its API.

atb00ker:
TODO: Change the link of the `openwisp-users` installation.
Closes #87


.travis.yml
```diff
@@ -18,6 +18,8 @@ install:
   - python setup.py -q develop
   - pip install -r requirements-test.txt
   - npm install -g jslint
+  # TODO: Remove when openwisp-users PR #123 is merged
+  - pip install --upgrade https://github.com/atb00ker/openwisp-users/tarball/swapper/
```
atb00ker, line 22: Remove this change before merging, only for testing.

tests/openwisp2/sample_users/models.py
```diff
@@ -0,0 +1,38 @@
...
+    AbstractUser,
+    BaseGroup,
+    BaseOrganization,
+    BaseOrganizationOwner,
+    BaseOrganizationUser,
+)
+
+
+class User(AbstractUser):
+    fiscal_number = models.CharField(max_length=16, null=True, blank=True)
```
nemesisdesign, line 19: use `social_security_number` please

tests/openwisp2/settings.py
```diff
@@ -165,6 +172,14 @@
     OPENWISP_RADIUS_ORGANIZATIONRADIUSSETTINGS_MODEL = (
         'sample_radius.OrganizationRadiusSettings'
     )
+    # Rename sample_app database
+    DATABASES['default']['NAME'] = os.path.join(BASE_DIR, 'sample_radius.db')
```
atb00ker, line 176: If you do 
```
SAMPLE_APP=1 python tests/manage.py migrate
./run-qa-checks
```
It fails because Migration order is disturbed, to avoid that, I've changed the name of the database! :smile:

.travis.yml
nemesisdesign: changed to master

PR #107, opened 2020-06-10
[models] don't save PDF to disk #102

### LINKED ISSUE #102: [radius/models] AbstractRadiusBatch: don't save PDF to disk
Saving the PDF to disk is not a good idea.
Since all the information generated during the batch is text, we could store a JSON string and generate the PDF on the fly.

The steps would be more or less the following:

- add a textfield/jsonfield
- when the username/passwords are generated, write them there (a list of dicts)
- add a model method to generate the PDF (write to in-memory `io.Bytes` object and return that)
- write an API view that serves the PDF, checks the user is either superuser or  is_staff, is member of the organization and has view permissions on `RadiusBatch` , the API view shall support `SessionAuthentication` and `UserTokenAuthentication`
- replace the pdf file field in the admin with a link to the view to download the PDF file
- replace the pdf file field in the API/serializer with a link the view to download the PDF file
- this is a backward incompatible change, add a note in the change log (`CHANGES.rst`)

atb00ker:
Closes #102


openwisp_radius/api/views.py
```diff
@@ -678,3 +675,39 @@ def create_phone_token(self, *args, **kwargs):
...
+
+
+class DownloadRadiusBatchPdfView(DispatchOrgMixin, APIView):
+    authentication_classes = (UserTokenAuthentication, SessionAuthentication)
+    permission_classes = (IsAllowedRadiusBatch,)
+
+    def get(self, request, *args, **kwargs):
+        # TODO:N if a superuser is not in org, should they be
+        # allowed? because they are superuser
+        # and can basically grant permission to themselves!
```
atb00ker, line 698: Quick question. What do you think about this change.
The old `validate_membership` doesn't even allow super to access without the superuser being part of the organization.
This is unexpected, as a superuser can basically give themselves the permission, so, why are they rejected?

openwisp_radius/settings.py
```diff
@@ -92,21 +92,21 @@ def get_settings_value(option, default):
...
+except ImportError as e:  # pragma: no cover
     raise ImproperlyConfigured(
         'OPENWISP_RADIUS_SMS_TOKEN_HASH_ALGORITHM is invalid: {}'.format(str(e))
     )
 
-try:  # pragma: no cover
+try:
     assert int(SMS_TOKEN_LENGTH) <= 8 and int(SMS_TOKEN_LENGTH) >= 4
-except AssertionError:
+except AssertionError:  # pragma: no cover
```
atb00ker, line 109: Move `# pragma: no cover` to except block instead of try block, because try block is cover, I think the intentions of whoever added them was to add it for the except block.

openwisp_radius/api/urls.py
```diff
@@ -62,4 +62,9 @@ def get_api_urls(api_views=None):
             api_views.change_phone_number,
             name='phone_number_change',
         ),
+        url(
+            r'^(?P<slug>[\w-]+)/radiusbatch/(?P<radbatch>[\w-]+)/download_pdf/$',
```
nemesisdesign, line 66: I'd change `download_pdf` to just `pdf` which is more restful

openwisp_radius/api/views.py
nemesisdesign: I am not sure if it was on purpose or not, but supporting superusers was not a priority because these endpoints are used mostly by non superusers.

Does any test fail if we change it to take superusers into account?

openwisp_radius/api/views.py
```diff
@@ -678,3 +675,39 @@ def create_phone_token(self, *args, **kwargs):
...
+        radbatch_model_name = swapper.get_model_name('openwisp_radius', 'RadiusBatch')
+        app_label, _ = swapper.split(radbatch_model_name)
+        return bool(
+            request.user
+            and (request.user.is_superuser or request.user.is_staff)
+            and request.user.has_perm(f'{app_label}.view_radiusbatch')
+        )
+
+
+class DownloadRadiusBatchPdfView(DispatchOrgMixin, APIView):
```
nemesisdesign, line 691: move this under `BatchView`

openwisp_radius/api/views.py
```diff
@@ -678,3 +675,39 @@ def create_phone_token(self, *args, **kwargs):
...
+        if radbatch.strategy == 'prefix':
+            pdf = generate_pdf(kwargs['radbatch'])
+            response = HttpResponse(content_type='application/pdf')
+            response[
+                'Content-Disposition'
+            ] = f'attachment; filename="{radbatch.name}.pdf"'
+            response.write(pdf)
+            return response
+        else:
+            raise Http404()
```
nemesisdesign, line 710: I would use the DRF exception which returns a nicer API response

openwisp_radius/api/views.py
atb00ker: No, none of tests failed, 
Hence, I've added in the `validate_membership` check:
```
if not (user.is_superuser or (self.organization.pk,) in user.organizations_pk):
  .. 
  raise..
```

Let me know if I should remove it or let it be! :smile:

openwisp_radius/utils.py
```diff
@@ -156,13 +155,15 @@ def prefix_generate_users(prefix, n, password_length):
     return users_list, user_password
 
 
-def generate_pdf(prefix, data):
+def generate_pdf(radbatch__uuid):
```
nemesisdesign, line 158: why the double underscore here?

openwisp_radius/base/models.py
```diff
@@ -816,11 +816,11 @@ class AbstractRadiusBatch(OrgMixin, TimeStampedEditableModel):
         max_length=20,
         help_text=_('Usernames generated will be of the format [prefix][number]'),
     )
-    pdf = models.FileField(
+    pdf_json = JSONField(
```
nemesisdesign, line 819: What if in the future we switch from JSON to a different format?

We could also say that now that we have the source data, we are not restricted to generating only PDF, in the future we could deliver that data to the user in other ways. So maybe we could call this field `user_credentials`.

What do you think?

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_change_form.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'admin/change_form.html' %}
+{% load i18n %}
+{% block object-tools-items %}
+{% if download_rad_batch_pdf_url %}
+<li>
+    <a href='{{ download_rad_batch_pdf_url }}' id='downloadpdflink'>{% trans "Download User List PDF" %}</a>
```
nemesisdesign, line 6: in the HTML we should use double quotes `"`

I would change the text to `Download user credentials`.

Can you update the user docs to point out to this new button in the relevant page please?

openwisp_radius/base/models.py
```diff
@@ -816,11 +816,11 @@ class AbstractRadiusBatch(OrgMixin, TimeStampedEditableModel):
...
         null=True,
         blank=True,
         verbose_name='PDF',
-        help_text=_('The pdf file containing list of usernames and passwords'),
+        help_text=_('list of usernames and passwords used to create PDF'),
```
nemesisdesign, line 823: this is not shown to the user or is it?
If it's not, I would remove the help text and convert it to a comment.

openwisp_radius/base/models.py
atb00ker: Renamed to `user_credentials`

openwisp_radius/base/models.py
atb00ker: Changed to comment.

atb00ker:
Rebased with master

atb00ker:
No clue why github was thought there where conflicts, rebasing didn't show any conflicts, so I've rebased it. :smile:

docs/source/user/generating_users.rst
```diff
@@ -25,10 +25,14 @@ This command generates users whose usernames start with a particular prefix. Usa
...
 
+In the response, you can find the field ``user_credentials`` containing the list of users created
+(example: [['username', 'password'], ['sample_user', 'BBuOb5sN']]") and field ``pdf_link``
+containing a download link to a PDF file containing user credentials.
+
 Adding from admin inteface
 --------------------------
 
-At the url ``/admin/openwisp_radius/radiusbatch/add`` one can directly generate users using the prefix and the number of users. A PDF can be downloaded immediately after the users have been generated.
+At the url ``/admin/openwisp_radius/radiusbatch/add`` one can directly generate users using the prefix and the number of users. A PDF containing the list of user credentials can be downloaded from the admin interface button:
```
nemesisdesign, line 35: I am not convinced of this passage because we mention the download button which is only shown at the end of the gif for a brief moment.
Can you add two screenshots (using png which should be lighter), one which shows just the download button and another one that shows the generated PDF?

Then, before the gif we could add something like:

`Here's a sequence of images that show how this feature works:` or something similar.

This would make the passage more complete and easy to understand at glance.

docs/source/user/generating_users.rst
```diff
@@ -25,10 +25,14 @@ This command generates users whose usernames start with a particular prefix. Usa
 
 Note that the expiration and password-length are optional parameters which default to never and 8 respectively.
 
+In the response, you can find the field ``user_credentials`` containing the list of users created
+(example: [['username', 'password'], ['sample_user', 'BBuOb5sN']]") and field ``pdf_link``
+containing a download link to a PDF file containing user credentials.
```
nemesisdesign, line 30: what if we change the command so it returns the pdf as its output and we change the instruction to suggest doing

```
./manage.py command > output.pdf
```

I think this is more unix friendly and more practical for most users.

openwisp_radius/api/views.py
```diff
@@ -377,17 +373,53 @@ def dispatch(self, *args, **kwargs):
...
+        return bool(
+            request.user
+            and (request.user.is_superuser or request.user.is_staff)
+            and request.user.has_perm(f'{app_label}.view_radiusbatch')
+        )
+
+
+class DownloadRadiusBatchPdfView(DispatchOrgMixin, APIView):
+    authentication_classes = (UserTokenAuthentication, SessionAuthentication)
+    permission_classes = (IsAllowedRadiusBatch,)
```
nemesisdesign, line 397: I think we could use `('IsAdminUser'', 'DjangoModelPermissions')`, (having to define IsAllowedRadiusBatch should not be needed), you will need to specify the `queryset` attribute (eg: `RadiusBatch.objects.all()`).

openwisp_radius/api/urls.py
```diff
@@ -62,4 +62,9 @@ def get_api_urls(api_views=None):
             api_views.change_phone_number,
             name='phone_number_change',
         ),
+        url(
+            r'^(?P<slug>[\w-]+)/radiusbatch/(?P<radbatch>[\w-]+)/pdf/$',
```
nemesisdesign, line 66: let's change this to `pk`

openwisp_radius/tests/test_api.py
```diff
@@ -890,6 +894,65 @@ def test_api_batch_add_users(self):
...
+            os.path.dirname(os.path.abspath(__file__)), 'static', 'test_batch.csv'
+        )
+        with open(path_csv, 'rt') as file:
+            data = self._get_post_defaults(
+                {'name': 'test_csv1', 'strategy': 'csv', 'csvfile': file}
+            )
+            response = self.client.post(
+                reverse('radius:batch'), data, HTTP_AUTHORIZATION=self.auth_header
+            )
+        self.assertEqual(response.status_code, 201)
```
nemesisdesign, line 946: can you create this using the model instead of doing API calls? The API calls are distracting in this test.

openwisp_radius/tests/test_api.py
```diff
@@ -890,6 +894,65 @@ def test_api_batch_add_users(self):
...
+        response_json = json.loads(response.content)
+        pdf_url = reverse(
+            'radius:download_rad_batch_pdf',
+            args=[response_json['organization'], response_json['id']],
+        )
+        self.assertEqual(response_json['pdf_link'], None)
+        pdf_response = self.client.get(pdf_url)
+        self.assertEqual(pdf_response.status_code, 404)
```
nemesisdesign, line 954: you're testing only that the strategy is prefix but not checking a non existing valid UUID, nor an invalid UUID

note: the invalid UUID check may be avoided if we use path: https://docs.djangoproject.com/en/3.0/ref/urls/#path
As we're doing here:  https://github.com/openwisp/openwisp-firmware-upgrader/pull/84

openwisp_radius/tests/test_admin.py
```diff
@@ -228,14 +228,42 @@ def test_radiusbatch_change(self):
...
         )
-        url = reverse(
-            'admin:{0}_radiusbatch_change'.format(self.app_label), args=[obj.pk]
-        )
+        url = reverse(f'admin:{self.app_label}_radiusbatch_change', args=[obj.pk])
         response = self.client.get(url)
         self.assertContains(response, 'ok')
         self.assertNotContains(response, 'errors')
 
+    def test_radiusbatch_change_pdf_exists(self):
```
nemesisdesign, line 236: what about `test_radiusbatch_change_contains_pdf_download`?

openwisp_radius/tests/test_admin.py
```diff
@@ -228,14 +228,42 @@ def test_radiusbatch_change(self):
...
+            expiration_date='1998-01-28',
+        )
+        url = reverse(f'admin:{self.app_label}_radiusbatch_change', args=[obj.pk])
+        response = self.client.get(url)
+        pdf_url = reverse(
+            'radius:download_rad_batch_pdf', args=[obj.organization, obj.pk],
+        )
+        self.assertContains(response, pdf_url)
+
+    def test_radiusbatch_change_csv_pdf_hidden(self):
```
nemesisdesign, line 250: what about `test_radiusbatch_change_not_contains_pdf_download`?

openwisp_radius/api/views.py
```diff
@@ -377,17 +373,53 @@ def dispatch(self, *args, **kwargs):
...
+
+    def get(self, request, *args, **kwargs):
+        self.validate_membership(request.user)
+        radbatch = RadiusBatch.objects.get(pk=kwargs['radbatch'])
```
nemesisdesign, line 401: so what happens if this line raises RadiusBatch.DoesNotExist? It doesn't seem this is catched.
Please write a test for this case as mentioned in one of the previous comments.

openwisp_radius/api/urls.py
nemesisdesign: I would start using path and use the uuid converter: https://docs.djangoproject.com/en/3.0/ref/urls/#path
See also here: openwisp/openwisp-firmware-upgrader#84

The converters are handy because they allow us to avoid having to implement checks ourselves (try supplying a bogus UUID with the current code and see what happens).

atb00ker: To change this, I'll have to change `DispatchOrgMixin` used by other urls, i.e change other urls as well -- on it. :smile:

openwisp_radius/tests/test_api.py
atb00ker: If I don't use the API, I wouldn't be able to test:
`self.assertEqual(response_json['pdf_link'], None)`


Do you think I should create another test for this and change this test to : 
```
    def test_batch_model_csv_pdf_link_404(self):
        reader = [['rohith', "", 'rohith@openwisp.com', "", ""]]
        batch = self._create_radius_batch(
            name='test_csv3', strategy='csv', csvfile=self._get_csvfile(reader)
        )
        pdf_url = reverse(
            'radius:download_rad_batch_pdf',
            args=[self._get_org().pk, batch.id],
        )
        self._superuser_login()
        pdf_response = self.client.get(pdf_url)
        self.assertEqual(pdf_response.status_code, 404)
```

openwisp_radius/api/views.py
atb00ker: Thanks for this, much better! :smile:

docs/source/user/generating_users.rst
atb00ker: I made a mistake here, It's suppose, I have made some changes, it used to say `Generated a batch of users with prefix <given prefix>`; I have reverted to this output, do you think it's better to output the pdf instead?

nemesisdesign: we have to think like users, if you were using this feature to generate users and get the credentials, what would you prefer?
I think I would prefer to have the PDF.

openwisp_radius/tests/test_api.py
nemesisdesign: I have so much work to review and merge that I can't bother with these minor details anymore, do what you feel is best, I trust you :+1:

atb00ker: Okay, I am not making any change here, I think it's better to check the two things together. :smile:

openwisp_radius/api/urls.py
```diff
@@ -7,59 +7,64 @@ def get_api_urls(api_views=None):
...
+        path('accounting/', api_views.accounting, name='accounting'),
+        path('batch/', api_views.batch, name='batch'),
         # registration differentiated by organization
-        url(r'^(?P<slug>[\w-]+)/account/$', api_views.register, name='rest_register'),
+        path('<uuid:pk>/account/', api_views.register, name='rest_register'),
         # password reset
-        url(
-            r'^(?P<slug>[\w-]+)/account/password/reset/confirm/$',
+        path(
+            '<uuid:pk>/account/password/reset/confirm/',
```
nemesisdesign, line 18: next time try to implement these changes in a separate PR

openwisp_radius/api/views.py
```diff
@@ -350,44 +355,71 @@ def post(self, request, *args, **kwargs):
...
+            self.organization = Organization.objects.get(pk=kwargs['pk'])
         except Organization.DoesNotExist:
             raise Http404()
         return super().dispatch(*args, **kwargs)
 
     def validate_membership(self, user):
-        if (self.organization.pk,) not in user.organizations_pk:
-            message = _('User "{}" is not member ' 'of "{}"').format(
-                user.username, self.organization.slug
+        if not (user.is_superuser or (self.organization.pk,) in user.organizations_pk):
```
nemesisdesign, line 380: the outer parenthesis are redundant here

PR #109, opened 2020-06-16
[radius/models] Use UUID or natural keys #97

### LINKED ISSUE #97: [radius/models] Use UUID or natural keys
Some objects, like the accounting object, have keys which can be used as natural primary keys (eg: session ID or unique accounting ID), but we need to double check if this is ok.

In other objects, we could just switch to UUID.

I don't want to rely on auto-increment IDs.

NoumbissiValere:
Fixed the issue by replacing primary keys with
natural keys or UUID in case there was no
possible natural key

Fixes #97


NoumbissiValere:
@nemesisdesign please can you review this?

NoumbissiValere:
Thanks for reviewing

atb00ker:
@NoumbissiValere there is a conflict. :smile:

NoumbissiValere:
@atb00ker will fix it. Thanks :smile:

NoumbissiValere:
@atb00ker conflict resolved :blush:

PR #110, opened 2020-06-17
[improvement] Seperate logger name for settings

atb00ker:
Similar to https://github.com/openwisp/openwisp-network-topology/pull/67

### DIFF
--- a/openwisp_radius/settings.py
+++ b/openwisp_radius/settings.py
@@ -6,7 +6,10 @@ from uuid import UUID
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
-logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


openwisp_radius/settings.py
```diff
@@ -6,7 +6,10 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
-logger = logging.getLogger(__name__)
+# The logger 'pre_django_setup' is suppose to be
```
nemesisdesign, line 9: should be supposed?

atb00ker: Didn't know it's supposed to be `supposed` when `suppose` is used phrase in a case like this. Changed it.

PR #113, opened 2020-06-22
[private-storage] Implement private storage for csv files #101 

### LINKED ISSUE #101: [radius/private-storage] Implement private storage for AbstractRadiusBatch files
AbstractRadiusBatch stores files (csv and pdf) in the public media folder.

We should change this to store files in a private location that can be accessed only by staff users with the right permissions, like we do in openwisp-firmware-upgrader:

- [openwisp_firmware_upgrader/base/models.py](https://github.com/openwisp/openwisp-firmware-upgrader/blob/5db431e/openwisp_firmware_upgrader/base/models.py#L131-L133)
- [openwisp_firmware_upgrader/private_storage](https://github.com/openwisp/openwisp-firmware-upgrader/tree/master/openwisp_firmware_upgrader/private_storage)

See how it's implemented in the link above and replicate it in this module, don't forget to update the documentation to mention this and write tests.

atb00ker:
Closes #101


openwisp_radius/urls.py
```diff
@@ -13,17 +14,22 @@ def get_urls(api_views=None, social_views=None):
...
+                'freeradius/social-login/<slug>/',
                 social_views.redirect_cp,
                 name='redirect_cp',
             )
         )
     return urls
 
 
-app_name = 'radius'
-urlpatterns = get_urls()
```
atb00ker, line 29: This is a backwards incompatible change, as a different output is given by the file, but I don't think it's used in any dependency.

openwisp_radius/urls.py
```diff
@@ -13,17 +14,22 @@ def get_urls(api_views=None, social_views=None):
...
     if not social_views:
         from .social import views as social_views
-    urls = [url(r'api/v1/', include(get_api_urls(api_views)))]
+    urls = [path('api/v1/', include(get_api_urls(api_views)))]
     if app_settings.SOCIAL_LOGIN_ENABLED:
         urls.append(
-            url(
-                r'^freeradius/social-login/(?P<slug>[\w-]+)/$',
+            path(
+                'freeradius/social-login/<slug>/',
```
nemesisdesign, line 21: use the converter: `<slug:slug>`, see https://docs.djangoproject.com/en/3.0/topics/http/urls/#path-converters

tests/openwisp2/sample_users/migrations/0003_update_admins_permissions.py
```diff
@@ -0,0 +1,19 @@
+# Manually Created
```
nemesisdesign, line 1: I included `update_admins_permissions` in `0002_default_groups_and_permissions.py` on `master`, please remove this (let's try to not add too many migrations in the sample app to keep them easy to scan if needed.

openwisp_radius/views.py
```diff
@@ -0,0 +1,23 @@
...
+    slug_field = 'csvfile'
+    slug_url_kwarg = 'csvfile'
+
+    def can_access_file(self, private_file):
+        user = private_file.request.user
+        if not user.is_authenticated:
+            return False
+        org = (self.object.organization.pk,)
+        user_organizations = user.organizations_pk.all()
+        return user.is_superuser or (user.is_staff and org in user_organizations)
```
nemesisdesign, line 20: does this take in consideration if the user is flagged as `is_admin` of the organization?

I noticed that also this doesn't do it: https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/private_storage/views.py and opened an issue for it https://github.com/openwisp/openwisp-firmware-upgrader/issues/90

atb00ker: Thanks, the different testcases make much more sense now! :smile:

tests/openwisp2/settings.py
```diff
@@ -39,6 +39,7 @@
     # openwisp radius
     'openwisp_radius',
     'openwisp_users',
+    'private_storage',
```
nemesisdesign, line 42: Seems you forgot to add this in the setup instructions.

tests/openwisp2/settings.py
```diff
@@ -102,6 +103,7 @@
 USE_L10N = False
 USE_TZ = True
 MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
+PRIVATE_STORAGE_ROOT = os.path.join(MEDIA_ROOT, 'private')
```
nemesisdesign, line 106: Seems you forgot to add this in the setup instructions.

openwisp_radius/views.py
```diff
@@ -0,0 +1,27 @@
...
+    model = load_model('openwisp_radius', 'RadiusBatch')
+    model_file_field = 'csvfile'
+    slug_field = 'csvfile'
+    slug_url_kwarg = 'csvfile'
+
+    def can_access_file(self, private_file):
+        user = private_file.request.user
+        if not user.is_authenticated:
+            return False
+        org = (self.object.organization.pk,)
```
nemesisdesign, line 20: we should now simplify this block of code with just the following:

```python
return user.is_superuser or (user.is_staff and user.is_manager(self.object.organization))
```

See [Model helpers in openwisp-users](https://github.com/openwisp/openwisp-users/#model-helpers).

atb00ker: Done! :smile:

Issue #94, opened 2020-06-04
[api/radius] Fix BatchView permission and authorization


nemesisdesign:
The `BatchView` of the API should be usable only by admin/staff users.

The organization should be part of the URL, eg: `/api/v1/<org-slug>/radius/batch/`.

It should use the [user authentication from openwisp-users](https://github.com/openwisp/openwisp-users/#obtain-authentication-token) (not the other authentication designed to be used by freeradius).
It should use the django permissions to reject requests coming from users which do not have permissions to manipulate batch objects, it should allow superusers though.

Test will have to be changed. API docs need to be updated.

PR #120, opened 2020-07-01
[multi-tenancy] Freeradius instance without the organization info

atb00ker:


openwisp_radius/api/views.py
```diff
@@ -326,6 +353,9 @@ def get_queryset(self):
 
 
 class BatchView(TokenAuthorizationMixin, generics.CreateAPIView):
+    # TODO:N Who can use obtaintokenview?
+    # Should they be allowed to use batchview to create
+    # users? (Uses TokenAuthentication)
```
atb00ker, line 358: `BatchView` uses `TokenAuthentication`, so basically anyone having access to `ObtainAuthTokenView` can have access to `BatchView`; this looks undesirable because my understanding is that `ObtainAuthTokenView` can be accessed by user.

I want to confirm the same before making any change.

openwisp_radius/api/views.py
```diff
@@ -418,15 +448,23 @@ def get(self, request, *args, **kwargs):
...
-        radius_token, _ = RadiusToken.objects.get_or_create(user=user)
+    def get_or_create_radius_token(self, user, organization=None, expired=False):
+        try:
+            radius_token, _ = RadiusToken.objects.get_or_create(
+                user=user, organization=organization, is_expired=expired
+            )
+        except IntegrityError:
+            radius_token = RadiusToken.objects.get(user=user)
+            radius_token.organization = organization
+            radius_token.is_expired = expired  # TODO:N condition required before chaning value because of validation?
```
atb00ker, line 459: I want to understand which operations should set `is_expired` as `True` and if by default it should be `True` or `False`.
Also, since this flag is used to check if the token should be used for authentication,
what do you think about renaming `is_auth_allowed` instead?

openwisp_radius/migrations/__init__.py
```diff
@@ -11,6 +11,12 @@ def get_swapped_model(apps, app_name, model_name):
     return apps.get_model(app, model)
 
 
+def delete_old_radius_token(apps, schema_editor):
+    # TODO:N Do we want to really delete all the old tokens?
+    # What consequenses should I think about?
```
atb00ker, line 16: Just wanted to inform this change require deleting all old radius tokens, alternative would be allowing org value as `None`.

openwisp_radius/api/views.py
```diff
@@ -71,6 +77,27 @@ class TokenAuthentication(BaseAuthentication):
...
+                )
+            except RadiusToken.DoesNotExist:
+                message = (
+                    'User radius token does not exist & organization '
+                    'uuid & token not provided in request'
+                )
+                logger.warning(message)
+                # TODO:N Returning 403 not 401 as suggested in the issue
+                # rest_auth: WWW-Authenticate header for 401 responses, else coerce to 403
+                # After some reading on it, I agree, details missing is 403.
```
atb00ker, line 93: In the issue it's mention this case should return `401` but rest_auth suggests, or rather forces to return `403` because auth failed due to missing information.

openwisp_radius/api/views.py
```diff
@@ -71,6 +77,27 @@ class TokenAuthentication(BaseAuthentication):
...
         uuid, token = self.get_uuid_token(request)
+        if not uuid and not token:
+            try:
+                radtoken = RadiusToken.objects.get(
+                    user__username=request.data.get('username')
+                )
+            except RadiusToken.DoesNotExist:
+                message = (
+                    'User radius token does not exist & organization '
+                    'uuid & token not provided in request'
```
nemesisdesign, line 88: it has two &, I think it should be or

openwisp_radius/api/views.py
```diff
@@ -71,6 +77,27 @@ class TokenAuthentication(BaseAuthentication):
...
+                logger.warning(message)
+                # TODO:N Returning 403 not 401 as suggested in the issue
+                # rest_auth: WWW-Authenticate header for 401 responses, else coerce to 403
+                # After some reading on it, I agree, details missing is 403.
+                raise NotAuthenticated(message)
+            if not radtoken.is_expired:
+                return (AnonymousUser(), str(radtoken.organization_id))
+            else:
+                raise AuthenticationFailed(
+                    'Radius token expired & organization uuid & token not provided in request'
```
nemesisdesign, line 99: same here

openwisp_radius/api/views.py
nemesisdesign: leave this untouched, there's a dedicated issue for this view: https://github.com/openwisp/openwisp-radius/issues/94

openwisp_radius/api/views.py
nemesisdesign: when the radius session expires, the captive portal or radius itself will send an accounting stop request.
We could intercept that and emit a signal, then connect a receiver to that signal to do this.

A good name for this field could be `can_auth` or `is_auth_valid`.
An alternative could be to delete the object, but I'm not 100% sure freeradius does not perform any other operation after the user logs out or the session expires.

openwisp_radius/migrations/__init__.py
nemesisdesign: currently they're disposable, they're created, consumed and destroyed right afterwards

openwisp_radius/social/views.py
```diff
@@ -20,10 +20,15 @@ def get(self, request, *args, **kwargs):
         """
         if not request.GET.get('cp'):
             return HttpResponse(_('missing cp GET param'), status=400)
-        self.authorize(request, *args, **kwargs)
-        return HttpResponseRedirect(self.get_redirect_url(request))
+        slug = kwargs.get('slug')
+        try:
+            org = Organization.objects.get(slug=slug)
```
nemesisdesign, line 25: hey I just noticed that maybe here we can use `get_object_or_404`?

openwisp_radius/social/views.py
```diff
@@ -55,7 +55,7 @@ def get_redirect_url(self, request):
         Token.objects.filter(user=user).delete()
         RadiusToken.objects.filter(user=user).delete()
         token = Token.objects.create(user=user)
-        rad_token = RadiusToken.objects.create(user=user)
+        rad_token = RadiusToken.objects.create(user=user, organization=organization)
```
nemesisdesign, line 58: what if a token already exists?

openwisp_radius/api/views.py
```diff
@@ -71,6 +77,27 @@ class TokenAuthentication(BaseAuthentication):
     def authenticate(self, request):
         self.check_organization(request)
         uuid, token = self.get_uuid_token(request)
+        if not uuid and not token:
+            try:
+                radtoken = RadiusToken.objects.get(
+                    user__username=request.data.get('username')
+                )
```
nemesisdesign, line 84: nice, let's remember to use caching here, the cache key could be `rt-{username}` (timeout 24 hours) and its contents should be updated whenever a radiustoken is created/updated, the retrieval can then be done using the cache first, then query the DB if in the cache is not present

openwisp_radius/social/views.py
atb00ker: Token is deleted 2 lines above the highlighted line, however in interest of reusing code, I am going for `get_or_create_radius_token` here as well! :smile:

docs/source/user/api.rst
```diff
@@ -353,7 +366,7 @@ Login (Obtain User Auth Token)
 
 .. code-block:: text
 
-    /api/v1/account/token/
+    /api/v1/<organization-slug>/account/token/
```
atb00ker, line 369: Not, I noticed this did not include `<organization-slug>` but in usage, I was putting <org-slug> to get the token, I assume this one only a typing mistake.

docs/source/user/api.rst
```diff
@@ -385,7 +398,7 @@ Return also the radius user token in the response.
 
 .. code-block:: text
 
-    /api/v1/account/token/validate/
+    /api/v1/<organization-slug>account/token/validate/
```
atb00ker, line 401: Not, I noticed this did not include `<organization-slug>` but in usage, I was putting <org-slug> to get the token validation, I assume this one only a typing mistake.

openwisp_radius/api/views.py
```diff
@@ -419,15 +453,24 @@ def get(self, request, *args, **kwargs):
...
-    def get_or_create_radius_token(self, user):
-        radius_token, _ = RadiusToken.objects.get_or_create(user=user)
+    def get_or_create_radius_token(self, user, organization, enable_auth=False):
+        try:
+            radius_token, _ = RadiusToken.objects.get_or_create(
+                user=user, organization=organization
+            )
+        except IntegrityError:
+            radius_token = RadiusToken.objects.get(user=user)
+            radius_token.organization = organization
```
atb00ker, line 463: When user is using `org1` and sends a `user_auth_token` request to org2, the organization is changed, I think we also need to send accounting `Stop` request to `org1`, right? Or is the request send from elsewhere when user stops using the connection (like captive portal, freeradius etc)?

nemesisdesign: we can check if the user has an open session (RadiusAccounting), retrieve it and flag it as terminated, as termination cause we have to choose the best value from this list: https://tools.ietf.org/html/rfc2866#page-19

atb00ker:
```
NAS Request          NAS ended session for a non-error reason not otherwise listed here.
```

This looks good to me! :smile:

openwisp_radius/tests/test_api.py
```diff
@@ -177,7 +238,11 @@ def test_postauth_400(self):
     def test_postauth_no_token_403(self):
         response = self.client.post(reverse('radius:postauth'), {})
         self.assertEqual(response.status_code, 403)
-        self.assertEqual(response.data, {'detail': 'Token authentication failed'})
+        self.assertEqual(
+            response.data['detail'],
+            'Radius token does not exist. Obtain new token or provide '
+            'API token and organization UUID.',
```
nemesisdesign, line 244: What about: "Obtain a new radius token or provide the organization UUID and API token."?

openwisp_radius/api/views.py
nemesisdesign: yes, we can use this termination cause to close active user sessions

docs/source/user/api.rst
```diff
@@ -52,6 +56,21 @@ or in the form of a query string parameter as shown below.
...
+* Radius Token (recommended)::
+
+      curl -X POST http://localhost:8000/api/v1/authorize/ \
+           -d "username=<username>&password=<password>"
+
+.. warning::
+    If you are using Radius Token method, remember that one
+    user account can only authenticate with one organization
+    at a time, i.e a single user account cannot consume
+    services from multiple organizations simultaneously.
```
nemesisdesign, line 72: nice :+1:

docs/source/user/api.rst
```diff
@@ -52,6 +56,21 @@ or in the form of a query string parameter as shown below.
...
 Requests which contain an invalid token will receive a ``403`` HTTP error.
 
+To use Radius Token authentication method `send a request using user API to obtain the auth token <#login-obtain-user-auth-token>`_. This can be integrated in captive portal to be send on user login.
+The token is valid for authentication only as long as freeradius accounting ``Stop`` request
+is not send or token is not deleted. Once the token is created, the following can be used to send the request:
+
+* Radius Token (recommended)::
+
+      curl -X POST http://localhost:8000/api/v1/authorize/ \
+           -d "username=<username>&password=<password>"
```
nemesisdesign, line 66: I think it should be:

`"username=<username>&password=<radius_token>"`

docs/source/user/api.rst
```diff
@@ -52,6 +56,21 @@ or in the form of a query string parameter as shown below.
 
 Requests which contain an invalid token will receive a ``403`` HTTP error.
 
+To use Radius Token authentication method `send a request using user API to obtain the auth token <#login-obtain-user-auth-token>`_. This can be integrated in captive portal to be send on user login.
+The token is valid for authentication only as long as freeradius accounting ``Stop`` request
+is not send or token is not deleted. Once the token is created, the following can be used to send the request:
```
nemesisdesign, line 61: since this is the recommended way, we can move it before the old method

openwisp_radius/api/views.py
atb00ker: Worked on it.

docs/source/user/api.rst
atb00ker: While `<password>` & `<radius_token.key>` both work for `authorize`, I used `<password>` because it's used in the examples above as well, should I change it for all the examples?

docs/source/user/api.rst
atb00ker: Done! :smile:

openwisp_radius/tests/test_api.py
atb00ker: Done! :smile:

openwisp_radius/api/views.py
```diff
@@ -442,11 +485,21 @@ def get_response_data(self, user):
...
         data['radius_user_token'] = radius_token.key
         return data
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        user = self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(
+            self.get_response_data(user, self.organization),
```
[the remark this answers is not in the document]
atb00ker: Thanks for the info, fixed! :smile:

openwisp_radius/api/views.py
```diff
@@ -71,6 +78,36 @@ class TokenAuthentication(BaseAuthentication):
     def authenticate(self, request):
         self.check_organization(request)
         uuid, token = self.get_uuid_token(request)
+        if not uuid and not token:
```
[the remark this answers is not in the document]
atb00ker: Done! :smile:

docs/source/user/api.rst
nemesisdesign: no, only here

openwisp_radius/api/views.py
```diff
@@ -466,8 +524,8 @@ def post(self, request, *args, **kwargs):
         )
         serializer.is_valid(raise_exception=True)
         user = self.get_user(serializer, *args, **kwargs)
-        token, created = UserToken.objects.get_or_create(user=user)
-        radius_token = self.get_or_create_radius_token(user)
+        token, _ = UserToken.objects.get_or_create(user=user)
+        radius_token = self.get_or_create_radius_token(user, self.organization, True)
```
nemesisdesign, line 528: two things:

1. let's pass `enable_auth=True` instead of `True`, for readability
2. we should add another flag to this method to renew the token, eg:

```python
radius_token = self.get_or_create_radius_token(user, self.organization, enable_auth=True, renew=True)
```

We should renew the token each time the user uses obtain auth token or validate auth token, this is related to `DISPOSABLE_RADIUS_USER_TOKEN`.

Please also write tests for this (1 for obtain auth token view and another one for validate auth token view).

openwisp_radius/apps.py
```diff
@@ -53,6 +55,11 @@ def connect_signals(self):
             sender=Organization,
             dispatch_uid='openwisp_radius_org_post_save',
         )
+        post_save.connect(
+            radiustoken_post_save,
+            sender=RadiusToken,
+            dispatch_uid='openwisp_radius_radiustoken_post_save',
+        )
```
nemesisdesign, line 62: here you can add also the `post_delete`

openwisp_radius/receivers.py
```diff
@@ -44,3 +46,12 @@ def organization_post_save(instance, **kwargs):
         rg.name = rg.name.replace(instance.__old_slug, instance.slug)
         rg.full_clean()
         rg.save()
+
+
+def radiustoken_post_save(instance, **kwargs):
+    username = f'rt-{instance.user.username}'
```
nemesisdesign, line 52: this could be a method in the `RadiusToken` model, which would make it consistent with `delete()`

openwisp_radius/base/models.py
```diff
@@ -983,6 +984,10 @@ def save(self, *args, **kwargs):
             self.key = self.generate_key()
         return super().save(*args, **kwargs)
 
+    def delete(self, *args, **kwargs):
+        cache.delete(f'rt-{self.user.username}')
+        super().delete(*args, **kwargs)
```
nemesisdesign, line 989: delete is often not called (eg bulk delete or other cases), you could rename this to `delete_cache` and call this method from a post delete signal

atb00ker: I covered the bulk delete case from admin, but I'll do this for robustness, thanks.

docs/source/user/api.rst
```diff
@@ -15,9 +15,33 @@ consumed by FreeRADIUS.
 Organization API Token
 ----------------------
 
-Only requests containing the right API token and Organization UUID will able
-to talk to the API endpoints consumed by freeradius
-(`Authorize`_, `Post Auth`_, `Accounting`_).
+There are three ways to authenticate your requests:
+
+1. Radius token (recommended)
+2. Bearer token
```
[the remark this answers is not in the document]
atb00ker: Are you say this line should be added as a short description here? Because the information is mentioned below.
I am changing the order to make things more clear.

docs/source/user/api.rst
```diff
@@ -15,9 +15,33 @@ consumed by FreeRADIUS.
...
+3. Querystring
```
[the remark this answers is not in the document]
atb00ker: Are you say this line should be added as a short description here? Because the information is mentioned below.
I am changing the order to make things more clear.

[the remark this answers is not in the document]
atb00ker: Completely changed the order in order to make it more clear, please let me know if it needs any further change.

docs/source/user/api.rst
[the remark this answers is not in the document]
atb00ker: Completely changed the order in order to make it more clear, please let me know if it needs any further change.

docs/source/user/api.rst
```diff
@@ -15,9 +15,33 @@ consumed by FreeRADIUS.
...
+
+To use Radius token authentication method `send a request using
+user API to obtain the auth token <#login-obtain-user-auth-token>`_.
+This can be integrated in captive portal to be send on user login.
```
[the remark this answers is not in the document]
atb00ker: Tried making this more clear, let me know if this one also needs more attention. :smile:

openwisp_radius/receivers.py
```diff
@@ -44,3 +44,11 @@ def organization_post_save(instance, **kwargs):
...
         rg.full_clean()
         rg.save()
+
+
+def radiustoken_post_save(instance, **kwargs):
+    instance.change_in_cache()
+
+
+def radiustoken_post_delete(instance, **kwargs):
+    instance.delete_cache()
```
atb00ker, line 54: @nemesisdesign Okay, after reading some of your reviews, my understanding was that you suggest moving the operation of the signals inside the model itself and calling the function from the model itself, this hasn't been done anywhere else so I still have doubts that I've understood you correctly or not, please confirm! :smile:

openwisp_radius/api/views.py
atb00ker: Just to be clear, the renew flag will delete the existing `RadiusToken` and create a new one, is that correct?

nemesisdesign: I would just update the current one for efficiency, these methods can be used by many thousands of users multiple times a day

openwisp_radius/receivers.py
nemesisdesign: Look here: https://github.com/openwisp/openwisp-network-topology/pull/76/files#diff-7e33562dabdb281320b3fc9a933ccfc1R16-R37

We've been using similar patterns lately.

I would define the receiver as a class method in the AppConfig class for convenience (it's easier to read and follow the code this way), then call the model method like you're doing here.

openwisp_radius/api/views.py
atb00ker: The `key` attribute of `RadiusToken` is the model's primary key, and as far as I know, primary key cannot be changed without creating a new model, I think we should create a new `UUID` primary key, what do you think, that will allow us to not delete the `RadiusToken` for the change!?

nemesisdesign: oh damn, you're right, I forgot there was no UUID on this model. Let's keep it how you did it. Thanks

atb00ker:
The auth flags:
- RegisterView (enable_auth=True,  renew=False)
- ValidateAuthTokenView (enable_auth=False,  renew=True)
- ObtainAuthTokenView(enable_auth=True, renew=True)
- RedirectCaptivePageView (enable_auth=True, renew=True)

atb00ker:
@nemesisdesign to recap, the flow of this PR has changed a couple of times, that has made the cache useless. (It's saved but never read -- because the RadiusToken is renewed after every request and cache gets deleted everytime RadiusToken is renewed.)

Now, since multiple flags were introduced during different reviews, I want to confirm again where which flag is `True` depending on that, I'll work fixing cache so that it's read, currently, [flags are set as comment above](https://github.com/openwisp/openwisp-radius/pull/120#issuecomment-660022869).

After the cache is fixed, this should be ready!

atb00ker:
@nemesisdesign  https://github.com/openwisp/openwisp-radius/pull/120#pullrequestreview-453771489 contradicts with the docstring of the `RedirectCaptivePageView.get_redirect_url` : `refreshes token and returns the captive page URL`, shouldn't that need a renew as well?

Also the renew action only happens after each request when `DISPOSABLE_RADIUS_USER_TOKEN` is `True` or should it happen always?

docs/source/user/api.rst
```diff
@@ -9,15 +9,69 @@ API Documentation
...
+3. The captive login must initiate the ``POST`` request to the captive portal, using the radius user token as password, example:
+
+.. code-block:: text
+
+    curl -X POST http://localhost:8000/api/v1/authorize/ \
+         -d "username=<username>&password=<radius_token>"
+
+The ``<radius_token>`` is valid for authentication only as long
+as freeradius accounting ``Stop`` request is not sent or token
+is not deleted.
```
nemesisdesign, line 33: I think this is incorrect, the radius token is valid for authenticaiton only once if the disposable radius token setting is true (which si the default)

docs/source/user/api.rst
```diff
@@ -9,15 +9,69 @@ API Documentation
...
+
+      curl -X POST http://localhost:8000/api/v1/authorize/ \
+           -H "Authorization: Bearer <org-uuid> <token>" \
+           -d "username=<username>&password=<password>"
+
+3. Querystring
+--------------
+
+Similar to Bearer token method, Organization UUID and Organization
+Token is authorization information required to be sent in querystring.
```
nemesisdesign, line 62: I find this passage hard to read, can you try to make it simpler please?

docs/source/user/api.rst
atb00ker: `token is not deleted.` would do it, but let me make this clearer by explicitly mentioning the flag.

atb00ker:
```
By default, ``<radius_token>`` is valid for authentication for one
request only. A new ``<radius_token>`` needs to be `obtained for
each request <#login-obtain-user-auth-token>`_.
However, if `OPENWISP_RADIUS_DISPOSABLE_RADIUS_USER_TOKEN
</user/settings.html#openwisp-radius-disposable-radius-user-token>`_
is set to ``False``, the ``<radius_token>`` is valid for authentication
as long as freeradius accounting ``Stop`` request is not sent
or token is not deleted.
```

openwisp_radius/api/views.py
```diff
@@ -70,9 +79,40 @@
...
+                message = _(
+                    'Radius token does not exist. Obtain a new radius token '
+                    'or provide the organization UUID and API token.'
+                )
+                logger.warning(message)
+                raise NotAuthenticated(message)
+            if radtoken.can_auth:
+                org_uuid = str(radtoken.organization_id)
+                cache.set(f'rt-{username}', org_uuid, 86400)
+                return (AnonymousUser(), org_uuid)
```
nemesisdesign, line 102: I think there's a misunderstanding here.

The `can_auth` attribute should be used when the user tries to log in via the captive page, which happens in the `authorize` method: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/views.py#L171-L183
I see you already changed that method in this PR so that should be enough.

If we leave this change here, after the user authenticates via the captive portal, their token will be flagged as `can_auth` False and if for some reason the cache is cleaned, this if block will be executed and the system won't be able to recognize accounting requests for this user (accounting requests are sent periodically by the NAS while the user is using the network).

We have to explicitly document this, `can_auth` indicates whether the token can be used for the freeradius authorization step (authorize view).

In this class instead we just deal with finding out to which organization the incoming API requests coming from freeradius belong.

It may be a good idea to change the name of this class altogether to avoid ambiguities.

We could rename it like:

- FreeradiusApiAuthentication
- FreeradiusAuthentication
- RadiusMultitenancyAuth

Or anything else that makes it explicit that here we're implementing multi-tenancy for HTTP API requests coming from freeradius, we're just making the system aware to which organization the request coming from freeradius belongs to, because freeradius itself has no concept of `organization` or `tenant`.

atb00ker:
Note: currently needs to be rebased with https://github.com/openwisp/openwisp-radius/pull/142

openwisp_radius/api/views.py
```diff
@@ -162,25 +194,27 @@ def authenticate_user(self, request, user):
...
         """
-        return user.check_password(
-            request.data.get('password')
-        ) or self.check_user_token(
-            request, user
-        )  # noqa
+        return bool(
+            user.check_password(request.data.get('password'))
+            or self.check_user_token(request, user)
+        )
```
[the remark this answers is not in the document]
atb00ker: Also, user can bypass the `can_auth` flag status and authenticate with password.
But I think that's correct behavior because if they have the password, the can get a new token anyway!

PR #123, opened 2020-07-02
[fix] cache checks for string 'uuid' not org's uuid

atb00ker:


openwisp_radius/tests/test_api.py
```diff
@@ -1441,14 +1441,21 @@ def test_cache(self):
...
-        rad.delete()
-        self.assertEqual(None, cache.get(rad.organization.pk))
+        with self.subTest('Cache & token match'):
+            self.assertEqual(rad.token, cache.get(rad.organization.pk))
+        with self.subTest('Force changed corrupt cache returns 403'):
+            cache.set(rad.organization.pk, 'wrong-value')
+            response = self.client.post(
+                post_url, {'username': 'tester', 'password': 'tester'}
+            )
+            self.assertEqual(response.status_code, 403)
```
atb00ker, line 1451: Added a new testcase to make testing of the cache more robust.
If the value of cache is not as expected (here forcefullly changed), then it returns 403; this test would have reported the problem solved in this issue.

PR #125, opened 2020-07-03
[radius/models] Use UUID or natural keys #97

nemesisdesign:
Follow up to #109. Thanks @NoumbissiValere!


NoumbissiValere:
Thank @nemesisdesign. I see the improvements

NoumbissiValere:
Yes @nemesisdesign I am still working on this.

atb00ker:
this PR makes 3 migration files, I think that can be avoided! :smile:

NoumbissiValere:
> this PR makes 3 migration files, I think that can be avoided!

Yes @atb00ker  . But i don't think it should be avoided. 

- The first one is used to add the null uuid field
- The second, to populate it with unique values
- The third to make the uuid field the id

we can trim it down to 2 but i think for easy future debugging, it should be 3.

I will get back here when i am done with the tasks i have. certainly this weekend

atb00ker:
I recently [added a non-nullable field](https://github.com/openwisp/openwisp-radius/blob/8c88f1ad96f38c3fb792f0eefec887788aa6d37f/openwisp_radius/migrations/0011_radiustoken_freeradius_auth.py), but I had a privilege to delete old data, so, I can't be certain without trying but it maybe possible to it to move it to one migration file.

nemesisdesign:
@atb00ker one time I had an issue with a data migration that was being run in the same migration file of the migration which altered the DB. I discovered that it's better to run data migrations on their own.
We can run multiple data migration operations in a single migration file, but if there's a schema change, that must go on its own.

nemesisdesign:
@atb00ker It's in the [django docs of RunPython](https://docs.djangoproject.com/en/3.0/ref/migration-operations/#runpython):

PR #129, opened 2020-07-13
[enchancement] Add django-freeradius upgrader script

atb00ker:


docs/source/user/management_commands.rst
```diff
@@ -84,3 +84,30 @@ This command deletes users that have expired (and should have been deactivated b
     ./manage.py delete_old_users --older-than-months <duration_in_months>
 
 Note that the default duration is set to 18 months.
+
+``upgrade_from_django_freeradius``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you are upgrading from django-freeradius to openwisp-radius, there
```
nemesisdesign, line 91: please add a link to django-freeradius

atb00ker: The following is mentioned after this paragraph:

```
You can follow the `tutorial to migrate database from django-freeradius <https://github.com/openwisp/django-freeradius/blob/master/README.rst>`_.
```

However, I have added one here as well:

```
If you are upgrading from `django-freeradius <https://github.com/openwisp/django-freeradius>`_
```

PR #130, opened 2020-07-15
[api] add username in validate token api response

Vivekrajput20:


openwisp_radius/api/views.py
```diff
@@ -501,6 +501,7 @@ def post(self, request, *args, **kwargs):
                     'response_code': 'AUTH_TOKEN_VALIDATION_SUCCESSFUL',
                     'auth_token': token.key,
                     'radius_user_token': radius_token.key,
+                    'username': radius_token.user.username,
```
nemesisdesign, line 504: The queryset above should be changed:

```python
UserToken.objects.select_related('user').get(key=request_token)
```

Can you please add an assertion that checks for the presence of this attribute in the response?
https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/tests/test_api.py#L1401-L1407

docs/source/user/api.rst
```diff
@@ -383,7 +383,7 @@ Validate user auth token
 
 Used to check whether the auth token of a user is valid or not.
 
-Return also the radius user token in the response.
+Return also the radius user token and username in the response.
```
nemesisdesign, line 386: Let's change this to `It returns also the...`

Vivekrajput20:
@nemesisdesign  I've updated the PR.

Issue #91, opened 2020-06-04
[radius/api] Use bearer token for user authentication


nemesisdesign:
Use `openwisp_users.authentication.BearerAuthentication` instead of `rest_framework.authentication.TokenAuthentication`.

Update documentation to mention this.

This app will have to be upgraded to support this change: https://github.com/openwisp/openwisp-wifi-login-pages

Vivekrajput20:
@nemesisdesign  I think as long as authentication API is returning the new token, we will not have to upgrade  https://github.com/openwisp/openwisp-wifi-login-pages

nemesisdesign:
> @nemesisdesign I think as long as authentication API is returning the new token, we will not have to upgrade https://github.com/openwisp/openwisp-wifi-login-pages

@Vivekrajput20 thanks, I think we will have to update the authorization header, which now should be `Authorization: Token <token>` to `Authorization: Bearer <token>`

PR #131, opened 2020-07-17
[fix] BatchView permission and authorization #94

### LINKED ISSUE #94: [api/radius] Fix BatchView permission and authorization

atb00ker:
Closes #94

### DIFF
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -474,9 +474,11 @@ phone_number       string
 Batch user creation
 -------------------
 
+**Requires the user auth token (Bearer Token)**.
+
 .. code-block:: text
 
-    /api/v1/batch/
+    /api/v1/<organization-slug>/radiusbatch/
 
 .. note::
   This API endpoint allows to use the features described in :doc:`/user/importing_users`
--- a/openwisp_radius/api/urls.py
+++ b/openwisp_radius/api/urls.py
@@ -10,7 +10,6 @@ def get_api_urls(api_views=None):
         path('authorize/', api_views.authorize, name='authorize'),
         path('postauth/', api_views.postauth, name='postauth'),
         path('accounting/', api_views.accounting, name='accounting'),
-        path('batch/', api_views.batch, name='batch'),
         # registration differentiated by organization
         path('<slug:slug>/account/', api_views.register, name='rest_register'),
         # password reset
@@ -62,6 +61,7 @@ def get_api_urls(api_views=None):
             api_views.change_phone_number,
             name='phone_number_change',
         ),
+        path('<slug:slug>/radiusbatch/', api_views.batch, name='batch'),
         path(
             '<slug:slug>/radiusbatch/<uuid:radbatch>/pdf/',
             api_views.download_rad_batch_pdf,
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -42,6 +42,8 @@ from rest_framework.response import Response
 from rest_framework.throttling import BaseThrottle  # get_ident method
 from rest_framework.views import APIView
 
+from openwisp_users.api.authentication import BearerAuthentication
+
 from .. import settings as app_settings
 from ..exceptions import PhoneTokenException
 from ..utils import generate_pdf, load_model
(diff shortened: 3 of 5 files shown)


openwisp_radius/api/views.py
```diff
@@ -27,6 +27,7 @@
 from rest_framework import generics, serializers, status
 from rest_framework.authentication import BaseAuthentication, SessionAuthentication
 from rest_framework.authentication import TokenAuthentication as UserTokenAuthentication
+from openwisp_users.api.authentication import BearerAuthentication
```
atb00ker, line 30: BatchView uses `BearerAuthentication` but others use `UserTokenAuthentication`; I think all of them should be moved to `BearerAuthentication`, should I make a PR for it? :smile:

openwisp_radius/api/urls.py
```diff
@@ -62,6 +61,7 @@ def get_api_urls(api_views=None):
             api_views.change_phone_number,
             name='phone_number_change',
         ),
+        path('<slug:slug>/radiusbatch/', api_views.batch, name='batch'),
```
atb00ker, line 64: Note the url is `/radiusbatch/` instead of `/radius/batch/` suggested in the issue.
This is because the other radiusbatch download url also has `/radiusbatch/` and I wanted to maintain the pattern.
If you think we should change it either in here or both places, please let me know! :smile:

docs/source/user/api.rst
```diff
@@ -474,9 +474,11 @@ phone_number       string
 Batch user creation
 -------------------
 
+**Requires the user auth token (Bearer Token)**.
+
```
atb00ker, line 478: I have added this like other API urls, but I think it would be nice to include an example working curl request; I personally find them useful and I assume the user would as well.
Let me know if I should create an issue for adding sample curl requests in the documentation. :smile:

openwisp_radius/api/views.py
atb00ker: Nevermind, found https://github.com/openwisp/openwisp-radius/issues/91 ; I'll get to that as soon as possible! :smile:

openwisp_radius/api/urls.py
nemesisdesign: the change is fine to me, there are 2 issues:

- although adding `slug` here doesn't make much sense, because the other endpoints are designed to be used from the captive pages (or openwisp-wifi-login-pages). This endpoint is more like an API which could replace the django admin, so it's a bit different than the other ones.
- docs will need to be updated

openwisp_radius/api/views.py
```diff
@@ -326,13 +328,35 @@ def get_queryset(self):
...
-    authentication_classes = (TokenAuthentication,)
+class DispatchOrgMixin(object):
+    def dispatch(self, *args, **kwargs):
+        try:
+            self.organization = Organization.objects.get(slug=kwargs['slug'])
+        except Organization.DoesNotExist:
+            raise Http404()
+        return super().dispatch(*args, **kwargs)
+
+    def validate_membership(self, user):
```
nemesisdesign, line 339: I'm not sure if this is the right moment to do it, but take a look at https://github.com/openwisp/openwisp-users/issues/150
It would be great to solve that issue and create a general way for the API views of OpenWISP to do this kind of check.

openwisp_radius/api/urls.py
atb00ker: I think there is some confusion here:
1. `<slug:slug>/radiusbatch/` change is [mentioned in the issue](https://github.com/openwisp/openwisp-radius/issues/94), and when we [make a isOrganizationMember check in openwisp-users](https://github.com/openwisp/openwisp-users/issues/150) this would be required by that as well otherwise we will have to implement organization check for this endpoint separately. (my understanding is that `isOrganizationMember` is required regardless where the API is used when it's exposed to `staff/operator` users.) 
2. I have [already updated the API docs](https://github.com/openwisp/openwisp-radius/pull/131/files#diff-29397e8ff3afc883a7fb7f52c5a227e1R481), I don't see `/api/v1/batch/` anywhere else, am I missing any other file?

atb00ker:
Rebased! :smile:

openwisp_radius/api/urls.py
nemesisdesign: you're right! I have wrote that but realized it may not be the best solution.
However, I don't think it's wise to invest a lot more time in this now. So it's good enough, thank you!

PR #133, opened 2020-07-18
[api] Update last_login in ObtainAuthTokenView #112

### LINKED ISSUE #112: [radius/api] Update last login date in ObtainAuthTokenView
Update the last login date in ObtainAuthTokenView after successful authentication.

atb00ker:
Closes #112

### DIFF
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -11,6 +11,7 @@ from django.contrib.auth.tokens import default_token_generator
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.http import Http404, HttpResponse
+from django.utils import timezone
 from django.utils.decorators import method_decorator
 from django.utils.encoding import force_bytes, force_text
 from django.utils.http import urlsafe_base64_decode, urlsafe_base64_encode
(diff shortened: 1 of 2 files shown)


openwisp_radius/api/views.py
```diff
@@ -468,6 +469,8 @@ def post(self, request, *args, **kwargs):
         user = self.get_user(serializer, *args, **kwargs)
         token, created = UserToken.objects.get_or_create(user=user)
         radius_token = self.get_or_create_radius_token(user)
+        user.last_login = timezone.now()
+        user.save(update_fields=['last_login'])
```
nemesisdesign, line 473: I tried commenting out these two lines and the test passes anyway, can you investigate why?
I thought this wasn't being done by the token serializer but I may have been wrong

atb00ker: Amazing, lol, I'll check it out! :smile:

atb00ker: The testcase had a problem; fixed it.

openwisp_radius/api/views.py
```diff
@@ -502,6 +505,8 @@ def post(self, request, *args, **kwargs):
                     'auth_token': token.key,
                     'radius_user_token': radius_token.key,
                 }
+                token.user.last_login = timezone.now()
+                token.user.save(update_fields=['last_login'])
```
nemesisdesign, line 509: this code is repeated twice, please add a helper, eg:
```python
update_user_last_login(token.user)
```

openwisp_radius/api/views.py
nemesisdesign: from now on I expect you to do this on every test you write, that's how it should be done :blush:

PR #136, opened 2020-07-20
[bug] Fix API prefix add error #132

### LINKED ISSUE #132: [api] Batch add user with prefix method fails
I am getting:
```
django.db.utils.IntegrityError: FOREIGN KEY constraint failed
```
when I try to make request: 

```
curl -XPOST -H "Content-type: application/json" \
-H "Authorization: Bearer <org-uuid> <org-token>" \
     -d '{
            "name": "sample",
            "strategy": "prefix",
            "prefix": "sample",
            "number_of_users": 2,
            "expiration_date": "2021-02-02"
}' 'http://127.0.0.1:8000/api/v1/batch/'
```

The failing line is, `self.users.add(user)`:
```
    def save_user(self, user):
        OrganizationUser = swapper.load_model('openwisp_users', 'OrganizationUser')
        user.save()
        self.users.add(user)
````

This is odd because the same line with same input from admin interface works just fine.
Need to investigate this further.
Please let me know if this can be reproduced by someone else as well. :smile:

### PR DESCRIPTION
**Pull request by atb00ker**
Closes #132

### DIFF
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -476,7 +476,7 @@ Batch user creation
 
 .. code-block:: text
 
-    /api/v1/<organization-slug>/radiusbatch/
+    /api/v1/radiusbatch/
 
 .. note::
   This API endpoint allows to use the features described in :doc:`/user/importing_users`
@@ -488,26 +488,28 @@ It is possible to generate the users of the ``RadiusBatch`` with two different s
 
 The csv method needs the following parameters:
 
-===============    ===============================
+=================  =================================
 Param              Description
-===============    ===============================
+=================  =================================
 name               Name of the operation
 strategy           csv
 csvfile            file with the users
 expiration_date    date of expiration of the users
-===============    ===============================
+organization_slug  slug of organization of the users
+=================  =================================
 
 These others are for the prefix method:
 
-===============    ==================================
+=================  ==================================
 Param              Description
-===============    ==================================
+=================  ==================================
 name               name of the operation
 strategy           prefix
 prefix             prefix for the generation of users
 number_of_users    number of users
 expiration_date    date of expiration of the users
-===============    ==================================
+organization_slug  slug of organization of the users
+=================  ==================================
 
 When using this strategy, in the response you can find the field
 ``user_credentials`` containing the list of users created
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -23,6 +23,7 @@ RadiusPostAuth = load_model('RadiusPostAuth')
 RadiusAccounting = load_model('RadiusAccounting')
 RadiusBatch = load_model('RadiusBatch')
 OrganizationUser = swapper.load_model('openwisp_users', 'OrganizationUser')
+Organization = swapper.load_model('openwisp_users', 'Organization')
 User = get_user_model()
 
 
@@ -102,11 +103,23 @@ class UserSerializer(serializers.ModelSerializer):
         fields = '__all__'
 
 
+class RadiusOrganizationField(serializers.SlugRelatedField):
+    def get_queryset(self):
+        queryset = Organization.objects.all()
+        request = self.context.get('request', None)
+        if not request.user.is_superuser:
+            queryset = queryset.filter(pk__in=request.user.organizations_dict.keys())
+        return queryset
+
+
 class RadiusBatchSerializer(serializers.ModelSerializer):
+    organization = serializers.PrimaryKeyRelatedField(read_only=True)
+    organization_slug = RadiusOrganizationField(
+        required=True, label='organization', slug_field='slug', write_only=True
+    )
     users = UserSerializer(many=True, read_only=True)
     prefix = serializers.CharField(required=False)
     csvfile = serializers.FileField(required=False)
-    pdf = serializers.FileField(required=False, read_only=True)
     pdf_link = serializers.SerializerMethodField(required=False, read_only=True)
     number_of_users = serializers.IntegerField(
         required=False, write_only=True, min_value=1
@@ -128,11 +141,19 @@ class RadiusBatchSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError(
                 {'number_of_users': _('The field number_of_users cannot be empty')}
             )
-        return super().validate(data)
+        validated_data = super().validate(data)
+        # Additional Model Validation
+        batch_data = validated_data.copy()
+        batch_data.pop('number_of_users', None)
+        batch_data['organization'] = batch_data.pop('organization_slug', None)
+        instance = self.instance or self.Meta.model(**batch_data)
+        instance.full_clean()
+        return validated_data
 
     class Meta:
         model = RadiusBatch
         fields = '__all__'
+        read_only_fields = ('created', 'modified', 'user_credentials')
 
 
 class PasswordResetSerializer(BasePasswordResetSerializer):
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -328,6 +328,33 @@ class AccountingView(TokenAuthorizationMixin, ListCreateAPIView):
 accounting = AccountingView.as_view()
 
 
+class BatchView(CreateAPIView):
+    authentication_classes = (BearerAuthentication, SessionAuthentication)
+    permission_classes = (IsAdminUser, DjangoModelPermissions)
+    queryset = RadiusBatch.objects.all()
+    serializer_class = RadiusBatchSerializer
+
+    def post(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            valid_data = serializer.validated_data.copy()
+            num_of_users = valid_data.pop('number_of_users', None)
+            valid_data['organization'] = valid_data.pop('organization_slug', None)
+            batch = serializer.create(valid_data)
+            strategy = valid_data.get('strategy')
+            if strategy == 'csv':
+                batch.csvfile_upload()
+                response = RadiusBatchSerializer(batch, context={'request': request})
+            else:
+                batch.prefix_add(valid_data.get('prefix'), num_of_users)
+                response = RadiusBatchSerializer(batch, context={'request': request})
+            return Response(response.data, status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+batch = BatchView.as_view()
+
+
 class DispatchOrgMixin(object):
     def dispatch(self, *args, **kwargs):
         try:
@@ -344,53 +371,6 @@ class DispatchOrgMixin(object):
             )
             logger.warning(message)
             raise serializers.ValidationError({'non_field_errors': [message]})
-
-
-class BatchView(DispatchOrgMixin, CreateAPIView):
-    authentication_classes = (BearerAuthentication,)
-    permission_classes = (IsAdminUser, DjangoModelPermissions)
-    queryset = RadiusBatch.objects.all()
-    serializer_class = RadiusBatchSerializer
-
-    def post(self, request, *args, **kwargs):
-        self.validate_membership(request.user)
-        data = request.data.copy()
-        data.update(organization=self.organization.pk)
-        serializer = self.get_serializer(data=data)
-        if serializer.is_valid():
-            name = serializer.data.get('name')
-            expiration_date = serializer.data.get('expiration_data')
-            strategy = serializer.data.get('strategy')
-            if strategy == 'csv':
-                csvfile = request.FILES['csvfile']
-                options = dict(
-                    name=name,
-                    strategy=strategy,
-                    expiration_date=expiration_date,
-                    csvfile=csvfile,
-                    organization=self.organization,
-                )
-                batch = RadiusBatch(**options)
-                batch.csvfile_upload(csvfile)
-                response = RadiusBatchSerializer(batch)
-            elif strategy == 'prefix':
-                prefix = serializer.data.get('prefix')
-                options = dict(
-                    name=name,
-                    strategy=strategy,
-                    expiration_date=expiration_date,
-                    prefix=prefix,
-                    organization=self.organization,
-                )
-                batch = RadiusBatch(**options)
-                number_of_users = int(request.data['number_of_users'])
-                batch.prefix_add(prefix, number_of_users)
-                response = RadiusBatchSerializer(batch, context={'request': request})
-            return Response(response.data, status=status.HTTP_201_CREATED)
-        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
-
-
-batch = BatchView.as_view()
 
 
 class DownloadRadiusBatchPdfView(DispatchOrgMixin, APIView):
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -688,12 +688,12 @@ class TestApi(ApiTokenMixin, FileMixin, BaseTestCase):
         login_url = reverse('radius:user_auth_token', args=[self.default_org.slug])
         login_response = self.client.post(login_url, data=login_payload)
         header = f'Bearer {login_response.json()["key"]}'
-        url = reverse('radius:batch', args=[self.default_org])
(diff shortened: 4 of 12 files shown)

### CODE REVIEW

File: openwisp_radius/api/views.py [hunk 1]
```diff
@@ -359,6 +359,8 @@ def post(self, request, *args, **kwargs):
                     organization=Organization.objects.get(pk=self.request.auth),
                 )
                 batch = RadiusBatch(**options)
+                batch.full_clean()
+                batch.save()
```
atb00ker, line 363: It wasn't working because batch wasn't saved before it was used to create a ManyToMany relationship with users in `batch.prefix_add(prefix, number_of_users)` function.

But this raises more questions than it answers:
Why were the automated not able to find this error, that's what I'll have to do now, and fix tests accordingly!

atb00ker: Okay, I thought that the testcase was busted but that's not the case.
The error line: `self.users.add(user)` simply doesn't throw an error when used during testing and when request is made from admin.

Now I am in a situation where the patch works but I don't know why.
It's a dangerous situation, so please test it to see if you can replicate the `IntegrityError` when these lines are removed and API request to `BatchView` from prefix user generation is made.

openwisp_radius/api/views.py
```diff
@@ -359,6 +359,8 @@ def post(self, request, *args, **kwargs):
                     organization=Organization.objects.get(pk=self.request.auth),
                 )
                 batch = RadiusBatch(**options)
+                batch.full_clean()
```
nemesisdesign, line 362: RadiusBatchSerializer extends ModelSerializer from openwisp-utils, which calls the full_clean during validation, so I suspect this is redundant.

atb00ker:
To reproduce the issue, just make a request to batchview API for generating users, example:

```
curl -X POST -H "Content-type: application/json" \
     -H "Authorization: Bearer <key>" \
     -d '{
          "name": "sample2",
          "strategy": "prefix",
          "prefix": "sample2",
          "number_of_users": 2,
          "expiration_date": "2021-02-05"
     }' 'http://127.0.0.1:8000/api/v1/default/radiusbatch/'
```

atb00ker:
I have not found the reason for the original problem. To reproduce the same please, please see [the comment above](https://github.com/openwisp/openwisp-radius/pull/136#issuecomment-662265514).

However, I have moved the radius batch creation to serializer in the `create()`, I create the radiusbatch which takes care of the `save()` problem and solves the bug.

openwisp_radius/api/serializers.py
```diff
@@ -128,8 +134,47 @@ def validate(self, data):
             raise serializers.ValidationError(
                 {'number_of_users': 'The field number_of_users cannot be empty'}
             )
+        try:
+            self.batch_org = Organization.objects.get(slug=data['organization_slug'])
```
atb00ker, line 138: I think this maybe a little unconventional but I save `self.batch_org` to avoid making another request to the database in the `.create()` function.
Is it fine or should I remove this variable and make another request in the `.create()`?

nemesisdesign: question: aren't you already retrieving the organization from validating permissions?
In any case, this should be done in the view, we can access the view instance from the serializer.
See https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/serializers.py#L159, the org is already retrieve in the view so the serializer just gets it from the view.
Look also for the concept of "serializer context" in DRF.

openwisp_radius/api/serializers.py
```diff
@@ -126,10 +132,39 @@ def get_pdf_link(self, obj):
...
+            raise serializers.ValidationError(
+                {'organization_slug': 'Organization not found.'}
+            )
+        try:
+            RadiusBatch.objects.get(name=data['name'], organization=self.batch_org)
+        except RadiusBatch.DoesNotExist:
+            pass
+        else:
+            raise serializers.ValidationError(
+                {'name': 'RadiusBatch with same name exists for organization.'}
```
nemesisdesign, line 149: Look at this: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/base/models.py#L844-L865

Can we reuse that validation logic in this serializer? 

Maybe you don't know it, but DRF at some point changed the validation and stopped calling full_clean(), the argument is that API validation can be different from model validation.
I never loved this, and that's why we have [ValidatedModelSerializer in openwisp-utils](https://github.com/openwisp/openwisp-utils#openwisp-utils-api-serializers-validatedmodelserializer). Can you try adding it to this serializer? It could be a good chance to remove duplication and add more consistency to the codebase (ideally we should have all serializers based on `ValidatedModelSerializer` but we can look into that later on).

openwisp_radius/api/serializers.py
```diff
@@ -126,10 +132,39 @@ def get_pdf_link(self, obj):
...
+            strategy=strategy,
+            expiration_date=validated_data.get('expiration_date'),
+            organization=self.batch_org,
+        )
+        if strategy == 'csv':
+            options.update(csvfile=validated_data.get('csvfile'))
+            return RadiusBatch.objects.create(**options)
+        elif strategy == 'prefix':
+            options.update(prefix=validated_data.get('prefix'))
+            return RadiusBatch.objects.create(**options)
```
nemesisdesign, line 166: This could be:
```python
if strategy == 'csv':
    options.update(csvfile=validated_data.get('csvfile'))
elif strategy == 'prefix':
    options.update(prefix=validated_data.get('prefix'))
return RadiusBatch.objects.create(**options)
```

But why wouldn't the default create method work out of the box without the need of being overridden?
What is the mandatory step that has to be performed?
Adding the org? is the if really necessary?

openwisp_radius/api/serializers.py
nemesisdesign: PS: you could also add the organization to the data, eg:

```
data['organization'] = self.context['view'].organization
```
which would make the method below leaner.

atb00ker: > question: aren't you already retrieving the organization from validating permissions?

Yes, but as discussed in the [BatchView permissions PR](https://github.com/openwisp/openwisp-radius/pull/131#discussion_r457825262) we don't want slug in the url and hence, I am planning to get rid of it with the [openwisp-users change PR (coming soon)](https://github.com/openwisp/openwisp-users/issues/150)

> In any case, this should be done in the view, we can access the view instance from the serializer.

Sweet, I thought it needs to be done in serializer, moving it in view would be better, I am looking into it now!

openwisp_radius/tests/test_api.py
```diff
@@ -989,7 +989,7 @@ def test_batch_csv_pdf_link_404(self):
             response = self._radius_batch_post_request(data)
         self.assertEqual(response.status_code, 201)
         response_json = json.loads(response.content)
-        org = Organization.objects.get(pk=response_json['organization'])
+        org = Organization.objects.get(slug=response_json['organization'])
```
atb00ker, line 992: The API returns organzation slug instead of pk because I think that should be more user friendly, what do you think?

atb00ker:
> looking better, so you weren't able to reproduce this bug in tests at all?

No! :disappointed: 

Sorry, you'll have to review all the files again from starting, I found a lot of problems and changed everything!

openwisp_radius/api/serializers.py
```diff
@@ -102,15 +105,15 @@ class Meta:
...
+        queryset=Organization.objects.all(), slug_field='slug',
+    )
     users = UserSerializer(many=True, read_only=True)
     prefix = serializers.CharField(required=False)
     csvfile = serializers.FileField(required=False)
     pdf = serializers.FileField(required=False, read_only=True)
     pdf_link = serializers.SerializerMethodField(required=False, read_only=True)
-    number_of_users = serializers.IntegerField(
-        required=False, write_only=True, min_value=1
-    )
```
atb00ker, line 113: To use `ValidatedModelSerializer`, I need to remove `number_of_users`; However, I don't think `full_clean()` is required here, all the values are checked by serializer now. What do you think?

docs/source/user/api.rst
```diff
@@ -476,7 +476,7 @@ Batch user creation
 
 .. code-block:: text
 
-    /api/v1/<organization-slug>/radiusbatch/
```
atb00ker, line 479: As discussed in the [BatchView permissions PR](https://github.com/openwisp/openwisp-radius/pull/131#discussion_r457825262) we don't want slug in the URL and hence, I am getting rid of it.

P.S after spending some time on DRF docs & stackoverflow, I realized it's best practice to do this validation in the serializer itself, so this check is done in `RadiusOrganizationField`. :smile:

nemesisdesign: nice thanks

docs/source/user/api.rst
```diff
@@ -476,7 +476,7 @@ Batch user creation
...
+    /api/v1/radiusbatch/add/
```
nemesisdesign, line 479: It should be just `/api/v1/radiusbatch/`, posting to a resource implies adding an object as per REST convention

openwisp_radius/api/serializers.py
```diff
@@ -102,7 +103,20 @@ class Meta:
         fields = '__all__'
 
 
+class RadiusOrganizationField(serializers.SlugRelatedField):
+    def get_queryset(self):
+        queryset = Organization.objects.all()
+        request = self.context.get('request', None)
+        if not request.user.is_superuser:
+            queryset = queryset.filter(pk__in=request.user.organizations_dict)
```
nemesisdesign, line 111: shouldn't this be `request.user.organizations_dict.keys()`?

openwisp_radius/api/serializers.py
```diff
@@ -102,7 +103,20 @@ class Meta:
...
+        return queryset
+
+    def to_representation(self, obj):
+        return getattr(obj, 'pk')
+
+
 class RadiusBatchSerializer(serializers.ModelSerializer):
+    organization = RadiusOrganizationField(slug_field='slug')
```
nemesisdesign, line 119: I think using the uuid was fine since we have all API endpoints doing that, if we do this change only here the API will be inconsistent, have you tought about this? I don't know if changing this for all the endpoints is the right thing to do now (we'de need to double check select_related is used when listing objects otherwise we'd generate 1 query for each object in the list for each API endpoint, that's my main concern)

openwisp_radius/api/serializers.py
atb00ker: It works either way, filter only looks at the keys. :smile:
I'll add `keys()` to help code be more readable / obvious in what it is doing! :smile:

openwisp_radius/api/serializers.py
atb00ker: I think It's consistent with the other endpoints. 
All the APIs are taking organization input slug of the organization from the URL and the output is `uuid`; I have maintained the same, input is `slug` and output contains the `uuid`. Previously for this endpoint as well, it was the same, input was `slug` from URL and output for `pk`; I have only moved the slug from URL to post data.

Do you think it'll be better to use `pk` in post data?

nemesisdesign: Thanks for the explanation, that's fine

openwisp_radius/api/views.py
```diff
@@ -341,45 +341,24 @@ def validate_membership(self, user):
             raise serializers.ValidationError({'non_field_errors': [message]})
 
 
-class BatchView(DispatchOrgMixin, generics.CreateAPIView):
+class BatchView(generics.CreateAPIView):
     authentication_classes = (BearerAuthentication,)
```
nemesisdesign, line 345: let's add `SessionAuthentication` it makes it easier to test (it's safe to add this authentication class to all endpoints which make use of `BearerAuthentication`.

I tried this, the Browsable API POST form shows up, and I noticed two problems:

- created and modified are shown in the POST form, but these fields should not be displayed (should be readonly)
- I tried posting some data and I got this:

```json
{
    "organization": [
        "Object with slug=342ad600-3ec7-46f2-9e7d-9af1dbd93205 does not exist."
    ]
}
```

atb00ker: I was using `curl` to make requests, the form showed some more errors, fixed them all.

openwisp_radius/api/serializers.py
```diff
@@ -128,11 +141,20 @@ def validate(self, data):
...
             )
-        return super().validate(data)
+        validated_data = super().validate(data)
+        # Additional Model Validation
+        batch_data = data.copy()
+        batch_data.pop('number_of_users', None)
+        slug = batch_data.pop('organization_slug')
+        batch_data.update(organization=Organization.objects.get(slug=slug))
+        instance = self.instance or self.Meta.model(**batch_data)
+        instance.full_clean()
```
atb00ker, line 151: While debugging on the web API interface, I noticed another issue that I didn't think about before, so I have added a model check with some modification. :smile:

openwisp_radius/api/serializers.py
```diff
@@ -102,11 +103,23 @@ class Meta:
...
+        if not request.user.is_superuser:
+            queryset = queryset.filter(pk__in=request.user.organizations_dict.keys())
+        return queryset
+
+
 class RadiusBatchSerializer(serializers.ModelSerializer):
+    organization = serializers.PrimaryKeyRelatedField(read_only=True)
+    organization_slug = RadiusOrganizationField(
+        required=True, label='organization', slug_field='slug', write_only=True
+    )
```
atb00ker, line 119: `organization_slug` is used for input and `organization` is used for returning `pk` of organization.

openwisp_radius/api/views.py
```diff
@@ -346,45 +346,25 @@ def validate_membership(self, user):
             raise serializers.ValidationError({'non_field_errors': [message]})
 
 
-class BatchView(DispatchOrgMixin, CreateAPIView):
-    authentication_classes = (BearerAuthentication,)
+class BatchView(CreateAPIView):
+    authentication_classes = (BearerAuthentication, SessionAuthentication)
```
atb00ker, line 350: thanks, adding `SessionAuthentication` brought forth more issues, I solved them, I'll use Web API interface in the future! :smile:

nemesisdesign: That's the way to go Ajay :+1:

openwisp_radius/api/serializers.py
```diff
@@ -128,11 +141,20 @@ def validate(self, data):
             raise serializers.ValidationError(
                 {'number_of_users': _('The field number_of_users cannot be empty')}
             )
-        return super().validate(data)
+        validated_data = super().validate(data)
+        # Additional Model Validation
+        batch_data = data.copy()
+        batch_data.pop('number_of_users', None)
+        slug = batch_data.pop('organization_slug')
+        batch_data.update(organization=Organization.objects.get(slug=slug))
```
nemesisdesign, line 149: looks like this raises an exception if the supplied value is not found. Ensure it's resilient by setting it to None if the org is not found so that it raises a validation error (needs an automated test)

atb00ker: @nemesisdesign I tried a couple of things, including:

1. Sending non-existing
2. Sending slug of org to which the user doesn't have the access.
3. Sending invalid slug 

In everycase, serializer validation caches the problem even before it reaches the errored line and now that I think about it,
I remember that didn't bother with try/catch for that `batch_data.update(organization=Organization.objects.get(slug=org_slug))` because in normal flow this cannot be empty/incorrect , because `organization_slug` value is checked by serializer validators before this function is called. 

1. Can you please confirm you are indeed in the latest of the PR?
2. Provide information about anything special / different about your instance.
3. Check what your organzation_slug is getting and if that is a valid slug in your database.

atb00ker: Since `queryset.filter(pk__in=request.user.organizations_dict.keys())` shows it in the drop down and serializer validators don't raise an error, I am reasonable certain that `organization_slug` should not contain an invalid value.

atb00ker:
Yes! Thank you for the information, I am able to reproduce it now.. on it! :smile:

openwisp_radius/api/serializers.py
atb00ker: Fixed

atb00ker:
Fixed the problem and made changes in tests to ensure such error is captured by automated tests.
However, because the changes in the tests was not 100% relevant to this PR, hence I have made another commit for it, let me know if I should move these changes to another PR.

PR #137, opened 2020-07-22
[api] Use permissions classes from openwisp-users

atb00ker:


.travis.yml
```diff
@@ -19,8 +19,10 @@ install:
   - python setup.py -q develop
   - pip install -r requirements-test.txt
   - npm install -g jslint
+  # TODO:N Remove before merging
+  - pip install --upgrade https://github.com/atb00ker/openwisp-users/tarball/issues/150-org-member-validator/
   # TODO: Remove when openwisp-users 0.3 is released
-  - pip install --upgrade https://github.com/openwisp/openwisp-users/tarball/master/
+  # - pip install --upgrade https://github.com/openwisp/openwisp-users/tarball/master/
```
atb00ker, line 25: Needs to be fixed with before merge.

nemesisdesign: @atb00ker let's update this

openwisp_radius/api/views.py
```diff
@@ -409,21 +411,15 @@ def validate_membership(self, user):
             raise serializers.ValidationError({'non_field_errors': [message]})
 
 
-class DownloadRadiusBatchPdfView(DispatchOrgMixin, APIView):
+class DownloadRadiusBatchPdfView(DispatchOrgMixin, RetrieveAPIView):
     authentication_classes = (BearerAuthentication, SessionAuthentication)
-    permission_classes = (IsAdminUser, DjangoModelPermissions)
+    permission_classes = (IsOrganizationMember, IsAdminUser, DjangoModelPermissions)
```
nemesisdesign, line 416: didn't spot this before, I think we have to use IsOrganizationManager, because this kind of functionality is aimed at administrators, not end-users (public wifi users).

atb00ker: I was under the assumption that all operators are not necessarily `is_admin` for all organizations.
I have fixed it.

nemesisdesign: your assumption is correct, one operator will be manager of organization A but may be end-user in organization B, therefore it must not be able to access the radius batch API of organization B.

PR #138, opened 2020-07-22
[enchancement] celery-beat for management commands #117

### LINKED ISSUE #117: [radius/management] Add celery beat tasks to replace management commands
In a high availability scenario using multiple containers, setting up the management commands to run in a cronjob would result in the commands being executed multiple times.

For this reason we should provide the same commands as celery tasks that can be set up with celery beat.

atb00ker:
Closes #117

### DIFF
--- a/docs/source/developer/how_to_extend.rst
+++ b/docs/source/developer/how_to_extend.rst
@@ -302,8 +302,32 @@ monkey patching, you can proceed as follows:
     class RadiusBatchAdmin(BaseRadiusBatchAdmin):
         # add your changes here
 
-11. Create root URL configuration
+11. Setup Periodic tasks
+------------------------
+
+Some periodic commands are required in production environments to enable certain
+features and facilitate database cleanup:
+
+1. You need to create a `celery configuration file as it's created in example file <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/celery.py>`_.
+
+2. In the settings.py, `configure the CELERY_BEAT_SCHEDULE <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/settings.py#L141>`_. Some celery tasks take an argument, for instance
+``365`` is given here for ``delete_old_radacct`` in the example settings.
+These arguments are passed to their respective management commands. More information about these parameters can be
+found at the `management commands page </user/management_commands.html>`_.
+
+3. Add the following in your settings.py file::
+
+    CELERY_IMPORTS = ('openwisp_monitoring.device.tasks',)
+
+.. note::
+    Celery tasks do not start with django server and need to be
+    started seperately, please read about running `celery and
+    celery-beat </developer/setup.html#celery-usage>`_ tasks.
+
+12. Create root URL configuration
 ---------------------------------
+
+The root ``url.py`` file should have the following paths (please read the comments):
 
 .. code-block:: python
 
@@ -325,7 +349,7 @@ monkey patching, you can proceed as follows:
     For more information about URL configuration in django, please refer to the
     `"URL dispatcher" section in the django documentation <https://docs.djangoproject.com/en/dev/topics/http/urls/>`_.
 
-12. Import the automated tests
+13. Import the automated tests
 ------------------------------
 
 When developing a custom application based on this module, it's a good
--- a/docs/source/developer/setup.rst
+++ b/docs/source/developer/setup.rst
@@ -150,6 +150,84 @@ be able to use it with openwisp-radius (and extended apps) easily:
     ./manage.py migrate --fake openwisp-radius 0001_initial_freeradius
     ./manage.py migrate
 
+Automated periodic tasks
+------------------------
+
+Some periodic commands are required in production environments to enable certain
+features and facilitate database cleanup.
+There are two ways to automate these tasks:
+
+1. Celery-beat (Recommended Method)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. You need to create a `celery configuration file as it's created in example file <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/celery.py>`_.
+
+2. Add celery to ``__init__.py`` of your project:
+
+.. code-block:: python
+
+    from .celery import app as celery_app
+
+    __all__ = ['celery_app']
+
+3. In the settings.py, `configure the CELERY_BEAT_SCHEDULE <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/settings.py#L141>`_. Some celery tasks take an argument, for instance
+``365`` is given here for ``delete_old_radacct`` in the example settings.
+These arguments are passed to their respective management commands. More information about these parameters can be
+found at the `management commands page </user/management_commands.html>`_.
+
+.. note::
+    Celery tasks do not start with django server and need to be
+    started seperately, please read about running `celery and
+    celery-beat </developer/setup.html#celery-usage>`_ tasks.
+
+2. Crontab (Legacy Method)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Edit the crontab with:
+
+.. code-block:: shell
+
+    crontab -e
+
+Add and modify the following lines accordingly:
+
+.. code-block:: shell
+
+    # This command deletes RADIUS accounting sessions older than 365 days
+    30 04 * * * <virtualenv_path>/bin/python <full/path/to>/manage.py delete_old_radacct 365
+
+    # This command deletes RADIUS post-auth logs older than 365 days
+    30 04 * * * <virtualenv_path>/bin/python <full/path/to>/manage.py delete_old_postauth 365
+
+    # This command closes stale RADIUS sessions that have remained open for 15 days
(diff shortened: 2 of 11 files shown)


tests/openwisp2/celery.py
```diff
@@ -0,0 +1,31 @@
...
+        delete_old_users,
+    )
+
+    # You can configure parameters here to change when the commands will be
+    # automatically executed just like crontab.
+    sender.add_periodic_task(crontab(hour=0, minute=0), deactivate_expired_users)
+    sender.add_periodic_task(crontab(hour=0, minute=10), delete_old_users.s(365))
+    sender.add_periodic_task(crontab(hour=0, minute=20), cleanup_stale_radacct.s(365))
+    sender.add_periodic_task(crontab(hour=0, minute=30), delete_old_postauth.s(365))
+    sender.add_periodic_task(crontab(hour=0, minute=40), delete_old_radacct.s(365))
```
nemesisdesign, line 31: can we keep consistency with [openwisp-monitoring and use CELERY_BEAT_SCHEDULE](https://github.com/openwisp/openwisp-monitoring/blob/master/tests/openwisp2/settings.py#L165-L172) please?

tests/openwisp2/settings.py
```diff
@@ -123,10 +123,20 @@
...
+    CELERY_BROKER_URL = os.getenv('REDIS_URL', f'redis://{redis_host}/1')
+else:
     OPENWISP_RADIUS_GROUPCHECK_ADMIN = True
     OPENWISP_RADIUS_GROUPREPLY_ADMIN = True
     OPENWISP_RADIUS_USERGROUP_ADMIN = True
+    CELERY_TASK_ALWAYS_EAGER = True
+    CELERY_TASK_EAGER_PROPAGATES = True
+    CELERY_BROKER_URL = 'memory://'
+
+TEST_RUNNER = 'openwisp_utils.tests.TimeLoggingTestRunner'
```
nemesisdesign, line 138: :+1:

tests/openwisp2/settings.py
```diff
@@ -123,10 +123,20 @@
...
+
```
nemesisdesign, line 139: one too much extra blank line maybe?

openwisp_radius/tests/test_celery.py
```diff
@@ -0,0 +1,79 @@
+from celery import Celery
```
nemesisdesign, line 1: in the other modules we use "test_tasks", can you follow this style please?
https://github.com/openwisp/openwisp-firmware-upgrader/blob/master/openwisp_firmware_upgrader/tests/test_tasks.py

openwisp_radius/tests/test_celery.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class TestCelery(FileMixin, BaseTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        app = Celery('openwisp2')
+        app.config_from_object('django.conf:settings', namespace='CELERY')
+        app.autodiscover_tasks()
+        cls.celery_worker = start_worker(app)
```
nemesisdesign, line 24: are you starting the worker here? We aren't doing this in the other module but I want to understand more about it.
Can you explain a bit more what this test class does? Adding a few comments would be very useful.

tests/openwisp2/celery.py
```diff
@@ -0,0 +1,11 @@
+import os
```
atb00ker, line 1: This file can moved to `openwisp_radius.celery` and imported in `__init__.py` from there, I don't know why but the same is not done in the openwisp-controller and openwisp-monitoring.

The `__init__.py` can have `os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'openwisp2.settings')`.

Is the reason for `celery.py` to be outside the app is that we don't want `DJANGO_SETTINGS_MODULE` in `__init__.py` or is there another reason for it? :smile:

openwisp_radius/tests/test_tasks.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class TestCelery(FileMixin, BaseTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        app = Celery('openwisp2')
+        app.config_from_object('django.conf:settings', namespace='CELERY')
+        app.autodiscover_tasks()
+        cls.celery_worker = start_worker(app)
```
atb00ker, line 24: The tests here done are a little different than the ones in other modules, in the celery docs, I read that ideally, celery should only call tasks and shouldn't process any data.

> A Celery task is much like a web view, in that it should only define how to perform the action in the context of being called as a task.

This makes celery testing not required in most modules, however, since I wanted to test the calling of celery tasks, I created an instance of celery from instance the test class to ensure that celery is in the same process as tests.

This helps me avoid mocking. 
With mocking, I couldn't have tested the calls rather I would be testing management command operations only which are tested in `test_management` anyway.

What do you think about this approach?

openwisp_radius/tests/test_commands.py
```diff
@@ -42,11 +28,9 @@ def test_cleanup_stale_radacct_command(self):
         self.assertEqual(session.update_time, session.stop_time)
 
     def test_delete_old_postauth_command(self):
-        options = dict(username='steve', password='jones', reply='ghdhd')
+        options = dict(username='steve', password='jones', reply='value1')
```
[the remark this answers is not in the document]
nemesisdesign: that's a lot better, thank you

tests/openwisp2/celery.py
nemesisdesign: I think we're just following the guidelines of celery when integrating in django, if you find different guidelines let me know

openwisp_radius/tests/test_tasks.py
nemesisdesign: Looks good, I think the closer we get to replicating real world usage, the better, so your effort here seems to go in the right direction, but I'll be able to confirm this to you only in the future once we try to replicate this in other modules and find out if it actually catches more bugs than the approach we follow currently. For the moment, it's good, let's try it.

docs/source/developer/setup.rst
```diff
@@ -150,6 +150,96 @@ be able to use it with openwisp-radius (and openwisp-radius) easily:
...
+In the settings.py, configure the ``CELERY_BEAT_SCHEDULE`` as `done in the example <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/settings.py#L141>`_.
+
+.. note::
+    Some celery tasks take an argument, example ``365`` is given here for
+    ``delete_old_radacct``. These arguments are passed to their respective management
+    commands. More information about these parameters can be
+    found at the `management commands page </user/management_commands.html>`_.
+
+Next, `install docker-compose <https://docs.docker.com/compose/install/>`_
+and start redis-server inside docker:
```
nemesisdesign, line 182: it looks like running docker compose is required to use this feature, I think we should make it explicit that this is a suggestion, only a redis instance is required, the user may already have a cluster and will surely wonder if they can reuse the cluster they already have.

docs/source/developer/setup.rst
```diff
@@ -150,6 +150,96 @@ be able to use it with openwisp-radius (and openwisp-radius) easily:
     ./manage.py migrate --fake openwisp-radius 0001_initial_freeradius
     ./manage.py migrate
 
+Automated periodic tasks
+------------------------
+
+Some periodic commands are necessary to enable certain
+features and facilitate database cleanup.
+There are two ways to automate these tasks:
```
nemesisdesign, line 158: let's clarify that these steps are needed for production usage, in development and testing this is not really required

docs/source/developer/setup.rst
atb00ker: Fixed this. 
And removed the docker-compose file and moved to docker only, since docker-compose maybe overkill for using just one container.

docs/source/developer/how_to_extend.rst
```diff
@@ -302,9 +302,33 @@ monkey patching, you can proceed as follows:
...
+
+Some periodic commands are required in production environments to enable certain
+features and facilitate database cleanup:
+
+1. You need to create a `celery configuration file as it's created in example file <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/celery.py>`_.
+
+2. In the settings.py, `configure the CELERY_BEAT_SCHEDULE <https://github.com/openwisp/openwisp-radius/tree/master/tests/openwisp2/settings.py#L141>`_. Some celery tasks take an argument, for instance
+``365`` is given here for ``delete_old_radacct`` in the example settings.
+These arguments are passed to their respective management commands. More information about these parameters can be
+found at the `management commands page </user/management_commands.html>`_.
```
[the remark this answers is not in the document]
nemesisdesign: not optimal for us maintainers but good for the user, let's see how it plays out over time

docs/source/developer/how_to_extend.rst
```diff
@@ -302,9 +302,33 @@ monkey patching, you can proceed as follows:
...
+.. note::
+    Celery tasks do not start with django server and need to be
+    started seperately, please read about running `celery and
+    celery-beat </developer/setup.html#celery-usage>`_ tasks.
+
+12. Create root URL configuration
 ---------------------------------
 
+The root ``url.py`` file should have the following paths (please read the comments):
+
```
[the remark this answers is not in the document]
nemesisdesign: ok

PR #139, opened 2020-07-23
[api] Use BearerAuthentication #91

### LINKED ISSUE #91: [radius/api] Use bearer token for user authentication

atb00ker:
Closes #91

### DIFF
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -248,8 +248,13 @@ obtaining access tokens, validating their phone number, etc.).
   `Organization API Token <#organization-api-token>`_
   described in the beginning of this document.
 
-  Some of them require the sending of the user API access token
-  sent in the form of a "Bearer Token".
+Some endpoints require the sending of the user API access
+token sent in the form of a "Bearer Token", example:
+
+.. code-block:: shell
+
+    curl -H "Authorization: Bearer <user-token>" \
+         'http://localhost:8000/api/v1/default/account/session/'
 
 User Registration
 -----------------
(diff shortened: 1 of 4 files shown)


docs/source/user/api.rst
```diff
@@ -248,8 +248,13 @@ obtaining access tokens, validating their phone number, etc.).
...
 
-  Some of them require the sending of the user API access token
-  sent in the form of a "Bearer Token".
+Some endpoints require the sending of the user API access
+token sent in the form of a "Bearer Token", example:
+
+.. code-block:: shell
+
+    curl -H "Authorization: Bearer <user-token>" \
+         'http://localhost:8000/api/v1/default/account/session/'
```
atb00ker, line 257: Removed example from below and added it before all User APIs as the example to use `Bearer <user-token>` is applicable for multiple User APIs

openwisp_radius/api/views.py
```diff
@@ -126,7 +125,6 @@ def get_serializer(self, *args, **kwargs):
 
 
 class AuthorizeView(TokenAuthorizationMixin, APIView):
-    authentication_classes = (TokenAuthentication,)
```
atb00ker, line 129: Removed from here because `TokenAuthorizationMixin` already contains authenication_classes`

PR #140, opened 2020-07-27
[minor] Clean up

atb00ker:
Todo: 
- [x] Use `f-string` where possible.
- [x] Logger translate
- [x] "


openwisp_radius/urls.py
```diff
@@ -25,7 +25,7 @@ def get_urls(api_views=None, social_views=None):
     if app_settings.SOCIAL_LOGIN_ENABLED:
         urls.append(
             path(
-                'freeradius/social-login/<slug:slug>/',
+                'radius/social-login/<slug:slug>/',
```
atb00ker, line 28: Backwards incompatible alert.

PR #142, opened 2020-07-29
[minor] Improves sample-radius migrations

atb00ker:


atb00ker:
openwisp_radius's migrations already had [0001_initial_freeradius for freeradius 3.0 compatibility](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/migrations/0001_initial_freeradius.py).

I didn't think that `sample_app` would need it but now I think it should have it as well because if user is extending openwisp_radius and they want this feature, they can simply refer the sample_radius's migration! :smile:

I have implemented the same for sample_app! :smile:

PR #145, opened 2020-08-06
[feature] Restrict freeradius API usage with ip address

### LINKED ISSUE #96: [radius/api] Restrict IP addresses allowed to use freeradius API endpoints
To be done after #88.

There must be a way to restrict the IP addresses which can consume the API endpoints `authorize`, `accounting` and `postauth`.

I would add this at organization level using OrganizationRadiusSettings, using something like a string field: a comma separated string with the ip-addresses (the string will have to be validated).

I would add a setting which defines the default allowed IP addresses to consume the freeradius API, defaulting to empty list.

If the new setting is not empty, the model field should be allowed to be left blank, and the IP addresses from the setting should be used.

If the new setting is empty, the model field shall not be allowed to be blank.

If the model field is filled, it will override what's in the settings.

We need to document this.

atb00ker:
Closes #96


openwisp_radius/base/models.py
```diff
@@ -1034,7 +1035,14 @@ class AbstractOrganizationRadiusSettings(UUIDModel):
         null=True,
         blank=True,
         help_text=_(
-            'Additional configuration for SMS backend in JSON format, if needed'
+            'Additional configuration for SMS backend in JSON format (optional)'
```
atb00ker, line 1038: I think `Additional configuration for SMS backend in JSON format, if needed` -> `Additional configuration for SMS backend in JSON format (optional)` is more clear and I was creating a migration, so I added this, let me know if I should revert it.

.travis.yml
```diff
@@ -16,6 +16,8 @@ branches:
 
 install:
   - pip install $DJANGO
+  # Testing, normal installation has issues
+  - pip install django-ipware
```
atb00ker, line 20: See: https://github.com/un33k/django-ipware/issues/57

This has been bothering me throughout the day,
It's most probably because a `.pyc` file made it to the release of django-ipware.
Hopefully, they will re-release/fix it soon! :smile:

atb00ker: Fixed

docs/source/user/settings.rst
```diff
@@ -172,6 +172,21 @@ the following to your project ``settings.py``:
...
+``OPENWISP_RADIUS_FREERADIUS_ALLOWED_HOSTS``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Default**: ``[]``
+
+List of host IP addresses allowed to consume the freeradius
+API endpoints, i.e Authorize, Accounting and Postauth.
+
+The organization administrators can override this setting in
+their organization dashboard.
```
nemesisdesign, line 184: I would change this to "This value can be overridden per organization in the organization change page" or something similar.

openwisp_radius/api/views.py
```diff
@@ -84,13 +85,33 @@
 
 
 class FreeradiusApiAuthentication(BaseAuthentication):
+    def _check_client_ip_and_return(self, request, uuid):
+        client_ip, _is_routable = get_client_ip(request)
+        try:
+            ip_addresses = OrganizationRadiusSettings.objects.get(
```
nemesisdesign, line 91: we have to cache this to avoid repeating this query for each request

openwisp_radius/api/views.py
```diff
@@ -84,13 +85,33 @@
...
+                organization__pk=uuid
+            ).freeradius_allowed_hosts
+            allowed_ip_list = ip_addresses.split(',') if ip_addresses else None
```
nemesisdesign, line 94: why not create a model method (even with a `@property`) which returns the right value ready for usage?

Untested hint:

```python
@property
def freeradius_allowed_hosts_list(self):
    addresses = self.freeradius_allowed_hosts
    if not addresses:
        return app_settings.FREERADIUS_ALLOWED_HOSTS
    return self.freeradius_allowed_hosts.split(',')
```

openwisp_radius/api/views.py
```diff
@@ -84,13 +85,42 @@
...
+        client_ip, _is_routable = get_client_ip(request)
+        ip_list = self._get_ip_list(uuid)
+        if bool(
+            (ip_list and client_ip in ip_list)
+            or (not ip_list and client_ip in app_settings.FREERADIUS_ALLOWED_HOSTS)
+        ):
+            return (AnonymousUser(), uuid)
+        message = _(
+            f'Request rejected: Client IP address ({client_ip}) '
+            'is not allowed by the organization admin for freeradius API.'
```
nemesisdesign, line 112: I would change this to:

`is not in the list of IP addresses allowed to consume the freeradius API`

@morrme any useful suggestion?

openwisp_radius/base/models.py
```diff
@@ -1055,15 +1070,38 @@ def clean(self):
                     )
                 }
             )
+        # Clean `freeradius_allowed_hosts`
+        iplist = self.freeradius_allowed_hosts_list
```
nemesisdesign, line 1074: I'd change this var to `ip_list` for consistency with `ip_addr` below (which you could rename `ip_address` since it's not very long and help readability, both suggested changes improve readability.

openwisp_radius/tests/test_api.py
```diff
@@ -1597,6 +1597,83 @@ def test_user_auth_updates_last_login(self):
...
+
+    def test_ip_from_radsetting_invalid(self):
+        radsetting = OrganizationRadiusSettings.objects.get(
+            organization=self._get_org()
+        )
+        radsetting.freeradius_allowed_hosts = '127.0.0.500'
+        radsetting.save()
+        with mock.patch(self.freeradius_hosts_path, []):
+            response = self.client.post(reverse('radius:authorize'), self.params)
+        self.assertEqual(response.data['detail'], self.fail_msg)
```
nemesisdesign, line 1663: please assert also the status code of the response

openwisp_radius/tests/test_api.py
```diff
@@ -1597,6 +1597,83 @@ def test_user_auth_updates_last_login(self):
...
+
+    def test_ip_from_radsetting_not_exist(self):
+        org2 = self._create_org(**{'name': 'test', 'slug': 'test'})
+        self._create_org_user(**{'organization': org2})
+        self.client.post(reverse('radius:user_auth_token', args=[org2.slug]))
+        with self.subTest('FREERADIUS_ALLOWED_HOSTS is 127.0.0.1'):
+            response = self.client.post(reverse('radius:authorize'), self.params)
+            self.assertEqual(response.data, {'control:Auth-Type': 'Accept'})
```
nemesisdesign, line 1671: please assert also the status code of the response

openwisp_radius/tests/test_api.py
```diff
@@ -1597,6 +1597,83 @@ def test_user_auth_updates_last_login(self):
...
+        with self.subTest('Empty Settings'), mock.patch(self.freeradius_hosts_path, []):
+            response = self.client.post(reverse('radius:authorize'), self.params)
+            self.assertEqual(response.data['detail'], self.fail_msg)
```
nemesisdesign, line 1674: please assert also the status code of the response

openwisp_radius/base/models.py
```diff
@@ -1034,7 +1035,14 @@ class AbstractOrganizationRadiusSettings(UUIDModel):
...
         help_text=_(
-            'Additional configuration for SMS backend in JSON format, if needed'
+            'Additional configuration for SMS backend in JSON format (optional)'
+        ),
+    )
+    freeradius_allowed_hosts = models.TextField(
+        null=True,
+        blank=True,
+        help_text=(
+            'Comma seperated list of IP addressses allowed to access freeradius API'
```
nemesisdesign, line 1045: typo: seperated > separated
typo: addressses > addresses

I would change it to:

`Comma separated list of IP addresses allowed to consume the freeradius API endpoints.`

for some reason this is not appearing in the admin, It's most probably happening because you're redefining the admin widget to indicate the number of rows and columns, but you're not passing the `help_text`, so it's getting lost.

openwisp_radius/base/models.py
```diff
@@ -1055,15 +1073,38 @@ def clean(self):
                     )
                 }
             )
+        # Clean `freeradius_allowed_hosts`
+        ip_list = self.freeradius_allowed_hosts_list
```
pandafy, line 1077: I will suggest you to rename `ip_list` to `allowed_hosts` for better readability. What do you think @atb00ker ?

openwisp_radius/base/models.py
```diff
@@ -1046,6 +1057,13 @@ class Meta:
     def __str__(self):
         return self.organization.name
 
+    @property
+    def freeradius_allowed_hosts_list(self):
+        addresses = []
+        if self.freeradius_allowed_hosts:
+            addresses = self.freeradius_allowed_hosts.split(',')
+        return addresses
+
```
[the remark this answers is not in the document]
pandafy: I have not tested this code, so please make necessary changes where needed. :sweat_smile:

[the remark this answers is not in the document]
atb00ker: `freeradius_allowed_hosts_list(self)` is actually only used to convert string to list. I am not sure how making changes in it can reflect on the `admin` change form.
Although, I think I'll read this response again in the morning before giving a proper response.

openwisp_radius/admin.py
```diff
@@ -493,15 +493,36 @@ def has_delete_permission(self, request, obj=None):
 UserAdminInline.inlines += [RadiusUserGroupInline, PhoneTokenInline]
 
 
+class AllowedHostsField(forms.CharField):
+    def prepare_value(self, value):
+        if not value:
+            value = ','.join(app_settings.FREERADIUS_ALLOWED_HOSTS)
+        return super().prepare_value(value)
+
```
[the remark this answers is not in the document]
pandafy: I was unaware too, TIL.  :smile:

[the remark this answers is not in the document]
pandafy: @atb00ker I will suggested move to forms related classes to a dedicated `forms.py`. :smile:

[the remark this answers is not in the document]
nemesisdesign: nice! :+1: :+1:

openwisp_radius/base/models.py
```diff
@@ -142,6 +143,9 @@
 STRATEGIES = (('prefix', _('Generate from prefix')), ('csv', _('Import from CSV')))
 
 _NOT_BLANK_MESSAGE = _('This field cannot be blank.')
+GET_IP_LIST_HELP_TEXT = _(
```
nemesisdesign, line 146: let's flag this as private, I would also do the same for `STRATEGIES` above:

`_GET_IP_LIST_HELP_TEXT` and `_STRATEGIES`

openwisp_radius/base/models.py
```diff
@@ -1055,15 +1073,42 @@ def clean(self):
                     )
                 }
             )
+        # Clean `freeradius_allowed_hosts`
+        allowed_hosts_set = set(self.freeradius_allowed_hosts_list)
```
nemesisdesign, line 1077: this block is quite big, what about moving this to a dedicated method?
```python
self._clean_freeradius_allowed_hosts()
```

openwisp_radius/migrations/0015_freeradius_allowed_hosts.py
```diff
@@ -0,0 +1,32 @@
...
+            field=models.TextField(
+                blank=True,
+                help_text='Comma separated list of IP addresses allowed to access freeradius API',
+                null=True,
+            ),
+        ),
+        migrations.AlterField(
+            model_name='organizationradiussettings',
+            name='sms_meta_data',
+            field=jsonfield.fields.JSONField(
```
nemesisdesign, line 26: here we changed only the help_text right? Can you edit the previous migration instead of adding this `AlterField`? These kinds of small improvements make the code more easy to maintain in the long run. (Imagine this kind of change being done in every migration, it would make the whole migrations harder to follow).

atb00ker:
Hi,

Please check the changes.

> maybe, to ensure users won't miss this, we should add a reference in Configure the REST module too

I didn't make this change yet, freeradius setup is done before the django setup, so I think people who don't know freeradius and django might be confused and try to set it up on freeradius itself, since they are not aware of the django setup is to be done separately on the server.

What do you think?

nemesisdesign:
Great work :+1: :clap:

PR #150, opened 2020-08-13
[bug] unexpected freeradius authorize failed

atb00ker:


openwisp_radius/api/views.py
```diff
@@ -191,7 +191,7 @@ def get_user(self, request):
         # ensure user is member of the authenticated org
         # or RadiusToken for the user exists.
         if (
-            RadiusToken.objects.filter(user=user, can_auth=True).exists()
+            RadiusToken.objects.filter(user=user).exists()
```
atb00ker, line 194: Why `can_auth` is removed here:

> Send auth with "username=admin&password=admin"
**Success**

> Send auth with "username=admin&password=radtoken"
**Success**

> Send auth with "username=admin&password=admin"
**Failed Unexpectedly**

Why? Because `can_auth=True` is false now!

nemesisdesign: can you please add a test for this bug? Should be quick.

PR #151, opened 2020-08-14
[api] Fixed busted DRF Freeradius API views 

atb00ker:


openwisp_radius/api/views.py
```diff
@@ -290,75 +278,50 @@ class AccountingView(TokenAuthorizationMixin, ListCreateAPIView):
...
-    def _get_status_type(self, request):
+    def _pre_validation(self, invalid_data):
         try:
-            return request.data['status_type']
+            status_type = invalid_data['status_type']
         except KeyError:
             raise RestValidationError({'status_type': [_('This field is required.')]})
+        for field in ['session_time', 'input_octets', 'output_octets']:
+            if status_type == 'Start' and invalid_data[field] == '':
+                invalid_data[field] = 0
```
nemesisdesign, line 317: I think we could do this in the serializer

openwisp_radius/api/serializers.py
```diff
@@ -98,9 +104,23 @@ def validate(self, data):
             self._disable_token_auth(data['username'])
         return data
 
+    def create(self, validated_data):
+        if app_settings.API_ACCOUNTING_AUTO_GROUP:
+            username = validated_data.get('username', '')
+            try:
+                user = User.objects.get(username=username)
+            except User.DoesNotExist:
+                logging.info(_(f'No corresponding user found for username: {username}'))
```
nemesisdesign, line 113: this shouldn't be really happening. I'm not sure why I chose to log it as info.
Can you please change it to `warning` and update the [related docs](https://openwisp-radius.readthedocs.io/en/latest/user/settings.html#openwisp-radius-api-accounting-auto-group)?

openwisp_radius/api/views.py
```diff
@@ -87,7 +87,7 @@ class FreeradiusApiAuthentication(BaseAuthentication):
     def _radius_token_authenticate(self, request):
         # cached_orgid exists only for users authenticated
         # successfully in past 24 hours
-        username = request.data.get('username')
+        username = request.data.get('username') or request.query_params.get('username')
```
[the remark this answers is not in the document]
nemesisdesign: If i'm not mistaken `query_params` is a DRF thing, it should be ok to use just that and not both, did you try that? If it works, we should do it to simplify.

openwisp_radius/api/views.py
nemesisdesign: I'm not sure about two things:

- why is the argument called `invalid_data`? Is not that the data is invalid, it's our API which is a bit different and needs to preprocess the data in order to [be more forgiving with what it receives](https://en.wikipedia.org/wiki/Robustness_principle)
- have you already tried to override the `validate` method of the serializer to perform these operations there?

Something like:

```python
def validate(self, ...)
    # do these operations before anything
    return super().validate(.........)
```

What is it that you tried exactly?

nemesisdesign: PS: alternatively, we may also use [field-level validation](https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation)

openwisp_radius/api/views.py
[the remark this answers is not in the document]
atb00ker: I am not sure what you mean here, 
Do you want me to keep only `request.data.get('username')` or `request.query_params.get('username')`?
All requests are using `request.data.get('username')`, and Authorize endpoint is expecting username and password in the POST data, that's why this is not removed. Similarly `PostAuth` expects the same.
However, Account `GET` request doesn't contain username in postdata, hence to authenticate it with `radius_token` the username is provided as `?username=admin`.
Are you suggesting always providing username in querystring itself for all the endpoints or have I misunderstood you?

openwisp_radius/api/serializers.py
```diff
@@ -78,6 +86,12 @@ def _disable_token_auth(self, user):
             radius_token.can_auth = False
             radius_token.save()
 
+    def run_validation(self, data):
+        for field in ['session_time', 'input_octets', 'output_octets']:
+            if data.get('status_type', None) == 'Start' and data[field] == '':
+                data[field] = 0
```
atb00ker, line 92: This will get the job done. `validate()` would have not worked because `run_validate` is done before `validate()`.

openwisp_radius/api/views.py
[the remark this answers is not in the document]
nemesisdesign: I meant that `request.query_params.get('username')` should be equivalent to `request.data.get('username')` so using both seems redundant, so I suggested to just use `request.query_params.get('username')`.
https://www.django-rest-framework.org/api-guide/requests/#query_params

openwisp_radius/api/serializers.py
nemesisdesign: Great!

openwisp_radius/api/views.py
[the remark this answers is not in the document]
atb00ker: @nemesisdesign 

Okay, just to reconfirm, the reason I've added both is:

`request.query_params` is a more correctly named synonym for request.GET.
While:
`request.data` returns the parsed content of the request body. This is similar to the standard request.POST

Hence, to remove `request.data.get()`

I'll have to change:

```
curl -X POST http://localhost:8000/api/v1/freeradius/authorize/ \
          -d "username=admin&password=admin"
```

to 

```
curl -X POST http://localhost:8000/api/v1/freeradius/authorize/?username=admin \
          -d "password=admin"
```

similarly, I'll have to make changes in `postauth` & `accounting` then when I receive the value in `POST` request of these endpoints, I'll have to modify their request body.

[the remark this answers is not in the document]
nemesisdesign: I was mistaken, for some reason I was reading `request.GET` instead of `request.data`! Sorry for that.

PR #170, opened 2020-09-28
[feature] Can restrict some mobile prefixes #160

### LINKED ISSUE #160: [feature] Allow restricting international mobile prefixes allowed to register when SMS verification is enabled
Allow restricting international mobile prefixes allowed to register when SMS verification is enabled.

This should be implemented like the freeradius allowed host features.

By default all prefixes are allowed.

NoumbissiValere:
Closes #160


requirements.txt
```diff
@@ -1,5 +1,6 @@
 django>=2.2,<3.2
 swapper>=1.1.0,<1.2.0
+djangorestframework<3.12,>=3.11
```
NoumbissiValere, line 3: Added this because build was failing as showed by the `travis buddy` above

NoumbissiValere:
@atb00ker please can you review this work in your free time? Thanks

nemesisdesign (changes requested):
Quoting from: https://github.com/openwisp/openwisp-radius/issues/160

> Therefore we need to allow organizations to specify which international prefixes they allow.
> By default, if this is unspecified, all prefixes should be allowed.

Our need now is to specify which countries are supported, but right now you implemented a way to specify which countries are not supported, this is problematic, because the immediate need is to allow only a handful of countries, so restricting the rest of the countries would be time consuming.

We should either reverse the implementation or add another option and field (eg: `OPENWISP_RADIUS_ALLOWED_MOBILE_PREFIXES` and `allowed_mobile_prefixes`) so we can do both.

I lean in favor of keeping it simple and just switching to `OPENWISP_RADIUS_ALLOWED_MOBILE_PREFIXES` and `allowed_mobile_prefixes`.

I still have not tested it nor reviewed the rest of the code in detail, but overall looks good.

NoumbissiValere:
Ok @nemesisdesign I will revere it to allowed. Thank you for reviewing

NoumbissiValere:
@nemesisdesign I have reversed the implementation and tested it

nemesisdesign (changes requested):
The admin field is missing:

Screenshot from 2020-10-06 22-45-58

Please add it and make sure it's listed after `sender`, it works as `freeradius_allowed_hosts` and try to replicate a test similar to `test_organization_radsettings_freeradius_allowed_hosts`.

See also my comments below.

openwisp_radius/base/models.py
```diff
@@ -146,6 +147,9 @@
 _GET_IP_LIST_HELP_TEXT = _(
     'Comma separated list of IP addresses allowed to access freeradius API'
 )
+_GET_MOBILE_PREFIX_HELP_TEXT = _(
+    'Comma separated list of international mobile prefixes to be restricted.'
```
nemesisdesign, line 151: Comma separated list of allowed international mobile prefixes.

nemesisdesign (approved):
Good job Noumbissi, thanks! :+1: :clap:

nemesisdesign (approved):

PR #172, opened 2020-10-07
[deps] Migrate dependencie rest-auth to dj-rest-auth #168

tullur:
Django-rest-auth is not actively maintained, switched to the maintained fork dj-rest-auth.
upd: `dj-rest-auth` replaced `rest_framework_jwt` with `django-rest-framework-simplejwt` (jazzband/dj-rest-auth#3) and now 'RegisterView' object has two attributes 'access_token' and 'refresh_token'

PR #173, opened 2020-10-08
migrate to dj_rest_auth

SoumyaRanjanNaik:
replaced django-rest-auth with dj-rest-auth.
This resolves issue #172


nemesisdesign:
#172 is already in an advanced state, I prefer to focus on that one. Thanks.

PR #176, opened 2020-10-14
Set the right User model in AUTH_USER_MODEL

filippolauria:
Also set SET_ID as stated in openwisp-users doc


nemesisdesign (approved):
thanks

PR #183, opened 2020-11-13
[feature] Keep track of previous phone number during phone number change #179

### LINKED ISSUE #179: [feature] Implement change phone number in a way that keeps track of previous phone number
### Inactive users

Inactive users can change the phone number time during the verification process because they may have inserted a wrong phone number.

### Active users

Active users may need to update their phone number if they change their SIM card, but in this case we should not update the mobile phone number field in the user model straightaway, because that would open potential malicious usage (user commits fraud using the network, then changes phone number, does not verify it and leaves the new unverified phone number in the system).

Here's how we can solve this issue:

- [ ] add a phone number field on phone token model which defaults to NULL
- [ ] create a data migration which fills the new field with the phone of the respective user
- [ ] create a new migration that makes the phone number field NOT NULL
- [ ] change the `PhoneToken` model logic so that the new phone number field  is used as recipient
- [ ] change the `ValidatePhoneTokenView` logic so that when the validation is successful, the user phone number is updated
- [ ] update other parts of the code which create the phone token objects to ensure the number is always present
- [ ] update `ValidateAuthTokenView` so that if the user is not active, the phone number returned in the response is the one associated to the last `PhoneToken` created
- [ ] update API docs

NoumbissiValere:
Closes #179


openwisp_radius/api/views.py
```diff
@@ -608,13 +608,18 @@ def post(self, request, *args, **kwargs):
                 radius_token = self.get_or_create_radius_token(
                     user, self.organization, renew=renew_required
                 )
+                p_token = PhoneToken.objects.filter(user=user).first()
```
nemesisdesign, line 611: you need to ensure it's ordered correctly, there's no mention whatsoever of that.
If there are multiple tokens in the system, we must be 100% sure the last one (creation date) is taken and the code must be explicit about this.

**Also, the query should not be made at all if the user is active.**

openwisp_radius/api/serializers.py
```diff
@@ -403,11 +403,17 @@ def validate_phone_number(self, phone_number):
         return phone_number
 
     def validate(self, data):
+        old_phone_number = self.user.phone_number
         self.user.phone_number = data['phone_number']
```
nemesisdesign, line 407: I think here we should simply remove this. The model validation should be changed so that if the user is not active having an empty phone number won't generate the validation error.

We should however raise a validation error in this serializer if the phone number is missing from the serializer data when sms verification is enabled.

nemesisdesign (changes requested):
Good progress @NoumbissiValere, see my comments below to improve it.
I haven't tested it yet.

openwisp_radius/api/views.py
NoumbissiValere: We are sure because it is ordered by [date created](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/base/models.py#L1156)

openwisp_radius/api/serializers.py
NoumbissiValere: I Think I should just improve this code and allow `PhoneToken` to handle the errors that belongs to him as it has already be done [here](https://github.com/openwisp/openwisp-radius/pull/183/commits/5e98ab84bc20b0dd0fd9996e4110eb0181723342#diff-4854b36face6ccda7d4d4cbe14d4eef0c25257dbf67e88f049173d80ba5954c6L809)

openwisp_radius/tests/test_api.py
```diff
@@ -2386,8 +2427,77 @@ def test_change_phone_number_restriction(self):
...
+        self.assertEqual(phone_token_qs.count(), 0)
+        user.refresh_from_db()
+        self.assertEqual(user.phone_number, new_phone_number)
+        self.assertEqual(user.is_active, is_active)
+
+    def test_active_user_change_number_sms_off(self):
+        self._test_change_phone_sms_off_helper(True)
+
+    def test_inactive_user_change_number_sms_off(self):
+        self._test_change_phone_sms_off_helper(False)
```
nemesisdesign, line 2503: please simplify the tests so they assume sms verification is always on.

When it's off the usage of the endpoint will be rejected by #189.

nemesisdesign (changes requested):
Good progress @NoumbissiValere, see my comments below.

openwisp_radius/api/serializers.py
```diff
@@ -403,13 +403,16 @@ def validate_phone_number(self, phone_number):
...
+            try:
+                self.user.full_clean()
+            except ValidationError as e:
+                raise serializers.ValidationError(self._get_error_dict(e))
+        else:
+            self.user.is_active = False
         return data
 
     def save(self):
-        self.user.is_active = False
```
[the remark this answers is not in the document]
NoumbissiValere: @nemesisdesign should we still flag the user as inactive even when the sms verification is not enabled?

nemesisdesign (changes requested):
Hey @NoumbissiValere, I also realized another potential issue.

Can you please add a test case which creates 2 users and then use the change phone number view to change the phone number of on of the user so it's the same as the phone number of the other user? We should make sure another user cannot take the phone number of another user, before we were using the uniqueness of the `phone_number` field so we didn't need a test for this, but now that we're changing this behavior we need an additional explicit test for it and we need to be sure it's handled well.

openwisp_radius/api/serializers.py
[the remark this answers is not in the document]
nemesisdesign: Please read again carefully #189.
`ChangePhoneNumberSerializer` will not be used when SMS verification is disabled.
The execution of the code will fail earlier.

openwisp_radius/base/models.py
```diff
@@ -1163,9 +1165,18 @@ def clean(self):
...
-            raise ValidationError(_('This user does not have a phone number.'))
+        another_user_has_phone_number = (
+            self._meta.model.objects.filter(phone_number=self.phone_number)
+            .exclude(user=self.user)
+            .exists()
+        )
+        if another_user_has_phone_number:
+            logger.warning(
+                _(f'phone_number: {self.phone_number} belongs to another user')
+            )
```
nemesisdesign, line 1176: I guess we can avoid logging this since it's a normal case of the application.

openwisp_radius/base/models.py
```diff
@@ -1163,9 +1165,18 @@ def clean(self):
...
+            raise ValidationError(
+                {'phone_number': _('Another user has this phone number')}
```
nemesisdesign, line 1178: let's change this to `This phone number is already associated to another user.`

nemesisdesign (changes requested):
Good progress Noumbissi, some minor changes requested below, then we should be ready to merge.

openwisp_radius/base/models.py
```diff
@@ -1163,9 +1165,19 @@ def clean(self):
...
             logger.warning(_(f'user {self.user} is already active'))
             raise ValidationError(_('This user is already active.'))
-        if not self.user.phone_number:
-            logger.warning(_(f'user {self.user} does not have a phone number'))
-            raise ValidationError(_('This user does not have a phone number.'))
+        another_user_has_phone_number = (
+            self._meta.model.objects.filter(phone_number=self.phone_number)
+            .exclude(user=self.user)
+            .exists()
+        )
```
nemesisdesign, line 1172: please do the following:

- create a test in which a user is created with a phone number, but no phone token is created for this user (the number may be entered manually by an administrator), then create another user which tries to change the phone number via the API to the same phone number of the first user, expect a validation error. The test should fail
- fix the test by using the user model for the lookup (load the model with django's `get_user_model()`)

nemesisdesign (changes requested):
My last round of testing brought up another issue, see below.

nemesisdesign (approved):
Thanks @NoumbissiValere :+1:

PR #185, opened 2020-11-16
[fix] Fixed password reset for inactive users #181

### LINKED ISSUE #181: [bug] Reset password for inactive user is bugged
Use case:

- user registers, mobile phone verification is enabled
- user does not complete mobile phone verification
- user comes back a few days later, tries to log in, does not remember the password
- user performs a password reset

Expected outcome:

- API response returns 200
- password reset email is sent

Actual outcome:

- API response returns 200
- But password reset email is **not** sent

## 1. step: reproduce

You may use openwisp-wifi-login-pages with mobile verification enabled to reproduce this.

## 2. design/propose a solution

Look for a solution.

## 3. write failing test case

Write failing test case in API tests.

## 4. write the fix, test and send patch

Write the fix. Test it via openwisp-wifi-login-pages.
Once it works, send the patch.

NoumbissiValere:
Fixes #181

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -26,6 +26,7 @@ from rest_framework.authtoken.serializers import (
 from rest_framework.exceptions import APIException
 
 from .. import settings as app_settings
+from ..base.forms import PasswordResetForm
 from ..utils import load_model
 from .utils import ErrorDictMixin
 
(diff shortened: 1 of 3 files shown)


nemesisdesign (commented):
First impression is good! :+1: Will test asap.

nemesisdesign (approved):
Thanks :+1:

PR #186, opened 2020-11-17
[fix] Fixed default password reset url and added its documentation #182

### LINKED ISSUE #182: [docs/fix] Fix and document OPENWISP_RADIUS_PASSWORD_RESET_URLS
Document ``OPENWISP_RADIUS_PASSWORD_RESET_URLS``, explain that it should point to the frontend application which users will use to manage their account and if they don't have one they can use https://github.com/openwisp/openwisp-wifi-login-pages/.

Make sure the default value works: get the domain from the site name (`Site` model), change the default to have `{site.domain}` instead of `example.com` and make sure the substitution works.

NoumbissiValere:
Fixes #182

### DIFF
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -11,6 +11,7 @@ from dj_rest_auth.views import PasswordResetView as BasePasswordResetView
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.tokens import default_token_generator
+from django.contrib.sites.shortcuts import get_current_site
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.db.utils import IntegrityError
@@ -710,6 +711,9 @@ class PasswordResetView(DispatchOrgMixin, BasePasswordResetView):
         token = default_token_generator.make_token(user)
         password_reset_urls = app_settings.PASSWORD_RESET_URLS
         default_url = password_reset_urls.get('default')
+        domain = get_current_site(self.request).domain
+        if 'example.com' not in domain:
+            default_url = default_url.replace('http://localhost:8080', domain)
         if getattr(self, 'swagger_fake_view', False):
             organization_pk, organization_slug = None, None  # pragma: no cover
         else:
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -712,8 +712,6 @@ class PasswordResetView(DispatchOrgMixin, BasePasswordResetView):
         password_reset_urls = app_settings.PASSWORD_RESET_URLS
         default_url = password_reset_urls.get('default')
         domain = get_current_site(self.request).domain
-        if 'example.com' not in domain:
-            default_url = default_url.replace('http://localhost:8080', domain)
         if getattr(self, 'swagger_fake_view', False):
             organization_pk, organization_slug = None, None  # pragma: no cover
         else:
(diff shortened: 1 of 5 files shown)


openwisp_radius/settings.py
```diff
@@ -68,7 +68,8 @@ def get_settings_value(option, default):
 PASSWORD_RESET_URLS = {
     # fallback in case the specific org page is not defined
     'default': (
-        'https://example.com/{organization}/password/confirm/reset/{uid}/{token}'
+        'http://{domain}/api/v1/{organization}/account/password/reset/confirm/?'
+        'uid={uid}&token={token}'
```
nemesisdesign, line 72: this should be `https://example.com/{organization}/password/reset/confirm/{uid}/{token}`

docs/source/developer/setup.rst
```diff
@@ -301,6 +301,9 @@ Launch development server:
 
 You can access the admin interface at http://127.0.0.1:8000/admin/.
 
+Login to the admin interface and update the domain name of your `site` to the domain
+on which this application is running. From above, it should `127.0.0.1:8000`.
```
nemesisdesign, line 305: it should be. The slant quotes should be double, eg:

```
``127.0.0.1:8000``
```

nemesisdesign (changes requested):

openwisp_radius/settings.py
NoumbissiValere: @nemesisdesign  I think this should be `https://example.com/{organization}/password/reset/confirm/?uid={uid}&token={token}`

This is so that the frontend can extra the `uid` and `token` from the link and send it with the new password to the backend. if we don't do that, how will the frontend know the `uid` and `token` to send to the user?

nemesisdesign: @NoumbissiValere it should default to the URL used by [openwisp-wifi-login-pages](https://github.com/openwisp/openwisp-wifi-login-pages/) to perform the last step of this process, which allows the user to change their password (which is the URL I suggested, but please double check).
If a user uses a different system they'll be smart enough to know they have to change this setting.

nemesisdesign: Also, please make it default to HTTPS, but set this setting in the development environment (`test/settings.py`) to use `http://localhost:8080`.

NoumbissiValere: @nemesisdesign please if we are doing this, then what will be the use of `domain` from site

nemesisdesign: to use as prod default

docs/source/developer/setup.rst
NoumbissiValere: with the new implementation, This doc is no longer necessary so I will remove it

openwisp_radius/api/views.py
```diff
@@ -710,6 +711,9 @@ def get_serializer_context(self):
         token = default_token_generator.make_token(user)
         password_reset_urls = app_settings.PASSWORD_RESET_URLS
         default_url = password_reset_urls.get('default')
+        domain = get_current_site(self.request).domain
+        if 'example.com' not in domain:
+            default_url = default_url.replace('http://localhost:8080', domain)
```
nemesisdesign, line 716: don't do this, is really ugly and confusing. Pass `site=site.domain` to the `format` call below.

nemesisdesign (changes requested):
I improved the docs, please read it carefully, see the default value indicated and implement it as explained there.

To recap:

- the default value must make use of the `site` domain as in the django site framework
- the value `http://localhost:8080/{organization}/password/reset/confirm/{uid}/{token}` is only used in the dev env, there's should be no reference of localhost from the code of the application, see m code below

openwisp_radius/settings.py
```diff
@@ -66,9 +66,7 @@ def get_settings_value(option, default):
 
 PASSWORD_RESET_URLS = {
     # fallback in case the specific org page is not defined
-    'default': (
-        'https://example.com/{organization}/password/reset/confirm/{uid}/{token}'
-    ),
+    'default': ('https://{site}/{organization}/password/reset/confirm/{uid}/{token}'),
```
nemesisdesign, line 69: I guess we can remove the parenthesis from the string

nemesisdesign (changes requested):

nemesisdesign (approved):
:+1:

PR #190, opened 2020-11-21
[api] Disable endpoints for organizations without SMS verification #189

### LINKED ISSUE #189: [api] Make sure phone token API endpoints cannot be used if sms verification not enabled
If an organization does not enable SMS verification, the API endpoints should be turned off:

- an sms backend may not be configured
- or if configured, it may not have credit
- or if it has credit, it may not be intended to be consumed by openwisp-radius because the feature is disabled

For these reasons, we should add some kind of check, [probably a custom DRF permission class](https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions), which can be reused for every API view that has to do with the phone number.

This will also allow us to avoid repeating the same kind of check many different times in different places and in different ways.

Every view involving the phone number inherits this mixin which adds an `organization` attribute to the view: https://github.com/openwisp/openwisp-radius/blob/0819e35bef389b56b12fbba0ab85d9a679020106/openwisp_radius/api/views.py#L418-L430

That attribute can then be accessed as follows: `self.organization.radius_settings.sms_verification`.
If accessing this attribute fails (eg: the RadiusOrganizationSettings object is missing) or any other reason, we should return a 500 internal server error in the API response and log the problem using `logger.error()` (we need a test for this case).

In case the `sms_verification` is disabled, the custom permission class should log a warning (with `logger.warning`), eg:

```python
logger.warning(f'View {view.__class__.__name__} is being accessed for organization {self.organization.name} but SMS verification is disabled for this organization.')
```

We need a test for this case as well, the `warning` method can be mocked so we can assert it has been called.

Purhan:
Closes #189

### DIFF
--- a/openwisp_radius/api/permissions.py
+++ b/openwisp_radius/api/permissions.py
@@ -1,0 +1,27 @@
+import logging
+
+from django.utils.translation import gettext_lazy as _
+from ipware import get_client_ip
+from rest_framework.exceptions import PermissionDenied
+from rest_framework.permissions import BasePermission
+
+from .utils import is_sms_verification_enabled
+
+logger = logging.getLogger(__name__)
+
+
+class IsSmsVerificationEnabled(BasePermission):
+    def has_permission(self, request, view):
+        organization = getattr(view, 'organization')
+        client_ip = get_client_ip(request)[0]
+        verification = is_sms_verification_enabled(organization)
+        if not verification:
+            logger.warning(
+                f'View {view.__class__.__name__} is being accessed for organization '
+                f'{organization.name} but SMS verification is disabled for '
+                f'this organization. Client IP address: {client_ip}'
+            )
+            raise PermissionDenied(
+                _('SMS verification is not enabled for this organization')
+            )
+        return verification
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -2576,3 +2576,63 @@ class TestApiPhoneToken(ApiTokenMixin, BaseTestCase):
         # is_active state of user should not change because an error
         # occurred during phone token creation.
         self.assertTrue(user.is_active)
+
+
+class TestIsSmsVerificationEnabled(ApiTokenMixin, BaseTestCase):
+
+    _test_email = 'test@openwisp.org'
+
+    def setUp(self):
+        super().setUp()
+        radius_settings = self.default_org.radius_settings
+        radius_settings.sms_verification = False
+        radius_settings.full_clean()
+        radius_settings.save()
+
+    def test_register_201_phone_number_empty(self):
+        self._superuser_login()
+        self.assertEqual(User.objects.count(), 1)
+        url = reverse('radius:rest_register', args=[self.default_org.slug])
+        phone_number = '+393664255801'
+        r = self.client.post(
+            url,
+            {
+                'username': self._test_email,
+                'email': self._test_email,
+                'password1': 'password',
+                'password2': 'password',
+                'phone_number': phone_number,
+            },
+        )
+        self.assertEqual(r.status_code, 201)
+        self.assertIn('key', r.data)
+        self.assertEqual(User.objects.count(), 2)
+        user = User.objects.get(email=self._test_email)
+        self.assertTrue(user.is_member(self.default_org))
+        self.assertEqual(user.phone_number, None)
+        self.assertTrue(user.is_active)
+
+    def test_create_phone_token_403(self):
+        url = reverse('radius:phone_token_create', args=[self.default_org.slug])
+        r = self.client.post(url)
+        self.assertEqual(r.status_code, 403)
+        self.assertIn('SMS verification is not enabled', str(r.data))
+
+    def test_validate_phone_token_403(self):
+        url = reverse('radius:phone_token_validate', args=[self.default_org.slug])
+        r = self.client.post(url)
+        self.assertEqual(r.status_code, 403)
+        self.assertIn('SMS verification is not enabled', str(r.data))
+
+    def test_change_phone_number_403(self):
+        url = reverse('radius:phone_number_change', args=[self.default_org.slug])
+        r = self.client.post(url)
+        self.assertEqual(r.status_code, 403)
+        self.assertIn('SMS verification is not enabled', str(r.data))
+
+    def test_missing_radius_settings(self):
+        self.default_org.radius_settings.delete()
+        url = reverse('radius:phone_token_create', args=[self.default_org.slug])
+        r = self.client.post(url)
+        self.assertEqual(r.status_code, 500)
+        self.assertIn('Could not complete operation', str(r.data))
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -28,7 +28,7 @@ from rest_framework.exceptions import APIException
 from .. import settings as app_settings
 from ..base.forms import PasswordResetForm
 from ..utils import load_model
-from .utils import ErrorDictMixin
+from .utils import ErrorDictMixin, is_sms_verification_enabled
 
 logger = logging.getLogger(__name__)
 
@@ -326,22 +326,9 @@ class RegisterSerializer(
         default='',
     )
 
-    @property
-    def is_sms_verification_enabled(self):
-        org = self.context['view'].organization
-        try:
-            return org.radius_settings.sms_verification
-        except ObjectDoesNotExist:
-            raise APIException(
-                _(
-                    'Could not complete operation '
-                    'because of an internal misconfiguration'
-                )
-            )
-
     def validate_phone_number(self, phone_number):
         org = self.context['view'].organization
-        if self.is_sms_verification_enabled:
+        if is_sms_verification_enabled(org):
             if not phone_number:
                 raise serializers.ValidationError(_('This field is required'))
             mobile_prefixes = org.radius_settings.allowed_mobile_prefixes_list
@@ -349,6 +336,9 @@ class RegisterSerializer(
                 raise serializers.ValidationError(
                     _('This international mobile prefix is not allowed.')
                 )
+        else:
+            # Phone number should not be stored if sms verification is disabled
+            phone_number = None
         return phone_number
 
     def save(self, request):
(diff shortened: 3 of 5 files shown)


openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,36 @@
...
+        verification = self.has_sms_verification(view, organization)
+        if not verification:
+            message = _(
+                f'View {view.__class__.__name__} is being accessed for organization '
+                f'{organization.name} but SMS verification is disabled for '
+                'this organization.'
+            )
+            logger.warning(message)
+            raise PermissionDenied(message)
+        else:
```
nemesisdesign, line 25: let's avoid the else here since it's not necessary

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,36 @@
...
+        raises PermissionDenied if organization has disabled sms_verification
+        """
+        organization = getattr(view, 'organization')
+        verification = self.has_sms_verification(view, organization)
+        if not verification:
+            message = _(
+                f'View {view.__class__.__name__} is being accessed for organization '
+                f'{organization.name} but SMS verification is disabled for '
+                'this organization.'
+            )
```
nemesisdesign, line 22: let's keep this message only for internal logging, do not flag it as translatable.
Can you also log the request IP please?

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,36 @@
...
+            logger.warning(message)
+            raise PermissionDenied(message)
```
nemesisdesign, line 24: here we can use a translatable string, like `SMS verification is not enabled for this organization`.

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,36 @@
+import logging
+
+from django.utils.translation import gettext_lazy as _
+from rest_framework.exceptions import APIException, PermissionDenied
+from rest_framework.permissions import BasePermission
+
+logger = logging.getLogger(__name__)
+
+
+class HasSmsVerification(BasePermission):
```
nemesisdesign, line 10: I would rename this to `IsSmsVerificationEnabled`

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,36 @@
...
+            raise PermissionDenied(message)
+        else:
+            return True
+
+    def has_sms_verification(self, view, org):
+        try:
+            verification = org.radius_settings.sms_verification
+        except Exception as e:
+            # Raise exception in case an attribute is missing from org
+            logger.error(e)
```
nemesisdesign, line 33: I would use `logger.exception(f'Got exception while accessing radius_settings for {org.name}')`

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,36 @@
...
+            raise APIException(e)
```
nemesisdesign, line 34: Never leak internal system information to users.

Use a generic error message instead.

In this case, there's already another part of the system which does a similar check, we could reuse the same string: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/serializers.py#L335-L340
You could make the string a variable and use it in both places, so we avoid repetition.

openwisp_radius/tests/test_api.py
```diff
@@ -2414,3 +2414,64 @@ def test_change_phone_number_restriction(self):
...
+class TestApiRejectPhoneToken(ApiTokenMixin, BaseTestCase):
+
+    _test_email = 'test@openwisp.org'
+
+    def setUp(self):
+        super().setUp()
+        radius_settings = self.default_org.radius_settings
+        radius_settings.sms_verification = False
+        radius_settings.sms_sender = '+595972157632'
+        radius_settings.sms_meta_data = {
```
nemesisdesign, line 2428: avoid setting the sms related fields, this will give us a more realistic test env

openwisp_radius/tests/test_api.py
```diff
@@ -2414,3 +2414,64 @@ def test_change_phone_number_restriction(self):
...
+                'password2': 'password',
+                'phone_number': phone_number,
+            },
+        )
+        self.assertEqual(r.status_code, 201)
+        self.assertIn('key', r.data)
+        self.assertEqual(User.objects.count(), 2)
+        user = User.objects.get(email=self._test_email)
+        self.assertTrue(user.is_member(self.default_org))
+        self.assertEqual(user.phone_number, phone_number)
```
nemesisdesign, line 2455: if SMS verification is disabled, I thought the phone number field should not be saved, but it looks like it does.
This is an issue, because mobile phone verification may be enabled later on by an org and they'll have active users with the phone number already filled, and may think it's been verified but it hasn't.

Could you help me to fix this in the API? It could be handled here:
https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/serializers.py#L342-L352
If sms verification is not enabled, we could return an empty string, which should force the phone number to be empty.
The expected value in this case is an empty string.
If you can do this change, please do it in a separate commit (eg: `[change] Force User.phone_number to be empty if SMS verification disabled`).

openwisp_radius/tests/test_api.py
```diff
@@ -2414,3 +2414,64 @@ def test_change_phone_number_restriction(self):
...
+        self.assertTrue(user.is_active)
+
+    def test_create_phone_token_403(self):
+        url = reverse('radius:phone_token_create', args=[self.default_org.slug])
+        r = self.client.post(url)
+        self.assertEqual(r.status_code, 403)
```
nemesisdesign, line 2461: would be good to check also the error message to be 100% sure, eg: `self.assertIn('SMS verification', str(r.data))`, applies for the other tests as well.

nemesisdesign (changes requested):
Great progress @Purhan! :+1: 
See my comments below to improve it further.

openwisp_radius/api/permissions.py
Purhan: I just reused the same string. Is it verbose this way, should I use it as a variable instead? If yes, which file would be suitable for storing it in that case?

Purhan:
@nemesisdesign I have implemented all the recommended changes, please take another look :smile:.

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+            )
+            raise PermissionDenied(
+                _('SMS verification is not enabled for this organization')
+            )
+        return verification
+
+    def has_sms_verification(self, view, org):
+        try:
+            verification = org.radius_settings.sms_verification
+        except Exception:
```
nemesisdesign, line 30: please use `ObjectDoesNotExist`, we don't expect any other exception, if there will be any other exception it's better to let it fail and then fix it while also adding another test case for it.

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+                f'this organization. Client IP address: {client_ip}'
+            )
+            raise PermissionDenied(
+                _('SMS verification is not enabled for this organization')
+            )
+        return verification
+
+    def has_sms_verification(self, view, org):
+        try:
+            verification = org.radius_settings.sms_verification
```
nemesisdesign, line 29: here you can just return directly and remove the last return statement

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+                f'View {view.__class__.__name__} is being accessed for organization '
+                f'{organization.name} but SMS verification is disabled for '
+                f'this organization. Client IP address: {client_ip}'
+            )
+            raise PermissionDenied(
+                _('SMS verification is not enabled for this organization')
+            )
+        return verification
+
+    def has_sms_verification(self, view, org):
```
nemesisdesign, line 27: rename it to `is_sms_verification_enabled`, remove the `view` argument if not used

openwisp_radius/api/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+            logger.exception(
+                f'Got exception while accessing radius_settings for {org.name}'
+            )
+            raise APIException(
+                _(
+                    'Could not complete operation '
+                    'because of an internal misconfiguration'
+                )
+            )
+        return verification
```
nemesisdesign, line 40: please make this a reusable function in `openwisp_radius.api.utils` so we can eliminate the repetition between this code and `RegisterSerializer.is_sms_verification_enabled`.

eg:

```python
def is_sms_verification_enabled(organization):
    try:
        return organization.radius_settings.sms_verification
    except ObjectDoesNotExist:
        logger.exception(
            'Got exception while accessing '
            f'radius_settings for {organization.name}'
        )
        raise APIException(
            _(
                'Could not complete operation '
                'because of an internal misconfiguration'
            )
        )
```

openwisp_radius/tests/test_api.py
```diff
@@ -2414,3 +2414,63 @@ def test_change_phone_number_restriction(self):
             self.assertEqual(user.phone_number, phone_number)
 
         app_settings.ALLOWED_MOBILE_PREFIXES = []
+
+
+class TestApiRejectPhoneToken(ApiTokenMixin, BaseTestCase):
```
nemesisdesign, line 2419: I think this should be called `TestIsSmsVerificationEnabled`

openwisp_radius/tests/test_api.py
```diff
@@ -2414,3 +2414,63 @@ def test_change_phone_number_restriction(self):
...
+    _test_email = 'test@openwisp.org'
+
+    def setUp(self):
+        super().setUp()
+        radius_settings = self.default_org.radius_settings
+        radius_settings.sms_verification = False
+        radius_settings.full_clean()
+        radius_settings.save()
+
+    def test_register_201(self):
```
nemesisdesign, line 2430: Please change the name of the test or it will be confusing.

A good name in this case would be `test_register_201_phone_number_empty`, which indicates the API request will be successful but we expect the phone number to be empty

nemesisdesign (changes requested):
Great progress @Purhan! :+1: 

I left some comments to improve it further, after the requested changes are made it should be ready to merge.

Purhan:
@nemesisdesign Thanks, I have implemented all the suggested changes.

nemesisdesign (changes requested):
@Purhan can you rebase on master and resolve conflicts please?

Purhan:
> @Purhan can you rebase on master and resolve conflicts please?

@nemesisdesign Done.

nemesisdesign (approved):
Great work @Purhan :+1:

PR #193, opened 2020-12-01
[fix] Handled phone token validation error #191

### LINKED ISSUE #191: [api] Handle change phone number validation error without marking the user as inactive
Follow up to #183. 

When changing the phone number, if the creation of the token fails for some reason, the user will be flagged as inactive even though their number has not been changed.

We can solve this in different ways:

- catch the validation error and set is_active back to True, not really elegant
- allow the creation of phone token objects also for active users, flag the user as inactive only after the token has been created successfully - sounds cleaner

@NoumbissiValere what do you think?

NoumbissiValere:
Fixed #191

### DIFF
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -2051,16 +2051,14 @@ class TestApiPhoneToken(ApiTokenMixin, BaseTestCase):
         self.assertIn('phone_number', r.data)
         self.assertIn('This field cannot be null.', str(r.data['phone_number']))
 
-    def test_create_phone_token_400_user_already_active(self):
+    def test_create_phone_token_201_user_already_active(self):
         self.test_register_201()
         token = Token.objects.last()
         token.user.is_active = True
         token.user.save()
         url = reverse('radius:phone_token_create', args=[self.default_org.slug])
         r = self.client.post(url, HTTP_AUTHORIZATION=f'Bearer {token.key}')
-        self.assertEqual(r.status_code, 400)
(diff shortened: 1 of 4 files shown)


openwisp_radius/tests/test_api.py
```diff
@@ -2570,3 +2568,25 @@ def test_user_phone_number_unique(self):
...
+                'password1': 'password',
+                'password2': 'password',
+                'phone_number': '+237674479231',
+            }
+        )
+        user = User.objects.get(email='user2@gmail.com')
+        user.is_active = True
+        user.save()
+        self._test_phone_number_unique_helper('+23767779235')
+        self.assertTrue(user.is_active)
```
nemesisdesign, line 2592: I am not sure if this test covers a case that it was not already tested before, are you sure it's needed?
We already got to almost 2600 lines of code in this test file, we'll have to split it, but for the moment we can start double checking if new tests are really needed.

NoumbissiValere: Yes @nemesisdesign . The test is to ensure that when there is a problem during phone number change, the state of the user is not changed. There's no test  for this reason why I added it

nemesisdesign: Nor the test name nor the test code reflect that. Please make it more explicit and easy to read. Where is the error being triggered?

nemesisdesign (changes requested):

nemesisdesign (approved):
Thanks!

PR #198, opened 2020-12-03
[api] Reduce queries for radius_settings #195

### LINKED ISSUE #195: [api] Use JOIN when fetching organization in DispatchOrgMixin
Once https://github.com/openwisp/openwisp-radius/pull/190 is merged, we'll be accessing the organization radius settings object very often, which will generate at least one extra query each time.

Here's what we should do avoid that:

- [ ] add [assertNumQueries](https://docs.djangoproject.com/en/3.1/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) in each API call executed in tests that were added in https://github.com/openwisp/openwisp-radius/pull/190
- [ ] change DispatchOrgMixin https://github.com/openwisp/openwisp-radius/blob/5acaad70db27165dd8c51f1ad65a28988fcfcdea/openwisp_radius/api/views.py#L419-L422
we need to forgo using the [get_obejct_or_404](https://docs.djangoproject.com/en/3.1/topics/http/shortcuts/#get-object-or-404) shortcut (resorting to the extended form mentioned in the docs instead), and we'll have to execute the query using `Organization.objects.select_related('radius_settings').get(slug=kwargs['slug'])`
- [ ] run the tests again, we expect the `assertNumQueries` we added in the first step to fail, now we should have at least one less executed query for each test

Purhan:
Closes #195

### DIFF
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -424,7 +424,7 @@ class DispatchOrgMixin(object):
                 'radius_settings'
             ).get(slug=kwargs['slug'])
         except Organization.DoesNotExist:
-            raise Http404("No Organization matches the given query.")
+            raise Http404('No Organization matches the given query.')
         return super().dispatch(*args, **kwargs)
 
     def validate_membership(self, user):
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -419,7 +419,12 @@ batch = BatchView.as_view()
 
 class DispatchOrgMixin(object):
     def dispatch(self, *args, **kwargs):
-        self.organization = get_object_or_404(Organization, slug=kwargs['slug'])
+        try:
+            self.organization = Organization.objects.select_related(
+                'radius_settings'
+            ).get(slug=kwargs['slug'])
+        except Organization.DoesNotExist:
+            raise Http404("No Organization matches the given query.")
         return super().dispatch(*args, **kwargs)
 
     def validate_membership(self, user):
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -2599,8 +2599,8 @@ class TestIsSmsVerificationEnabled(ApiTokenMixin, BaseTestCase):
         url = reverse('radius:rest_register', args=[self.default_org.slug])
         phone_number = '+393664255801'
         numQueries = 50
-        # TODO: Remove this when support for django 2 is dropped
-        if django.get_version().startswith('2'):
+        # TODO: Remove this when support for django < 3.0 is dropped
+        if django.VERSION < (3, 0):
             numQueries = 53
         with self.assertNumQueries(numQueries):
             r = self.client.post(
--- a/openwisp_radius/tests/test_api.py
+++ b/openwisp_radius/tests/test_api.py
@@ -2597,16 +2597,17 @@ class TestIsSmsVerificationEnabled(ApiTokenMixin, BaseTestCase):
         self.assertEqual(User.objects.count(), 1)
         url = reverse('radius:rest_register', args=[self.default_org.slug])
         phone_number = '+393664255801'
-        r = self.client.post(
-            url,
-            {
-                'username': self._test_email,
-                'email': self._test_email,
-                'password1': 'password',
-                'password2': 'password',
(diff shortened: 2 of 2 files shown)


openwisp_radius/tests/test_api.py
```diff
@@ -2594,16 +2594,17 @@ def test_register_201_phone_number_empty(self):
...
-            url,
-            {
-                'username': self._test_email,
-                'email': self._test_email,
-                'password1': 'password',
-                'password2': 'password',
-                'phone_number': phone_number,
-            },
-        )
+        with self.assertNumQueries(50):
```
nemesisdesign, line 2597: :scream:

openwisp_radius/api/views.py
```diff
@@ -419,7 +419,12 @@ def post(self, request, *args, **kwargs):
 
 class DispatchOrgMixin(object):
     def dispatch(self, *args, **kwargs):
-        self.organization = get_object_or_404(Organization, slug=kwargs['slug'])
+        try:
+            self.organization = Organization.objects.select_related(
+                'radius_settings'
+            ).get(slug=kwargs['slug'])
+        except Organization.DoesNotExist:
```
nemesisdesign, line 426: If we use the [DRF NotFound exception](https://www.django-rest-framework.org/api-guide/exceptions/#notfound), the API response should be more consistent with the rest of the responses from the API. Can you try it and let me know?

nemesisdesign (commented):
Great @Purhan, looks good! See my comment below.

PS: please also rebase on the current master.

openwisp_radius/tests/test_api.py
Purhan: :laughing: I was also freaked out when I saw this. Is there something we can do to reduce this?

openwisp_radius/api/views.py
Purhan: For some reason, this is not being handled by the DRF exception handler. Look at the screenshot attached below, this is what I get.

image

openwisp_radius/tests/test_api.py
Purhan: Also looks like Travis checks failed for this on some versions of django. This needs a different approach altogether...

nemesisdesign (commented):
The failure on django 2.2 may be handled with an if which checks if the django version is less than 3 and in that case it would use 53 instead of 50 (adding also a TODO comment hinting to remove when django 2.2 support is removed).

openwisp_radius/tests/test_api.py
```diff
@@ -2594,16 +2595,21 @@ def test_register_201_phone_number_empty(self):
...
-                'username': self._test_email,
-                'email': self._test_email,
-                'password1': 'password',
-                'password2': 'password',
-                'phone_number': phone_number,
-            },
-        )
+        numQueries = 50
+        # TODO: Remove this when support for django 2 is dropped
+        if django.get_version().startswith('2'):
```
nemesisdesign, line 2600: this is an unusual way to do the comparison, I suggest to keep consistency with the way it's usually done, eg:
https://github.com/openwisp/openwisp-users/blob/fabef8d63912c0c492248079e82fb3092f9b8ab8/openwisp_users/base/models.py#L166

There are many ways of doing it, but that's the one of the simplest and the one we've been using in several ow modules.

nemesisdesign (changes requested):
Thanks @Purhan, only one minor issue below.

openwisp_radius/tests/test_api.py
Purhan: Done :+1:

openwisp_radius/api/views.py
```diff
@@ -419,7 +419,12 @@ def post(self, request, *args, **kwargs):
...
+            raise Http404("No Organization matches the given query.")
```
nemesisdesign, line 427: spotted one minor issue, please use single quotes for consistency

nemesisdesign (changes requested):
Tested and it's good, there's one minor issue below I want you to fix. I could fix it myself but if you fix it we'll avoid it next time.

openwisp_radius/api/views.py
Purhan: Done

nemesisdesign (approved):

PR #200, opened 2020-12-04
[qa] Move CI build to GitHub Actions

pandafy:
Removed Travis

Fixes #199

PR #204, opened 2020-12-10
[qa] Improve 404 broken link check

nemesisdesign:
The runsphinx-build created the build files in ./docs/build instead of
./docs/build/html like `make html` does.

This inconsistency caused that pylinkvalidator didn't spot internal
links returning 404.

Note: the crawler we use does not check external resources by default.
I gave it a run to check external resources and found 1 broken link to github, but I think we cannot do this in the build because the process is too slow.


nemesisdesign:
Damn. The build is once again not starting for who knows what reason. I will stop wasting time debugging Travis. I'm tired.
I'll just merge this. I hope #200 get merged soon.

PR #207, opened 2020-12-21
[feature] Added authorization by email and phone number #205

### LINKED ISSUE #205: [feature] Allow logging in with phone number, email or username
Depends on https://github.com/openwisp/openwisp-users/issues/206, once this is implemented we need to change the `authorize` API endpoint to allow logging in with these additional fields.

Important points:

- [ ] the string supplied for username, email or phone number must be not empty.
- [ ] we could use an OR query, in order of precedence: get user with this email or this phone number or this username
- [ ] it's possible the query will return multiple results, the precedence of the previous point should be respected: write a test in which user A has username: user@test.com and user B has email user@test.com, test the authentication backend and ensure user B gets authenticated correctly
- [ ] write a similar test like the one before but with a phone number

NoumbissiValere:
Closes #205

### DIFF
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -212,7 +212,7 @@ class AuthorizeView(GenericAPIView):
         serializer.is_valid(raise_exception=True)
         username = serializer.validated_data.get('username')
         password = serializer.validated_data.get('password')
-        user = self.get_user(request, username, password)
+        user = self.get_user(request, username)
         if user and self.authenticate_user(request, user, password):
             return Response(self.accept_attributes, status=self.accept_status)
         if app_settings.API_AUTHORIZE_REJECT:
@@ -220,14 +220,14 @@ class AuthorizeView(GenericAPIView):
         else:
             return Response(None, status=200)
 
-    def get_user(self, request, username, password):
+    def get_user(self, request, username):
         """
         return active user or ``None``
         """
         try:
-            user = auth_backend.authenticate(
-                request, username, password
-            ) or User.objects.get(username=username, is_active=True)
+            user = auth_backend.get_user(username) or User.objects.get(
+                username=username, is_active=True
+            )
         except User.DoesNotExist:
             return None
         # ensure user is member of the authenticated org
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -54,6 +54,7 @@ from rest_framework.throttling import BaseThrottle  # get_ident method
 
 from openwisp_users.api.authentication import BearerAuthentication
 from openwisp_users.api.permissions import IsOrganizationManager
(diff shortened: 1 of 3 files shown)


NoumbissiValere:
This PR needs [openwisp-users#206](https://github.com/openwisp/openwisp-users/pull/208) for testing

openwisp_radius/api/views.py
```diff
@@ -210,20 +212,22 @@ def post(self, request, *args, **kwargs):
...
-    def get_user(self, request, username):
+    def get_user(self, request, username, password):
         """
         return active user or ``None``
         """
         try:
-            user = User.objects.get(username=username, is_active=True)
+            user = auth_backend.authenticate(
+                request, username, password
+            ) or User.objects.get(username=username, is_active=True)
```
nemesisdesign, line 230: You can't run an authentication in this method, which is supposed to simply fetch the user.
It's very wrong for multiple reasons.

It breaks the radius user token authentication mechanism, because it checks the password, but end users authenticate using one time tokens (see `RadiusToken`).

The check password or check radius token operation is performed by this method:

https://github.com/openwisp/openwisp-radius/blob/5752ba2a419126722d637bc42edf4aa4a2296992/openwisp_radius/api/views.py#L244-L253

If you want to reuse the logic to fetch the user from `UserAuthenticationBackend`, which sounds good because it allows us to avoid the repetition, please separate the logic that fetches the user in the authentication backend, create a `get_user` method (which doens't check the password) so it can be reused here.

openwisp_radius/api/views.py
```diff
@@ -210,20 +212,22 @@ def post(self, request, *args, **kwargs):
...
+        user = self.get_user(request, username, password)
         if user and self.authenticate_user(request, user, password):
             return Response(self.accept_attributes, status=self.accept_status)
         if app_settings.API_AUTHORIZE_REJECT:
             return Response(self.reject_attributes, status=self.reject_status)
         else:
             return Response(None, status=200)
 
-    def get_user(self, request, username):
+    def get_user(self, request, username, password):
```
nemesisdesign, line 223: this is wrong!

nemesisdesign (changes requested):

openwisp_radius/api/views.py
```diff
@@ -223,7 +225,9 @@ def get_user(self, request, username):
         return active user or ``None``
         """
         try:
-            user = User.objects.get(username=username, is_active=True)
+            user = auth_backend.get_user(username) or User.objects.get(
+                username=username, is_active=True
+            )
```
nemesisdesign, line 230: I don't understand, why did you add the or?

I'm confused.

nemesisdesign (approved):

nemesisdesign (approved):

Issue #83, opened 2020-04-11
[radius-tests] Avoid error logging to print noisy output during tests


nemesisdesign:
Depends on https://github.com/openwisp/openwisp-utils/issues/98.

Use the decorators mentioned in the issue linked above to silence the output of noisy tests.

atb00ker:
Description: 

Easy issue for new contributors:

1. Setup the repository
2. Run tests (notice it outputs a lot of things)
3. Checkout the following functions:
- https://github.com/openwisp/openwisp-utils#openwisp_utilstestscapture_stdout
- https://github.com/openwisp/openwisp-utils#openwisp_utilstestscapture_stderr
- https://github.com/openwisp/openwisp-utils#openwisp_utilstestscapture_any_output
4. Use them to suppress the output of the tests.
5. Checkout our commit guidelines: https://openwisp.io/docs/developer/contributing.html
6. Create a pull request.

priyanshuone6:
Hey, can you please assign this issue to me

nemesisdesign:
@priyanshuone6 just go for it

priyanshuone6:
I am getting a variety of outputs and wanted to ask which ones to suppress

```
System check identified no issues (0 silenced).
..................../home/priyanshu/anaconda3/envs/radius/lib/python3.6/site-packages/django/contrib/admin/options.py:801: UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'openwisp_users.models.User'> QuerySet.
  return self.paginator(queryset, per_page, orphans, allow_empty_first_page)
./home/priyanshu/anaconda3/envs/radius/lib/python3.6/site-packages/django/db/models/fields/__init__.py:1370: RuntimeWarning: DateTimeField RadiusAccounting.start_time received a naive datetime (2017-06-10 10:50:00) while time zone support is active.
  RuntimeWarning)
```
```
Generated a batch of users with prefix test-prefix6
```
```
WARNING:root:No corresponding user found for username: admin
..WARNING:root:No corresponding user found for username: username
....WARNING:root:No corresponding user found for username: admin
..WARNING:root:No corresponding user found for username: admin
.....WARNING:openwisp_radius.api.views:User test_name1 is not member of organization test-org
```
```
from: +595972157632
to: +237675582042
flash: False
test org verification code: 558134
```
```
WARNING:openwisp_radius.base.models:user test@openwisp.org has reached the maximum daily SMS limit
[...]

PR #208, opened 2020-12-23
[fix] decorate a function to reduce noises #83

priyanshuone6:
I used `./manage.py test openwisp_radius.tests.test_token.TestPhoneToken._create_token` to test the function to check if the decoraor is supressing the noise, but even after applying it I get the same output.

Before and after applying the decorator
```
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
/home/priyanshu/anaconda3/envs/radius/lib/python3.6/site-packages/django/db/models/fields/__init__.py:1312: RuntimeWarning: DateTimeField PhoneToken.created received a naive datetime (2020-12-22 00:00:00) while time zone support is active.
  RuntimeWarning)
/home/priyanshu/anaconda3/envs/radius/lib/python3.6/site-packages/django/db/models/fields/__init__.py:1312: RuntimeWarning: DateTimeField PhoneToken.created received a naive datetime (2020-12-23 00:00:00) while time zone support is active.
  RuntimeWarning)
from: +595972157632
to: +393664351808
flash: False
test org verification code: 507357
-------------------------------------------------------------------------------
.
Summary of slow tests (>0.3s)


Total slow tests detected: 0

----------------------------------------------------------------------
Ran 1 test in 0.161s

OK
Destroying test database for alias 'default'...
```

Issue #83

### DIFF
--- a/openwisp_radius/tests/test_token.py
+++ b/openwisp_radius/tests/test_token.py
@@ -12,6 +12,8 @@ from ..utils import load_model
 from . import _TEST_DATE
 from .mixins import BaseTestCase
 
+from openwisp_utils.tests import capture_any_output
+
 User = get_user_model()
 PhoneToken = load_model('PhoneToken')
 RadiusToken = load_model('RadiusToken')
@@ -37,6 +39,7 @@ class TestPhoneToken(BaseTestCase):
         radius_settings.sms_sender = '+595972157632'
         radius_settings.save()
 
+    @capture_any_output()
     def _create_token(
         self, user=None, ip='127.0.0.1', phone_number='+393664351808', created=None
     ):


nemesisdesign (commented):
I guess that this output is generated by the django logger and I'm not sure how to deal with it.

Have you tried decorating the specific test methods that generate more output?
The goal of this task is to eliminate as much output as we can, especially the standard error output which sometimes confuses new contributors, who think errors are being raised, while the tests trigger error on purpose and catch them in order to test these conditions.

priyanshuone6:
I tried decorating the specific test methods like in `openwisp_radius.tests.test_api.TestApi` but I am still the same output.

devkapilbansal:
@priyanshuone6 may be you have to decorate the SetUp function here.

priyanshuone6:
@devkapilbansal I tried doing as you said but I am still getting the same output.

PR #144, opened 2020-07-31
[fix] Fix hardcoded path in migration

atb00ker:
https://github.com/openwisp/openwisp-radius/pull/142 is merged but was incomplete and had problems in migrations, fixed it here.

PR #210, opened 2020-12-29
[feature] Added UserAuthenticationBackend to login api #209 

### LINKED ISSUE #209: [feature] allow email , username and phone_number in  user auth token api authentication
Modify the `ObtainAuthTokenView` api to use the `UserAuthenticationBackend` created in [openwisp_users](https://github.com/openwisp/openwisp-users/pull/208) for authentication

NoumbissiValere:
Closes #209

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -41,6 +41,11 @@ Organization = swapper.load_model('openwisp_users', 'Organization')
 User = get_user_model()
 
 
+class AllowAllUsersModelBackend(UsersAuthenticationBackend):
+    def user_can_authenticate(self, user):
+        return True
+
+
 class AuthTokenSerializer(BaseAuthTokenSerializer):
     """
     Recognizes also inactive users.
@@ -54,12 +59,13 @@ class AuthTokenSerializer(BaseAuthTokenSerializer):
         password = attrs.get('password')
 
         if username and password:
-            backend = UsersAuthenticationBackend()
-            try:
-                user = backend.get_users(username)[0]
-            except IndexError:
-                user = None
-            if not user or not user.check_password(password):
+            backend = AllowAllUsersModelBackend()
+            user = backend.authenticate(
+                request=self.context.get('request'),
+                username=username,
+                password=password,
+            )
+            if not user:
                 msg = _('Unable to log in with provided credentials.')
                 raise serializers.ValidationError(msg, code='authorization')
         else:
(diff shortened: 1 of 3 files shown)


NoumbissiValere:
This PR is waiting for [openwisp_users#206](https://github.com/openwisp/openwisp-users/pull/208) and it is needed by [openwisp-wifi-login-pages #144](https://github.com/openwisp/openwisp-wifi-login-pages/pull/145) for testing

openwisp_radius/api/serializers.py
```diff
@@ -53,13 +54,12 @@ def validate(self, attrs):
...
-                username=username,
-                password=password,
-            )
-            if not user:
+            backend = UserAuthenticationBackend()
+            try:
+                user = backend.get_users(username)[0]
+            except IndexError:
+                user = None
+            if not user or not user.check_password(password):
```
nemesisdesign, line 62: I don't understand why you don't simply maintain the same code structure as before?

```
backend = UserAuthenticationBackend()
user = backend.authenticate(...)
```

Why are you repeating the whole logic of the authenticate() method again?

nemesisdesign (changes requested):

openwisp_radius/api/serializers.py
NoumbissiValere: Please @nemesisdesign I don't think logic is been repeated. Please, maybe we have forgotten how this login api works or how the `backend.authenticate` works. 

- This login api authenticates both an `active and in active users` that is the reason why it was using [AllowAllUsersModelBackend](https://github.com/django/django/blob/master/django/contrib/auth/backends.py#L163) which authenticates both `active and inactive users`
- UsersAuthenticationBackend().authenticate ensures only active users can authenticate when it checked [can_user_authenticate](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/backends.py#L17)

So to ensure this login api still works as before while using the `UsersAuthenticationBackend` i think the logic i implemented is the best way to go

or what do you think?

nemesisdesign: Just create a new `AllowAllUsersModelBackend` which inherits the backend of openwisp-users and adds the [can_authenticate method as django does](https://github.com/django/django/blob/master/django/contrib/auth/backends.py#L164-L165).

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #212, opened 2021-01-02
[bug] Radius check and reply should check for organization membership…

### LINKED ISSUE #162: [bug] Radius check and reply should check for organization membership
How to replicate:

- create two orgs, A and B
- create one user, member of A
- try creating a radius check and a radius reply for this new user, but set the organization to B

Expected outcome: validation error should be raised.
Actual outcome: validation passes and item is saved.

Membership should be checked using the [organization membership helpers](https://github.com/openwisp/openwisp-users/#organization-membership-helpers).

lehone-hp:
… #162

Added custom validation rule to RadiusCheckForm which which checks if the selected user belongs to the selected organization.
Created a new form; RadiusReplyForm to validate that user belongs to selected orrganization for RadiusReply model

Fixes #162

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -155,6 +155,7 @@ _GET_OPTIONAL_FIELDS_HELP_TEXT = _(
 _REGISTRATION_ENABLED_HELP_TEXT = _(
     'Whether the registration API endpoint should be enabled or not'
 )
+_ORGANIZATION_HELP_TEXT = _('The user is not a member of this organization')
 
 
 class AutoUsernameMixin(object):
@@ -162,8 +163,19 @@ class AutoUsernameMixin(object):
         """
         automatically sets username
         """
+        if (
+            self.username
+            and User.objects.filter(username=self.username).exists()
+            and not self.user
+        ):
+            self.user = User.objects.get(username=self.username)
+
         if self.user:
             self.username = self.user.username
+            if hasattr(self, 'organization') and not self.user.is_member(
+                self.organization
+            ):
+                raise ValidationError({'organization': _ORGANIZATION_HELP_TEXT})
         elif not self.username:
             raise ValidationError(
                 {'username': _NOT_BLANK_MESSAGE, 'user': _NOT_BLANK_MESSAGE}
--- a/openwisp_radius/tests/test_models.py
+++ b/openwisp_radius/tests/test_models.py
@@ -81,11 +81,17 @@ class TestRadiusCheck(BaseTestCase):
         self.assertIsInstance(radiuscheck.pk, UUID)
 
     def test_auto_username(self):
+        org = self.default_org
         u = get_user_model().objects.create(
             username='test', email='test@test.org', password='test'
         )
+        self._create_org_user(organization=org, user=u)
         c = self._create_radius_check(
-            user=u, op=':=', attribute='Max-Daily-Session', value='3600'
+            user=u,
+            op=':=',
+            attribute='Max-Daily-Session',
+            value='3600',
+            organization=org,
         )
         self.assertEqual(c.username, u.username)
 
@@ -100,11 +106,17 @@ class TestRadiusCheck(BaseTestCase):
             self.fail('ValidationError not raised')
 
     def test_change_user_username(self):
+        org = self.default_org
         u = get_user_model().objects.create(
             username='test', email='test@test.org', password='test'
         )
+        self._create_org_user(organization=org, user=u)
         c = self._create_radius_check(
-            user=u, op=':=', attribute='Max-Daily-Session', value='3600'
+            user=u,
+            op=':=',
+            attribute='Max-Daily-Session',
+            value='3600',
+            organization=org,
         )
         u.username = 'changed'
         u.full_clean()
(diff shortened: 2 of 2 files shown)


lehone-hp:
@nemesisdesign for some reason my tests keep failing on this PR with coveralls so on my local machine, I ran `./runtests.py` which ran successfully but when I run `SAMPLE_APP=1 ./runtests.py`, it fails before and after my add my changes. Please am I missing something here?

openwisp_radius/base/models.py
```diff
@@ -318,6 +324,11 @@ class Meta:
     def __str__(self):
         return self.username
 
+    def clean(self):
+        super().clean()
+        if self.user and not self.user.is_member(self.organization):
+            raise ValidationError({'organization': _ORGANIZATION_HELP_TEXT})
```
nemesisdesign, line 330: we already have logic to deal with these cases in the rest of openwisp, take a look at openwisp-users and openwisp-controller.

nemesisdesign (changes requested):
Tests are missing.

@lehone-hp build is failing for an unrelated issue. Should be fixed soon: https://github.com/openwisp/openwisp-radius/pull/214.

openwisp_radius/base/models.py
nemesisdesign: Take a look at these methods: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/mixins.py
Then look in openwisp-controller for these to see how they're used.

lehone-hp: @nemesisdesign I have tried using the [_validate_org_relation](https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/mixins.py#L16) to check if the selected user is in the same organization as the  RadiusReply but I get the error, `'User' object has no attribute 'organization_id'`. 
Looking at the implementation of `_validate_org_relation`, it seems the relating object must have an `organization_id` field.

This is my implementation in models.py
```
    def clean(self):
        super().clean()
        if self.user:
            self._validate_org_relation('user') 
```

nemesisdesign: ok, you're right, please just avoid the duplication of this method, it's not necessary to define it in `AbstractRadiusReply` since this class inherits `AutoUsernameMixin` which has the same code. Moreover, please always avoid duplicating any code. Follow the DRY principle.

nemesisdesign (changes requested):
@lehone-hp please rebase on the current latest master and add a unit test for this solution and we'll be ready to merge this.

openwisp_radius/base/models.py
lehone-hp: Okay, thanks for the feedback.

Purhan (commented):
@lehone-hp Have you tested the changes manually? I tried the following:

# 1. While creating a new check/reply
- Create organizations `org_a` and `org_b`
- Create a user `user_a` for organization `org_a`
- Create a radius check for `user_a` but select `org_b` for the organization

Expected:
Validation should be raised

Result:
Validation is not raised, the check is saved.
Same behavior for radius reply

# 2. While editing a pre-existing check/reply
- For user `user_a`, if I change the organization to `org_b` then a validation is raised, however there is no indicator over the `organization` field and might be confusing why the validation error occurs (see screenshot below)
image

Purhan (approved):
~~@lehone-hp Ignore the issue above, I tried on a completely fresh environment today and found no issues, all changes LGTM~~
(Pardon, I don't have perms to dismiss this review...)

Purhan (commented):
> # 1. While creating a new check/reply
> 
>     * Create organizations `org_a` and `org_b`
> 
>     * Create a user `user_a` for organization `org_a`
> 
>     * Create a radius check for `user_a` but select `org_b` for the organization
> 
> 
> Expected:
> Validation should be raised
> 
> Result:
> Validation is not raised, the check is saved.

After more thorough testing on a completely fresh environment, the problem is fixed for `guided mode` but persists for `custom mode`, please confirm this when you can

<details>
<summary>Screenshot</summary>

</details>

lehone-hp:
@Purhan the validation for org and user works only for guided mode because in custom mode, the username field is free text and so any username can be inputted. From what you are saying, it means we will first have to check to see if the username provided during custom mode exists and if the user belongs to the same org but looking at the code, this check wasn't done so I assumed this is the expected behaviour except otherwise.

nemesisdesign (changes requested):
Sorry for my late reply @lehone-hp.
Thanks @Purhan for your review!

In the custom mode, if the username entered relates to an existing user in our DB, we should ensure the organization field of radius checks and radius replies matches with one of the organizations the user is member of.

Radius checks can be used also without having user accounts created in the system, therefore the custom mode is left for backward compatibility with this use case (think a radius check that authenticates a specific router, without the need of creating a user account for the router, which would be weird).

lehone-hp:
@nemesisdesign and @Purhan, I have updated my PR to handle the organization membership check in custom mode when a username ties with a user in the system. Please kindly review and provide feedback.

Purhan (approved):
Looks good to me :+1:, tested and it works as expected.

nemesisdesign (approved):
Geat work @lehone-hp :+1: 
Thanks @Purhan for reviewing and testing!

PR #213, opened 2021-01-03
[qa] Split free_radius views and test_api #196

### LINKED ISSUE #196: [qa] Split huge views and test_api
The api views are huge, we should divide the API endpoints dedicated to `freeradius` from the other ones.

The `test_api.py` file is also huge, we should split it in multiple files, if it works we may use a subfolder, eg:

- test_api/test_freeradius_api.py
- test_api/test_registration.py
- test_api/test_rest_token.py
- test_api/test_phone_token.py

We should group related tests in specific files, we should not aim to have too many files, but we should aim to keep the length of each file lower than 1000 lines and ideally lower than 700 lines.

devkapilbansal:
Closes #196


nemesisdesign (changes requested):
There's some conflicts, please resolve them when you can.

openwisp_radius/api/free_radius/views.py
```diff
@@ -0,0 +1,230 @@
+import logging
```
nemesisdesign, line 1: let's simplify: please move the code to `openwisp_radius/api/freeradius_views.py` instead of `openwisp_radius/api/free_radius/views.py`.

Better to have both view files at the same level.

nemesisdesign (changes requested):
Good progress @devkapilbansal :+1: 

I found a discrepancy of the number of tests run between master and this branch:

- master
  - regular tests: 303
  - sample app tests: 381
- this branch:
  - regular tests: 303
  - sample app tests: 377

Can you find out why there's a difference of 4 tests in the sample app test?

To run both tests suites use:
```
./runtests.py --parallel && SAMPLE_APP=1 ./runtests.py --parallel
```

nemesisdesign (commented):
PS: let's wait for these PRs to be merged before proceeding with this:

- #220 and #222
- #221

nemesisdesign (approved):
Thanks a lot @devkapilbansal :+1:

PR #237, opened 2021-02-23
[feature] Add identity verification #233

### LINKED ISSUE #233: [feature] Implement explicit identity verification mechanism to allow banning users in OpenWISP RADIUS
The current implementation relies on `is_active` for knowing if a user is verified or not, but this prevents us from banning users, because inactive users can verify their mobile phone again.

We need to implement this as it was implemented in OpenWISP1:

- `is_active=False` means the user cannot log in 
- another field will let us know if the user has verified their identity

Therefore an user who has verified their identity but has `is_active=False` has been banned for illicit activity.

## RegisteredUser model

:heavy_check_mark: We should add a new model in openwisp-radius, called `RegisteredUser` which binds one to one with the User model. 

:heavy_check_mark: Records of this DB table will be created only when users self registers via OpenWISP RADIUS.

This model will have the fields described below.

## `identity_verification` field

:heavy_check_mark: `identity_verification`, char field, defaults to NULL, max length 64 characters and will be readonly in the admin. It should be a choice field, but **the choices should be dynamic (eg: returned by a function) and additional choices can be registered via python code like we do with notification types**.

The default choices will be:

:heavy_check_mark:   - No identity verification (default)
:heavy_check_mark:  - Mobile phone

:heavy_check_mark: The identity verification field shall be filled correctly during sign up: if the SMS verification is enabled and the mobile phone is present, the mobile phone option will be chosen, otherwise it will fallback to the default.

## `is_verified` field

:heavy_check_mark: We should also add a boolean field called to this new model, which is called`is_verified` which defaults to False, which tells us if the user has verified their account with one of the accepted identity verification methods.
This flag will have to be set to True once the user completes the verification process (eg: mobile verification with SMS token).

### OrganizationRadiusSettings.needs_identity_verification

:heavy_check_mark: Since the identity verification methods are extensible and we can't know beforehand which ones will be activated, we need a generic way to know if we need to require users to have verified their identity before letting them in. 

For this reason we shall add a boolean field called `needs_identity_verification` to `OrganizationRadiusSettings`, which defaults to `False` but will be set as `True` if `sms_verification` is enabled.
This field shall be implemented exactly as https://github.com/openwisp/openwisp-radius/issues/98, we should have a general setting which can be overridden per org.

Other modules or custom app extending openwisp-radius will also make sure to set it to True when additional identity verification methods are enabled for an org.

## Update the rest of the system accordingly

In order to make these change effective, we will have to update code an tests in the system.

:heavy_check_mark: Here we have to add `RegisteredUser.is_verified=True`:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L891-L893

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L592

:heavy_check_mark: Here we can use `needs_identity_verification` instead of `sms_verification_enabled`:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/serializers.py#L387-L388

:heavy_check_mark: In the following lines we have to replace `is_active` with `RegisteredUser.is_verified`:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L592

:heavy_check_mark: Here, if identitfy verification is needed, we have to modify the filter to look for verified users, eg:

```python
filter_kwargs = dict(is_active=True)
if org_settings.needs_identity_verification:
    filter_kwargs['registereduser__is_verified'] = True

 user = auth_backend.get_users(username).filter(**filter_kwargs)[0]
```

:heavy_check_mark:
https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L237

:heavy_check_mark:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L629 

:heavy_check_mark:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L646

:heavy_check_mark: Update the comments:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L927-L929

:heavy_check_mark: Here we have to make sure that if `needs_identity_verification` is `True`, we also need to check `RegisteredUser.is_verified` before letting the user log in:

https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L594

Purhan:
Closes #233


Purhan:
Not sure why the build is failing but atleast all the tests passed. I will write tests for the new settings now.

nemesisdesign (commented):
@Purhan looking at the build output, even though for some reason it doesn't show which command is failing, I think is this line: https://github.com/openwisp/openwisp-radius/blob/master/.github/workflows/ci.yml#L58

Because the output is typical of when failure happens in parallel testing mode, which we do only when testing the sample app. Probably migrations for the sample app are missing, try running test and the application with SAMPLE_APP=1.

Purhan:
Blocked by https://github.com/openwisp/openwisp-users/pull/232

openwisp_radius/models.py
```diff
@@ -97,3 +102,23 @@ class PhoneToken(AbstractPhoneToken):
...
+        RegisteredUser.objects.create(user=instance)
+
+
+# Create a new RegisteredUser object for every user
+signals.post_save.connect(
+    auto_create_registered_user,
+    sender=User,
+    weak=False,
+    dispatch_uid='models.auto_create_registered_user',
+)
```
nemesisdesign, line 124: I don't think we should do this, we want `RegisteredUser` objects be created only for users who register through the registration API.

tests/openwisp2/sample_radius/models.py
```diff
@@ -87,3 +92,22 @@ class Meta(AbstractOrganizationRadiusSettings.Meta):
...
+        RegisteredUser.objects.create(user=instance)
+
+
+# Create a new RegisteredUser object for every user
+signals.post_save.connect(
+    auto_create_registered_user,
+    sender=User,
+    weak=False,
+    dispatch_uid='models.auto_create_registered_user',
+)
```
nemesisdesign, line 113: same here

openwisp_radius/base/models.py
```diff
@@ -1335,3 +1348,23 @@ def __check(self, token):
...
+
+
+class AbstractRegisteredUser(models.Model):
+    user = models.OneToOneField(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    identity_verification = models.CharField(
+        max_length=64,
+        default=None,
+        blank=True,
+        null=True,
+        choices=IDENTITY_VERIFICATION_CHOICES,
```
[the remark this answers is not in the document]
nemesisdesign: what you're doing can be enough, you just need a way to register a new identity verification method and document how to do it, for example, a custom django app may implement a flow which verifies users through a National Digital ID card, then we need to provide a function that adds an entry to `IDENTITY_VERIFICATION_CHOICES` and we need to document this in the README

nemesisdesign (commented):
Thanks for your patience @Purhan and sorry it's taken so long to review this.
I left a couple of comments and I will test it more tomorrow or in the next days.

openwisp_radius/api/freeradius_views.py
```diff
@@ -173,7 +174,11 @@ def get_user(self, request, username):
         return active user or ``None``
         """
         try:
-            user = auth_backend.get_users(username).filter(is_active=True)[0]
+            filter_kwargs = dict(is_active=True)
+            organization = Organization.objects.get(pk=request._auth)
+            if self._needs_identity_verification(organization):
+                filter_kwargs['registereduser__is_verified'] = True
```
nemesisdesign, line 180: I'm not convinced of the implementation, because it will generate multiple queries, at least 2 (one to get the org and 1 to get the org radius settings).

openwisp_radius/api/views.py
```diff
@@ -374,9 +379,19 @@ def post(self, request, *args, **kwargs):
...
         serializer = self.serializer_class(instance=token, context=context)
-        response = {'radius_user_token': radius_token.key, 'is_active': user.is_active}
+        response = {
+            'radius_user_token': radius_token.key,
+            'is_active': user.is_active,
+            'is_verified': self._is_user_verified(user),
+        }
         response.update(serializer.data)
         status_code = 200 if user.is_active else 401
+        organization = Organization.objects.get(slug=kwargs['slug'])
```
nemesisdesign, line 389: here too.
I guess we could move the query which retrieves the org in the _needs_identity_verification() method, passing it the keyword arguments to get.

openwisp_radius/api/utils.py
```diff
@@ -15,6 +17,20 @@ def _get_error_dict(self, error):
         return dict_
 
 
+class IDVerificationHelper(object):
+    def _needs_identity_verification(self, org):
```
nemesisdesign, line 21: instead of org we could pass the keyword arguments to filter the org,
then we could use the following queryset:

```python
Organization.objects.select_related(
    'radius_settings'
).get(**organization_filter_kwargs)
```

openwisp_radius/api/views.py
```diff
@@ -335,6 +336,8 @@ class RegisterView(RadiusTokenMixin, DispatchOrgMixin, BaseRegisterView):
 
     def get_response_data(self, user):
         data = super().get_response_data(user)
+        # create a RegisteredUser object for every user that registers through API
+        RegisteredUser.objects.create(user=user)
```
nemesisdesign, line 340: Not convinced of this.
How do we know which identity verification method the user is registering with?
Maybe it makes sense to add this to the registration serializer, so the user will select the method from the captive page and we'll get it here without the need of adding complicated code, what do you think?

openwisp_radius/migrations/0023_add_identity_verification.py
```diff
@@ -0,0 +1,68 @@
...
+                        verbose_name='ID',
+                    ),
+                ),
+                (
+                    'identity_verification',
+                    models.CharField(
+                        blank=True,
+                        choices=[
+                            (None, 'No Identity Verification'),
+                            ('mobile', 'Mobile Phone'),
```
nemesisdesign, line 45: let's put `IDENTITY_VERIFICATION_CHOICES` here instead of the raw choices, to avoid having to generate a migration every time something registers a new method

tests/openwisp2/sample_radius/migrations/0022_add_identity_verification.py
```diff
@@ -0,0 +1,68 @@
...
+                        verbose_name='ID',
+                    ),
+                ),
+                (
+                    'identity_verification',
+                    models.CharField(
+                        blank=True,
+                        choices=[
+                            (None, 'No Identity Verification'),
+                            ('mobile', 'Mobile Phone'),
```
nemesisdesign, line 45: same here

openwisp_radius/migrations/0023_add_identity_verification.py
```diff
@@ -0,0 +1,68 @@
...
+        ('openwisp_radius', '0022_organizationradiussettings_registration_enabled'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='organizationradiussettings',
+            name='needs_identity_verification',
+            field=models.BooleanField(
+                blank=True,
+                default=False,
```
nemesisdesign, line 21: the default should be `None` otherwise it will always default to False even though the global setting is changed.

tests/openwisp2/sample_radius/models.py
```diff
@@ -14,8 +15,11 @@
     AbstractRadiusReply,
     AbstractRadiusToken,
     AbstractRadiusUserGroup,
+    AbstractRegisteredUser,
 )
 
+User = get_user_model()
```
nemesisdesign, line 21: what's this for?

openwisp_radius/models.py
```diff
@@ -16,10 +17,13 @@
     AbstractRadiusReply,
     AbstractRadiusToken,
     AbstractRadiusUserGroup,
+    AbstractRegisteredUser,
 )
 
 logger = logging.getLogger(__name__)
 
+User = get_user_model()
```
nemesisdesign, line 25: what's this for?

openwisp_radius/admin.py
```diff
@@ -498,7 +504,22 @@ def has_delete_permission(self, request, obj=None):
...
+
+    def has_delete_permission(self, request, obj=None):
+        return False
+
+
+UserAdminInline.inlines += [
+    RadiusUserGroupInline,
+    PhoneTokenInline,
+    RegisteredUserInline,
+]
```
nemesisdesign, line 522: I was trying to do some improvements to this file:

```diff
diff --git a/openwisp_radius/admin.py b/openwisp_radius/admin.py
index b5a78de..bcb3c43 100644
--- a/openwisp_radius/admin.py
+++ b/openwisp_radius/admin.py
@@ -9,8 +9,7 @@ from django.forms.widgets import Select
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
-from openwisp_users.admin import OrganizationAdmin
-from openwisp_users.admin import UserAdmin as UserAdminInline
+from openwisp_users.admin import OrganizationAdmin, UserAdmin
 from openwisp_users.multitenancy import MultitenantAdminMixin, MultitenantOrgFilter
 from openwisp_utils.admin import (
     AlwaysHasChangedMixin,
@@ -507,18 +506,16 @@ class PhoneTokenInline(TimeReadonlyAdminMixin, StackedInline):
 class RegisteredUserInline(StackedInline):
     model = RegisteredUser
     form = AlwaysHasChangedForm
-
-    def has_add_permission(self, request, obj):
-        return False
+    extra = 0
 
     def has_delete_permission(self, request, obj=None):
         return False
 
 
-UserAdminInline.inlines += [
+UserAdmin.inlines.insert(0, RegisteredUserInline)
+UserAdmin.inlines += [
     RadiusUserGroupInline,
     PhoneTokenInline,
-    RegisteredUserInline,
 ]
 
 
@@ -663,7 +660,7 @@ if app_settings.SOCIAL_LOGIN_ENABLED:
         def has_delete_permission(self, request, obj=None):
             return False
 
-    UserAdminInline.inlines += [SocialAccountInline]
+    UserAdmin.inlines += [SocialAccountInline]
 
     Token = apps.get_model('authtoken', 'Token')
```

But one test fails in the SAMPLE_APP:

```
======================================================================
FAIL: test_admin_change_non_superuser_readonly_fields (openwisp2.sample_users.tests.TestUsersAdmin) [User Permissions]
----------------------------------------------------------------------
  File "/home/nemesis/Code/openwisp/openwisp-users/openwisp_users/tests/test_admin.py", line 239, in test_admin_change_non_superuser_readonly_fields
    self.assertContains(response, html)
  File "/home/nemesis/.virtualenvs/openwisp2/lib/python3.8/site-packages/django/test/testcases.py", line 470, in assertContains
    self.assertTrue(real_count != 0, msg_prefix + "Couldn't find %s in response" % text_repr)
AssertionError: False is not true : Couldn't find '<div class="readonly">sample_users' in response

```
Any idea why?

nemesisdesign (changes requested):
Great progress @Purhan!

I gave it a deeper look today and did some basic testing via the admin. 

Got this warning while running qa checks:
```
settings.rst:413: WARNING: Unexpected indentation. looking
```

I will need to do more testing via the registration API, but for the moment see my comments below.

openwisp_radius/admin.py
Purhan: Probably the same reason as https://github.com/openwisp/openwisp-users/pull/232, did you update `openwisp-users` in your local repo?

openwisp_radius/api/freeradius_views.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/ef3686e4dbe8c5e85a5bec1769da36eaafd87ea8)

openwisp_radius/api/views.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/ef3686e4dbe8c5e85a5bec1769da36eaafd87ea8)

openwisp_radius/api/utils.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/ef3686e4dbe8c5e85a5bec1769da36eaafd87ea8)

openwisp_radius/migrations/0023_add_identity_verification.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/ef3686e4dbe8c5e85a5bec1769da36eaafd87ea8)

openwisp_radius/migrations/0023_add_identity_verification.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/c116994db22bca1cfb26a1a82fd26371bf263d96)

tests/openwisp2/sample_radius/migrations/0022_add_identity_verification.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/c116994db22bca1cfb26a1a82fd26371bf263d96)

docs/source/user/settings.rst
```diff
@@ -399,6 +399,38 @@ then edit a specific organization and scroll down to
...
+Indicates whether organizations require a user to be verified in order to login.
+This can be overridden globally or for each organization separately via the admin
+interface.
+
+If this is enabled, each registered user should be verified using a verification method,
+by default, the following choices are available:
+  - No Identity Verification (When verification is disabled)
+  - Mobile Phone
+
+A new verification method can be added or an existing method can be removed using the
```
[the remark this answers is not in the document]
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/c116994db22bca1cfb26a1a82fd26371bf263d96)

openwisp_radius/api/views.py
Purhan: Done (https://github.com/openwisp/openwisp-radius/pull/237/commits/f08a1ad5ef672c1d1120682b6f585c551ee2ee42)
Now I need to write some unit tests... working on it.

tests/openwisp2/sample_radius/migrations/0022_add_identity_verification.py
```diff
@@ -4,6 +4,8 @@
 from django.conf import settings
 from django.db import migrations, models
 
+from openwisp_radius.verification_methods import IDENTITY_VERIFICATION_CHOICES
+
```
nemesisdesign, line 8: can you please remove this extra blank line?

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Great progress @Purhan!

I made some improvements (which will push shortly) and have done some testing.

### Permissions

I realized we may be missing to add permissions for the groups operator and administrator, can you please add the following?

Administrators have all permissions.
Operators only have view.

### Testing the registration API

I created a user with an org which has SMS verification enabled.
Then tried to verify the account (using the API endpoints to create the SMS code and to verify it), but at the point of verifying the phone the API returns:

```json
{
  "non_field_errors": [
    "This user is already active."
  ]
}
```

I think the validation of PhoneToken should check whether the user is already verified, rather than already active.

### More tests to be made with the API

I need more time to test it further, I should be able to do this in the next 2 weeks.

For now, I can ask you questions like:

- what happens if we try to register a user with an org which has "needs_identity_verification" set to True but the value "no identity verification" is sent for the identity_verification field in the POST request of the registration?
- what happens if we try to register a user with an org which has "needs_identity_verification" set to False but "Mobile Phone" is sent for the identity_verification field in the POST request?

nemesisdesign (commented):
Only while testing I realized we will need to show this field in the admin, I made the changes to implement this in an efficient way, please double check, here's how it looks:

Screenshot from 2021-03-20 15-39-36

Now I realize we will surely need to filter by this field as well, can you add a filter which allows to filter by the 3 possible values?

The values are: yes, no and unknown.
Unknown means there's no `RegisteredUser` record related to the user, it can be implemented with a [custom admin filter like the one recently added to openwisp-controller which allows to filter devices which do not have a geographic location defined](https://github.com/openwisp/openwisp-controller/blob/80bb8a08ba827c5a19892ed8b75129fcc25c389c/openwisp_controller/geo/admin.py#L84-L97).

Let me know!

Purhan:
> Now I realize we will surely need to filter by this field as well, can you add a filter which allows to filter by the 3 possible values?
> 
> The values are: yes, no and unknown.
> Unknown means there's no `RegisteredUser` record related to the user, it can be implemented with a [custom admin filter like the one recently added to openwisp-controller which allows to filter devices which do not have a geographic location defined](https://github.com/openwisp/openwisp-controller/blob/80bb8a08ba827c5a19892ed8b75129fcc25c389c/openwisp_controller/geo/admin.py#L84-L97).
> 
> Let me know!

Done, added this (https://github.com/openwisp/openwisp-radius/pull/237/commits/3fac732c4484c1053b306948ee9bda61ecd2bfea).

---------------------------------------------------------------------------------------------------------------------------------------------------------

> - what happens if we try to register a user with an org which has "needs_identity_verification" set to True but the value "no identity verification" is sent for the identity_verification field in the POST request of the registration?

The API would return bad request and the user object is not created:
image

> - what happens if we try to register a user with an org which has "needs_identity_verification" set to False but "Mobile Phone" is sent for the identity_verification field in the POST request?

The RegisteredUser object is created and the user is registered however `is_verified` is false at the time of creation.
image

openwisp_radius/tests/test_admin.py
```diff
@@ -1129,8 +1129,57 @@ def test_get_is_verified_user_admin_list(self):
...
+        url = reverse(f'admin:{app_label}_user_changelist')
+
+        def get_expected_html(value):
+            return (
+                '<td class="field-get_is_verified"><img src="/static/admin/img/icon-'
+                f'{value}.svg" alt="{value}">'
+            )
+
+        with self.subTest('Verified users filter'):
+            response = self.client.get(f'{url}?is_verified=true')
```
nemesisdesign, line 1172: Try passing the querystring parameters as a dict, it should work:
```python
self.client.get(url, {'is_verified': 'true'})
```

It's a lot cleaner. It's better to use this form (see also https://github.com/openwisp/openwisp-users/issues/242).

nemesisdesign (changes requested):
> > Now I realize we will surely need to filter by this field as well, can you add a filter which allows to filter by the 3 possible values?
> > The values are: yes, no and unknown.
> > Unknown means there's no `RegisteredUser` record related to the user, it can be implemented with a [custom admin filter like the one recently added to openwisp-controller which allows to filter devices which do not have a geographic location defined](https://github.com/openwisp/openwisp-controller/blob/80bb8a08ba827c5a19892ed8b75129fcc25c389c/openwisp_controller/geo/admin.py#L84-L97).
> > Let me know!
> 
> Done, added this ([3fac732](https://github.com/openwisp/openwisp-radius/commit/3fac732c4484c1053b306948ee9bda61ecd2bfea)).

Great :+1:

> > * what happens if we try to register a user with an org which has "needs_identity_verification" set to True but the value "no identity verification" is sent for the identity_verification field in the POST request of the registration?
> 
> The API would return bad request and the user object is not created:
> image

Great :+1:


> > * what happens if we try to register a user with an org which has "needs_identity_verification" set to False but "Mobile Phone" is sent for the identity_verification field in the POST request?
> 
> The RegisteredUser object is created and the user is registered however `is_verified` is false at the time of creation.
> image
> 
> ~P.S.: I am also writing tests for these cases~ done ([10210d5](https://github.com/openwisp/openwisp-radius/commit/10210d5279168143982417eef643611e6af5801b))

The important thing is that the phone number does not get saved if SMS verification is not enabled, but this should be already implemented, please double check.
 
> > ### Testing the registration API
> > I created a user with an org which has SMS verification enabled.
> > Then tried to verify the account (using the API endpoints to create the SMS code and to verify it), but at the point of verifying the phone the API returns:
> > ```json
> > {
> >   "non_field_errors": [
> >     "This user is already active."
> >   ]
> > }
> > ```
> > 
> > 
> > I think the validation of PhoneToken should check whether the user is already verified, rather than already active.
> 
> Fixed this in [9e3239a](https://github.com/openwisp/openwisp-radius/commit/9e3239ab499a12e52feca293bad9dfac90d3819d)

I tried again, that problem is fixed, thank you!
Unfortunately, the `is_verified` flag was not saved and the user stayed unverified.
By looking at the code it seems you aren't calling `user.registered_user.save()` and therefore the `is_verified=True` bit does not gets committed to the DB. Can you please write a test for this?

openwisp_radius/admin.py
```diff
@@ -629,9 +689,9 @@ def has_add_permission(self, request, obj):
         def has_delete_permission(self, request, obj=None):
             return False
 
-    UserAdminInline.inlines += [SocialAccountInline]
+    UserAdmin.inlines += [SocialAccountInline]
 
-    Token = apps.get_model('authtoken', 'Token')
+    Token = apps.get_model('authtoken', 'TokenProxy')
```
Purhan, line 694: Changed because of an unrelated issue

Purhan:
> Unfortunately, the is_verified flag was not saved and the user stayed unverified.
By looking at the code it seems you aren't calling user.registered_user.save() and therefore the is_verified=True bit does not gets committed to the DB. Can you please write a test for this?

Fixed it (https://github.com/openwisp/openwisp-radius/pull/237/commits/a453dc217fd060c5e287d080f891fbd024c4ffa6), I didn't write a new test, just changed an existing test, tell me if thats fine.
The build failed because of an outage in coveralls, I think

nemesisdesign (changes requested):
Great progress @Purhan :+1: !

Here's some issues found during testing.

### RegisteredUser model using autoincrement ID instead of UUID

I want to avoid autoincrement ID as much as possible, so I made sure the model inherits `UUIDModel`.

I fixed it in ca70b83, please double check.

### RegisteredUser model was raising an exception when `clean()` was called

Why did you inherit from AutoUsernameMixin? The logic of that mixin was failing because there's no `username` field on this model.

I fixed it in ca70b83, please double check.

### ChangePhoneNumberSerializer shall make the user inverified instead of inactive

Good that I wrote detailed notes in the issue description.
I reviewed each one and we did almost everything, except this part:

> :x: Update the comments:
> 
> https://github.com/openwisp/openwisp-radius/blob/79fcc128dd4b4b5c743cfaabf6d61ab2ebd57135/openwisp_radius/api/views.py#L927-L929

By reading this bit of information, I realized that the issue description was not very clear about this.
If you look at `ChangePhoneNumberSerializer`, you will see the `save` method makes the user inactive, that's because we do not want somebody to change their phone number and leave it unverified while at the same time continue to use the system.

However, since with this change we consider an inactive user a banned one, flagging the user inactive because they changed their phone number does not make sense, we should instead flag them as unverified.

So please change that comment, the save method of `ChangePhoneNumberSerializer` and its test accordingly.

openwisp_radius/api/views.py
```diff
@@ -686,9 +701,12 @@ def post(self, request, *args, **kwargs):
         if not is_valid:
             return self._error_response(_('Invalid code.'))
         else:
+            user.registered_user.is_verified = True
+            user.registered_user.identity_verification = 'mobile'
             user.is_active = True
             user.phone_number = phone_token.phone_number
             user.save()
+            user.registered_user.save()
```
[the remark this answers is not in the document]
Purhan: Done

openwisp_radius/api/serializers.py
```diff
@@ -334,6 +339,14 @@ class RegisterSerializer(
     last_name = serializers.CharField(required=False)
     location = serializers.CharField(required=False)
     birth_date = serializers.DateField(required=False)
+    identity_verification = serializers.ChoiceField(
+        help_text=(
+            'Required only when the organization has mandatory identity '
```
nemesisdesign, line 344: flag this text as translatable

nemesisdesign (changes requested):

nemesisdesign (approved):
Looks good, I will spend some more time testing it because it's an important change, so let's leave it open for a bit more.

Thanks @Purhan :+1: :+1:

nemesisdesign (commented):
Status update: @codesankalp is working to implement support for these changes in openwisp-wifi-login-pages.

@Purhan could you fix the conflicts please?

nemesisdesign (commented):
First test with openwisp-wifi-login-pages successful: https://github.com/openwisp/openwisp-wifi-login-pages/pull/225#pullrequestreview-657409567 :+1:

nemesisdesign (changes requested):
@Purhan testing intensively with https://github.com/openwisp/openwisp-wifi-login-pages/pull/225.
You can do the same if you want to double check my testing.

This is what I found during testing.

### Mobile verification: sign up, log out, cannot log in again

When testing sign up with mobile verification enabled, try the following:

- sign up, wait to get to the verification page in which you have to add the
  the one time phone token which is sent via SMS
  (printed in the console of openwisp-radius during development)
- log out
- now try to log in again

**Expected result**:

- the user should be taken to the verify page again

**Actual result**:

- the user cannot log in because their account is flagged as `is_active=False`

Maybe we should always flag new accounts as `is_active=True`?

openwisp_radius/tests/test_api/test_phone_verification.py
```diff
@@ -81,7 +81,7 @@ def test_register_201(self):
         user = User.objects.get(email=self._test_email)
         self.assertTrue(user.is_member(self.default_org))
         self.assertEqual(user.phone_number, phone_number)
-        self.assertFalse(user.is_active)
+        self.assertTrue(user.is_active)
```
nemesisdesign, line 84: Good that we spotted it!
So if here we were expecting the is_active to be False, it means the user was not verified, so can you please add another assertion to ensure that?
Should apply to the other occurrences as well.

nemesisdesign (changes requested):
@Purhan spotted another small thing, see below.
Can you update on the latest master again please?

nemesisdesign (approved):
I resolved the outstanding issues, thanks!

PR #239, opened 2021-03-11
[radius/registration] Email Verification added when sms verification is enabled #118

### LINKED ISSUE #118: [radius/registration] Support email verification when sms verification is enabled
Right now when an organization has enabled the sms verification, if a user registers, no verification email will be sent.

This was a shortcut to avoid flagging the user as `is_active` if they verify their email address but not their phone number (by using the sms token verification) when signing up for a wifi service.

Eg: users sign up, they may have data active on their phone and will receive the email verification, verify the email address, skip the SMS verification, open the captive page in an incognito window and be able to login to the wifi service and pass the authentication.

We want to avoid this.

The ideal solution would be to just make sure verifying the email address does not flag the user as `is_active`.

codesankalp:
Used all-auth EmailAddress model in custom_signup to send_confirmation email.

Implemented the ideal solution: Verifying email address does not flag the
user as `is_active`.

Fixes #118

I am not sure about this is the best way but it implements the ideal solution mentioned in the issue.

Screenshot: 

Detail: Email received after making a POST request on the API endpoint for creating the user.

image

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -3,6 +3,7 @@ import logging
 import phonenumbers
 import swapper
 from allauth.account.adapter import get_adapter
+from allauth.account.models import EmailAddress
 from allauth.account.utils import setup_user_email
 from dj_rest_auth.registration.serializers import (
     RegisterSerializer as BaseRegisterSerializer,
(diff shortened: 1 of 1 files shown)


Saurav-Shrivastav (approved):

codesankalp:
Closing this as the issue is fixed (when the registration methods feature is added).

PR #241, opened 2021-03-28
[storage] Allow changing base_url for docker-openwisp

atb00ker:
Update:
Made firmware variable for generic here: https://github.com/openwisp/openwisp-firmware-upgrader/pull/137


devkapilbansal (changes requested):
@atb00ker probably we need to change this here too 
https://github.com/openwisp/openwisp-radius/blob/0e13154419982663d2c0ef31c417841e61605110/openwisp_radius/urls.py#L20

atb00ker:
> @atb00ker probably we need to change this here too

I don't remember, but last I checked the link was working fine..! :smile:
I am not sure If I understood you correctly, could you please elaborate what change is required here?

docs/source/user/settings.rst
```diff
@@ -204,6 +204,16 @@ API and user token related settings
 
 These settings control details related to the API and the radius user token.
 
+``OPENWISP_RADIUS_CSV_BASEURL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign, line 208: what does not convince me is the fact that we're adding the base url only for the CSV files, why is that?
And why for the fw-upgrader module the setting is more generic?

nemesisdesign (commented):
I have one doubt, see below

devkapilbansal:
> > @atb00ker probably we need to change this here too
> 
> I don't remember, but last I checked the link was working fine..! 
> I am not sure If I understood you correctly, could you please elaborate what change is required here?

Shouldn't this be write like this:
` f'{app_settings.RADIUS_CSV_BASEURL}/<path:csvfile>', `

docs/source/user/settings.rst
atb00ker: Even for Firmware upgrader the name I've given is `OPENWISP_FIRMWARE_IMAGE_BASEURL`,
I'll need to make it generic, but starting here to give a generic name! :smile:

atb00ker:
> Shouldn't this be written like this:
f'{app_settings.RADIUS_CSV_BASEURL}/<path:csvfile>',

@devkapilbansal 
I might be missing some use case, please correct me, but from what I see:
1. When I have a different API server, why would I even need `get_api()` in the main admin dashboard server.
2. Even if I suppose we do need it for some reason, why should we allow it to be changed from settings unlike other URLs? It would be an anti-pattern that all the other urls are manipulated from `urls.py` and this one from `settings.py`, let the user edit `urls.py`. 

I thought a lot about it, but I can't come up with a usecase for changing this or where it can be used! :smile:

devkapilbansal:
> I thought a lot about it, but I can't come up with a usecase for changing this!

I thought it was left by mistake 

> 2. Even if I suppose we do need it for some reason, why should we allow it to be changed from settings unlike other URLs? It would be an anti-pattern that all the other urls are manipulated from `urls.py` and this one from `settings.py`, let the user edit `urls.py`.

:+1:

docs/source/user/settings.rst
```diff
@@ -204,6 +204,16 @@ API and user token related settings
 
 These settings control details related to the API and the radius user token.
 
+``OPENWISP_RADIUS_API_BASEURL``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Default**: ``/radiusbatch/csv/``
```
nemesisdesign, line 210: are you sure this is still correct?

openwisp_radius/settings.py
```diff
@@ -26,6 +26,7 @@ def get_settings_value(option, default):
 
 
 RADIUS_API = get_settings_value('API', True)
+RADIUS_API_BASEURL = get_settings_value('API_BASEURL', '/radiusbatch/csv/')
```
nemesisdesign, line 29: I don't understand, the default API BASE URL is `/radiusbatch/csv/'`?
It's really confusing.

openwisp_radius/base/models.py
```diff
@@ -803,7 +803,7 @@ def __str__(self):
 
 
 _get_csv_file_private_storage = PrivateFileSystemStorage(
-    location=settings.PRIVATE_STORAGE_ROOT, base_url='/radiusbatch/csv/',
+    location=settings.PRIVATE_STORAGE_ROOT, base_url=app_settings.RADIUS_API_BASEURL,
```
nemesisdesign, line 806: can't this stay the way it is and then you can adapt the docker deployment to have:

`OPENWISP_RADIUS_API_BASEURL = 'https://radius.domain.com' `

And serve the csv files from `https://radius.domain.com/radiusbatch/csv/`

nemesisdesign (commented):
I still have doubts

openwisp_radius/base/models.py
atb00ker: Okay, so you want the after host origin part to be calculated as per `urls.py` ? 
That would make sense, let me look into that! :smile:

atb00ker:
Updated! :smile:

nemesisdesign (approved):
Thanks Ajay :+1:.

PR #254, opened 2021-05-14
[change] Added model fields to manage fallback values #230

### LINKED ISSUE #230: [change] Add model fields to manage fallback values and reuse across models and admin
We've been adding different fields to `AbstractOrganizationRadiusSettings` which if NULL default to the value of a specific setting.

We introduced a special admin form field to deal with this case, so we can avoid having to write redundant code like:

```python
if org_settings.api_enabled is not None:
    api_enabled = org_settings.api_enabled
else:
    api_enabled = app_settings.API_ENABLED
```

But this works only in the admin, so in all the non admin code, the redundant and ugly code above is needed.

We should fix this.

We should change the model field to work similarly to the fields used in the admin, I am not 100% sure if the same solution can be reused, maybe, maybe not, we'll have to investigate this. If the same solution can be used, we should just move the fallback fields to their own `fields.py` file and reuse them, then we should clean all the redundant code like the one above (any code that checks if the value is `None` and if its `None` it supplies the fallback setting value)

niteshsinha17:
@nemesisdesign 
Please check the implementation once. 
Its working fine but I am facing one problem.
```
if org_settings.api_enabled is not None:
    api_enabled = org_settings.api_enabled
else:
    api_enabled = app_settings.API_ENABLED
```
How we can return fallback back values in this situation using those fields? Can you please suggesting something. I was thinking to create a model method to get setting value and will return fallback value if it is None. Is there any other way?

closes #230


openwisp_radius/base/models.py
```diff
@@ -1040,6 +1042,12 @@ def __str__(self):
         return self.key
 
 
+def _enabled_disabled_helper():
+    if app_settings.REGISTRATION_API_ENABLED:
+        return _('Enabled')
+    return _('Disabled')
```
nemesisdesign, line 1048: is this thing really used? It doesn't seem to me.

openwisp_radius/tests/test_admin.py
```diff
@@ -478,43 +476,47 @@ def test_organization_radsettings_freeradius_allowed_hosts(self):
...
+        #         {'radius_settings-0-freeradius_allowed_hosts': '123.246.512.12'}
+        #     )
+        #     response = self.client.post(url, form_data)
+        #     self.assertContains(
+        #         response,
+        #         _(
+        #             'Invalid input. Please enter valid ip addresses '
+        #             'or subnets separated by comma. (no spaces)'
+        #         ),
+        #     )
```
nemesisdesign, line 519: why are these tests commented out?

nemesisdesign (changes requested):
Thanks for updating the PR @pandafy, during testing, I have noticed something wrong:

- the value is None, the default is Enabled
- I set the value to "Enabled" (not NULL) to force it as enabled
- after saving the value is NULL, goes back to Enabled (default)

Is this intentional or not?
This kind of behavior makes it impossible to override the value when a user knows the default is goint to be changed, say you know you are going to change the default to a Disabled, but you want a specific feature to be enabled for some orgs, the safe way would be to go to those orgs and set the value to "Enabled", but this behavior doesn't make that possible.

See also my comments below.

nemesisdesign (changes requested):
This is what happening now.

Peek 2022-04-14 16-13

I save a new org, the initial value is NULL, I set it to a NOT NULL value, then when I try to get it back to be NULL, it doesn't work as expected, see the GIF.

nemesisdesign (approved):
Ok, it's good, thanks! :+1:

PR #262, opened 2021-05-26
[feature] Added SAML registration method #259

### LINKED ISSUE #259: [feature] Add support for SAML registration method
We need this to add support for [SPID](https://www.spid.gov.it/) and other national identity verification methods based on the SAML protocol.

- [x] It should work in a similar way to social login (we'll probably need a similar view, maybe we can made some of that logic reusablew to avoid duplication), we'll need to pull a third party SAML library which I believe should be optional and installed by default in the test/dev environment
- [x] when enabled, we should register this registration method (like social login in https://github.com/openwisp/openwisp-radius/pull/258)
- [x] we must make sure that with this solution we can implement [SPID](https://www.spid.gov.it/)
- [ ] we may need to add an additional field to `RegisteredUser` to allow storing extra attributes and/or metadata
- [x] we need a way to configure this integration to flag `RegisteredUser.is_verified=True`, because some SAML implementations (like SPID) are valid for identity verification

For testing we may use mocks of identity provider services, a quick look on google yields some results, we should investigate more into this or look into how SAML libraries for django test this.

I would be nice to be able to configure the registration method label, so that instead of showing "Single Sign-On (SAML)" we could configure it to be something different (eg: "SPID").

pandafy:
Closes #259


requirements.txt
```diff
@@ -14,3 +14,4 @@ jsonfield~=3.1.0
 django-private-storage~=2.2
 celery~=4.4.0
 django-ipware~=3.0.0
+djangosaml2~=1.2.0
```
nemesisdesign, line 17: Let's make this dependency optional (eg: https://github.com/openwisp/openwisp-utils/blob/8331e52d32eea77056f9c600c1ff770dafbc8a19/setup.py#L48-L61). I'm ok with getting rid of `requirements.txt` and move the entire definition to `setup.py`, which is simpler (the method used now is just something that was copied over and over from the past).

tests/openwisp2/settings.py
```diff
@@ -48,9 +48,16 @@
...
     'openwisp2.integrations',
+    'djangosaml2',
 ]
 
-LOGIN_REDIRECT_URL = 'admin:index'
+# Bug in djangosaml2: https://git.io/JGUb9
+LOGIN_REDIRECT_URL = '/admin/index'
+
+AUTHENTICATION_BACKENDS = (
+    'django.contrib.auth.backends.ModelBackend',
```
nemesisdesign, line 58: let's take advantage to switch to `openwisp_users.backends.UsersAuthenticationBackend` which will be the new default, can you make sure to also document this please?

tests/openwisp2/settings.py
```diff
@@ -48,9 +48,16 @@
     'drf_yasg',
     'django_extensions',
     'openwisp2.integrations',
+    'djangosaml2',
```
nemesisdesign, line 51: have you tested it without the djangosaml2 library installed in the virtual env and without this line? Does the app keep running?

openwisp_radius/saml/views.py
```diff
@@ -0,0 +1,74 @@
...
+        # add user to organization
+        if not is_member:
+            orgUser = OrganizationUser(organization=org, user=user)
+            orgUser.full_clean()
+            orgUser.save()
+        try:
+            user.registered_user
+        except ObjectDoesNotExist:
+            registered_user = RegisteredUser(
+                user=user, method='social_login', is_verified=False
```
nemesisdesign, line 51: `social_login` > `saml`, you need to register this new method but only if saml is enabled.

nemesisdesign (changes requested):
Looks on the right track!

Seem to be missing:

- docs
- sample app tests

requirements.txt
pandafy: `requirements.txt` makes specifying dependencies from GitHub easier or cleaner in my opinion. I would like it stay.

nemesisdesign: What do you mean specifying dependencies from Github? Editing via the web interface?
It doesn't sound like a good reason to keep some unnecessary code around to me, moreover, the only way to specify this as optional dependency compatible with `setup.py` is to use `extra_requires`, let's simplify please. I hate this `requirements.txt` parsing code repeated in every module, it's really dumb and it was my mistake to put it there in the first place but it's never too late to fix it.

tests/openwisp2/settings.py
pandafy: Yes! I added djangosaml2 in requirements-test.txt. Without it, tests were failing.

requirements.txt
pandafy: Ok, got it.

docs/source/developer/how_to_extend.rst
```diff
@@ -96,6 +103,9 @@ Install (and add to the requirement of your project) openwisp-radius::
 
     pip install openwisp-radius
 
+.. note::
+    Use ``pip install openwisp-radius[saml]`` if you intend to use
+    `Single Sign-On (SAML) <../user/saml_login.html>`_. feature.
```
nemesisdesign, line 108: punctuation on this line looks wrong, I think it should be:

to use the Single Sign-On (..) feature.

docs/source/developer/how_to_extend.rst
```diff
@@ -388,8 +397,9 @@ The root ``url.py`` file should have the following paths (please read the commen
 
     from openwisp_radius.urls import get_urls
     # Only imported when views are extended.
-    # from .sample_radius.api.views import views as api_views
-    # from .sample_radius.social.views import views as social_views
```
nemesisdesign, line 392: thanks for fixing this :+1:

docs/source/developer/setup.rst
```diff
@@ -120,19 +120,33 @@ These modules are optional, add them only if you need the
         'allauth.socialaccount.providers.google',
     ]
 
+This module is optional, add this only if you need the
+`Single Sign-On (SAML) <../user/saml_login.html>`_ feature:
+
+.. code-block:: python
+
+    INSTALLED_APPS += [
+        # sAML login
```
nemesisdesign, line 129: sAML > SAML

docs/source/user/saml_login.rst
```diff
@@ -0,0 +1,98 @@
...
+
+    pip install openwisp-radius[saml]
+
+Ensure your ``settings.py`` looks like the following:
+
+.. code-block:: python
+
+    INSTALLED_APPS = [
+        # ... other apps ..
+        # apps needed for SA<L> login
```
nemesisdesign, line 39: `SA<L>`?

docs/source/user/saml_login.rst
```diff
@@ -0,0 +1,98 @@
...
+After successfully configuring SAML settings for your Identity Provider,
+you will need an HTML button similar to the one in the following example.
+
+This example needs the slug of the organization to assign the new user to
+the right organization:
+
+.. code-block:: html
+
+    <a href="https://openwisp2.mywifiproject.com//radius/saml2/login/?RelayState=https://captivepage.mywifiproject.com%3Forg%3Ddefault
+       class="button">Log in with Facebook
```
nemesisdesign, line 89: `Log in with Facebook`?

openwisp_radius/apps.py
```diff
@@ -46,6 +46,10 @@ def ready(self, *args, **kwargs):
 
         if app_settings.SOCIAL_LOGIN_ENABLED:
             register_registration_method('social_login', _('Social login'))
+        if app_settings.SAML_LOGIN_ENABLED:
+            register_registration_method(
+                'saml', _(app_settings.SAML_REGISTRATION_METHOD_LABEL)
```
nemesisdesign, line 51: gettext won't find this, you have to call `_()` at the line in which the string is defined

openwisp_radius/settings.py
```diff
@@ -40,6 +40,11 @@ def get_settings_value(option, default):
 BATCH_MAIL_SENDER = get_settings_value('BATCH_MAIL_SENDER', settings.DEFAULT_FROM_EMAIL)
 API_AUTHORIZE_REJECT = get_settings_value('API_AUTHORIZE_REJECT', False)
 SOCIAL_LOGIN_ENABLED = 'allauth.socialaccount' in settings.INSTALLED_APPS
+SAML_LOGIN_ENABLED = 'djangosaml2' in settings.INSTALLED_APPS
+SAML_REGISTRATION_METHOD_LABEL = get_settings_value(
+    'SAML_REGISTRATION_METHOD_LABEL', 'Single Sign-On (SAML)'
```
nemesisdesign, line 45: here:
```
'SAML_REGISTRATION_METHOD_LABEL', _('Single Sign-On (SAML)')
```

requirements-test.txt
```diff
@@ -9,3 +9,4 @@ freezegun
 django-extensions
 openwisp-utils[qa]
 pylinkvalidator
+djangosaml2~=1.2.0
```
nemesisdesign, line 12: why don't we avoid this duplicated requirement and modify the github action workflow to do `pip install -e .[saml]` instead of `pip install -e .`?

nemesisdesign (changes requested):
I took some time to review the code and found a number of minor issues, see below.

I am going to test this now and will let you know what I find out.

docs/source/user/saml_login.rst
```diff
@@ -0,0 +1,98 @@
...
+
+After successfully configuring SAML settings for your Identity Provider,
+you will need an HTML button similar to the one in the following example.
+
+This example needs the slug of the organization to assign the new user to
+the right organization:
+
+.. code-block:: html
+
+    <a href="https://openwisp2.mywifiproject.com//radius/saml2/login/?RelayState=https://captivepage.mywifiproject.com%3Forg%3Ddefault
```
nemesisdesign, line 88: there's an extra slash `mywifiproject.com//radius`

docs/source/user/saml_login.rst
```diff
@@ -0,0 +1,98 @@
...
+        'djangosaml2.middleware.SamlSessionMiddleware',
+    ]
+
+Ensure your main ``urls.py`` contains the ``allauth.urls``:
+
+.. code-block:: python
+
+    urlpatterns = [
+        # .. other urls ...
+        path('accounts/', include('allauth.urls')),
```
nemesisdesign, line 67: why is this needed for SAML?

We normally have this:

```
url(r'^accounts/', include('openwisp_users.accounts.urls'))
```

Which includes the subset of URLs allauth.account URLs that are strictly needed for the service to work.
See https://github.com/openwisp/openwisp-users/blob/70e0ffa2522132a2cd3e277a447039e537231f95/openwisp_users/accounts/urls.py.

These URLs are automatically included by openwisp-controller: https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/urls.py#L5-L10.

docs/source/user/saml_login.rst
```diff
@@ -0,0 +1,98 @@
...
+    ]
```
nemesisdesign, line 68: What about `(r'saml2/', include('djangosaml2.urls')),`?

docs/source/user/saml_login.rst
pandafy: I took reference from https://openwisp-radius.readthedocs.io/en/latest/user/social_login.html#setup.
Without accounts URL the SAML feature doesn't work. I have updated both occurrences

docs/source/user/saml_login.rst
pandafy: We don't need it because I wrote the all the routing in https://github.com/openwisp/openwisp-radius/blob/8fdb24b1079bf0e52cb0482d2127b51a0481d0cb/openwisp_radius/saml/urls.py since we extended views

docs/source/developer/setup.rst
```diff
@@ -120,19 +120,33 @@ These modules are optional, add them only if you need the
...
+Also, add ``AUTH_USER_MODEL``, ``AUTHENTICATION_BACKENDS`` and ``SITE_ID`` to
+your ``settings.py``:
 
 .. code-block:: python
 
     AUTH_USER_MODEL = 'openwisp_users.User'
     SITE_ID = 1
+    AUTHENTICATION_BACKENDS = (
+        'openwisp_users.backends.UsersAuthenticationBackend',
+    )
```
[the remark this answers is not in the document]
pandafy: I agree. I have removed it.

docs/source/user/saml_login.rst
nemesisdesign: But without that, while testing, something wasn't working. Are you extra sure this is not needed?

docs/source/user/saml_login.rst
nemesisdesign: Yeah I think that doc page is outdated too.
It should work with the url routes defined in openwisp-users, I tested this, please double check.

docs/source/user/saml_login.rst
pandafy: Yes! I am absolutely sure. `djangosaml2` should be present in `INSTALLED_APPS` and the URLs get added.

nemesisdesign (commented):
I get this error while testing:

```
[ERROR 2021-06-28 22:19:55,375] module: views, process: 22821, thread: 140032059889408
SAMLResponse Error
  File "/opt/openwisp2/env/lib/python3.7/site-packages/djangosaml2/views.py", line 433, in post
    outstanding_queries)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/saml2/client_base.py", line 812, in parse_authn_request_response
    xmlstr, AuthnResponse, "assertion_consumer_service", binding, **kwargs
  File "/opt/openwisp2/env/lib/python3.7/site-packages/saml2/entity.py", line 1512, in _parse_response
    response.verify(keys)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/saml2/response.py", line 1047, in verify
    res = self._verify()
  File "/opt/openwisp2/env/lib/python3.7/site-packages/saml2/response.py", line 419, in _verify
    valid = self.issue_instant_ok() and self.status_ok()
  File "/opt/openwisp2/env/lib/python3.7/site-packages/saml2/response.py", line 389, in issue_instant_ok
    self.timeslack).timetuple()
  File "/opt/openwisp2/env/lib/python3.7/site-packages/saml2/time_util.py", line 229, in shift_time
    return dtime + timedelta(seconds=shift)
TypeError: unsupported type for timedelta seconds component: str
```

Any idea why this is happening?

docs/source/developer/setup.rst
```diff
@@ -350,6 +354,6 @@ If you encounter any issue during installation, run:
 
 .. code-block:: shell
 
-    pip install -r requirements.txt -r requirements-test.txt
+    pip install . -r requirements-test.txt
```
nemesisdesign, line 357: shouldn't this be:
```
pip install -e . -r requirements-test.txt
```

Does it work also without the `-e`?

nemesisdesign (commented):
There's a minor QA check issue.

Let's fix it so we can get this merged soon!

nemesisdesign (changes requested):
Please document `SAML_REGISTRATION_METHOD_LABEL`.

openwisp_radius/settings.py
```diff
@@ -40,6 +40,11 @@ def get_settings_value(option, default):
 BATCH_MAIL_SENDER = get_settings_value('BATCH_MAIL_SENDER', settings.DEFAULT_FROM_EMAIL)
 API_AUTHORIZE_REJECT = get_settings_value('API_AUTHORIZE_REJECT', False)
 SOCIAL_LOGIN_ENABLED = 'allauth.socialaccount' in settings.INSTALLED_APPS
+SAML_LOGIN_ENABLED = 'djangosaml2' in settings.INSTALLED_APPS
+SAML_REGISTRATION_METHOD_LABEL = get_settings_value(
+    'SAML_REGISTRATION_METHOD_LABEL', _('Single Sign-On (SAML)')
+)
+SAML_IS_VERIFIED = get_settings_value('SAML_IS_VERIFIED', False)
```
nemesisdesign, line 47: can you please implement (and document) this to be overridable from the `OrganizationRadiusSettings` please (not urgent)?

nemesisdesign: Ignore this comment, since saml is optional, probably this is overkill now.

openwisp_radius/templates/djangosaml2/auth_error.html
```diff
@@ -0,0 +1,24 @@
+{% comment %}
+This template is taken from djangosaml2 with sole purpose of
+updating name space of the URL identifier.
+{% endcomment %}
+
```
nemesisdesign, line 5: It should not be hard to extend account/base, see here: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/accounts/templates/account/login.html

openwisp_radius/templates/djangosaml2/auth_error.html
```diff
@@ -0,0 +1,24 @@
...
+</head>
+
+<body>
+    <h1>Authorization error</h1>
+    <p>You are already logged in and you are trying to go to the login page again.</p>
+
+    <p>You may have been redirected here when trying to access some content
+        that required extra privileges that you do not have.</p>
+
+    <p>Please <a href="{% url 'radius:saml2_logout' %}">logout</a> and login as a different user</p>
```
nemesisdesign, line 21: Let's flag strings and paragraphs as translatable using `trans` and `blocktrans`

openwisp_radius/templates/djangosaml2/wayf.html
```diff
@@ -0,0 +1,26 @@
...
+    <ul>
+        {% for url, name in available_idps %}
+        <li><a
+                href="{% url 'radius:saml2_login' %}?idp={{ url }}{% if came_from %}&next={{ came_from }}{% endif %}">{{ name|default_if_none:url }}</a>
+        </li>
+        {% endfor %}
+    </ul>
+</body>
+
+</html>
```
nemesisdesign, line 26: same here

docs/source/developer/setup.rst
nemesisdesign: @pandafy can you reply to this please? :point_up:

openwisp_radius/tests/test_saml/utils.py
```diff
@@ -0,0 +1,30 @@
+import base64
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+from djangosaml2.cache import OutstandingQueriesCache
+
+
+class TestSamlUtilities:
```
nemesisdesign, line 8: for consistency with the rest of the openwisp codebase we should call this something like `TestSamlMixins` or similar (see other test mixins used around the codebase)

nemesisdesign (changes requested):
Great progress @pandafy :+1: 

It's working well and I will be doing more testing next week in a production env with a production SPID setting.

See my comments below for more request for improvement. Mostly minor things!

docs/source/developer/setup.rst
pandafy: Earlier the instructions were to install the package using `python setup.py develop`. I have updated that to `pip install -e .[saml]. I have also updated this part too.

openwisp_radius/templates/djangosaml2/auth_error.html
```diff
@@ -1,24 +1,14 @@
...
-updating name space of the URL identifier.
-{% endcomment %}
+{% extends "account/base.html" %}
+{% load i18n %}
 
-<!DOCTYPE html
-    PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
+{% block content %}
+    <h1> {% blocktrans %} Authorization error {% endblocktrans %} </h1>
```
nemesisdesign, line 5: you should be able to use just `translate` or `trans` when it's just one line.

Can you please remove the shitespace between the HTML tags and the words please?

Eg:

`<h1>{% trans 'Word' %}</h1>`

Applies to other occurrencets as well.

nemesisdesign (changes requested):
One small change requested.

nemesisdesign (changes requested):
There seems to be a regression:

```
Internal Server Error: /radius/saml2/acs/
  File "/opt/openwisp2/env/lib/python3.7/site-packages/openwisp_radius/saml/views.py", line 37, in get_org_slug_from_relay_state
    org_slug = parse_qs(parsed_url.query)['org'][0]
KeyError: 'org'

During handling of the above exception, another exception occurred:

  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/djangosaml2/views.py", line 522, in post
    self.post_login_hook(request, user, session_info)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/openwisp_radius/saml/views.py", line 54, in post_login_hook
    org = self.get_organization_from_relay_state()
  File "/opt/openwisp2/env/lib/python3.7/site-packages/openwisp_radius/saml/views.py", line 44, in get_organization_from_relay_state
    org_slug = self.get_org_slug_from_relay_state()
  File "/opt/openwisp2/env/lib/python3.7/site-packages/openwisp_radius/saml/views.py", line 39, in get_org_slug_from_relay_state
    raise ImproperlyConfigured('Organization slug not provided')
django.core.exceptions.ImproperlyConfigured: Organization slug not provided
```

nemesisdesign (commented):
BTW, do you know where this template comes from?
Screenshot from 2021-07-08 16-25-54
Maybe we can just show a generic loader here (something like the l[oader we use in openwisp-wifi-login-pages, just CSS](https://github.com/openwisp/openwisp-wifi-login-pages/blob/b6b28232a48007cb3c9b960adb17b7918be63d9f/public/index.html#L16-L44)).

There's also a few previous comments for minor things that remain to be addressed.

BTW, I am going to change the login redirect parameters a bit:

- we don't need to include the radius_token, frontend apps will have to validate the token received anyway so they'll have to call the validate token endpoint, which if succesful generates a new radius token and returns it in the response
- we have to include something like `saml=true` to let the frontend knows the user logged in via SAML, since the user can log in in multiple ways, a user registered with mobile phone can log in also with SAML, but the registration method won't change, so the user logs in with SAML but registration method says "mobile_phone", the frontend app then needs to know if the user is logged in via SAML to perform the logout properly.

I am testing these changes and if succesful I'll push here and then create an issue to do something similar for social login.

nemesisdesign (commented):
It's looking good and working in test/staging, waiting to test this in prod.

openwisp_radius/saml/views.py
```diff
@@ -0,0 +1,109 @@
...
+        try:
+            parsed_url = urlparse(
+                self.request.POST.get(
+                    'RelayState', self.request.GET.get('RelayState', None),
+                )
+            )
+
+            org_slug = parse_qs(parsed_url.query)['org'][0]
+        except (KeyError, IndexError):
+            raise ImproperlyConfigured('Organization slug not provided')
```
nemesisdesign, line 39: let's not use `ImproperlyConfigure`, it's not suited for these kind of exceptions, here it's better to use a plain `ValueError`

openwisp_radius/saml/views.py
```diff
@@ -0,0 +1,109 @@
...
+        sso_kwargs['isPassive'] = service_config.get('isPassive', False)
+        sso_kwargs['attribute_consuming_service_index'] = service_config.get(
+            'attribute_consuming_service_index', '1'
+        )
+
+    def get(self, request, *args, **kwargs):
+        # Check correct organization slug is present in the request
+        try:
+            org_slug = self.get_org_slug_from_relay_state()
+            assert Organization.objects.filter(slug=org_slug).exists()
```
nemesisdesign, line 103:
```suggestion
            assert Organization.objects.filter(slug=org_slug).exists(), 'Organization with the provided slug does not exist'
```

openwisp_radius/saml/views.py
```diff
@@ -0,0 +1,109 @@
...
+        except (ImproperlyConfigured, AssertionError) as error:
+            if isinstance(error, AssertionError):
+                logger.error('Organization with the provided slug does not exist')
```
nemesisdesign, line 106:
```suggestion
            logger.error(error)
```

nemesisdesign: the suggestions above should fix the logging issue

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #275, opened 2021-07-21
[enhancement] Seperate API urlconf support

atb00ker:


nemesisdesign (commented):
Thanks @atb00ker! Can you double check the tests please?

atb00ker:
Yes, still working on tests. Will update when done! 😄

openwisp_radius/tests/test_models.py
```diff
@@ -588,7 +588,7 @@ def setUp(self):
 
     def _download_csv_file_status(self, status_code):
         response = self.client.get(
-            reverse('radius:serve_private_file', args=[self.csvfile],)
+            reverse('radius_privstore:serve_private_file', args=[self.csvfile],)
```
nemesisdesign, line 591: why is this kind of thing needed?
How does this play with the already deployed instances? For example, how does this change play with the work we did in ansible-openwisp2?

atb00ker: So, previously we were using the same namespace for multiple apps.
This is not ideal. Django usually wants separate namespaces for each application. All other openwisp applications have separated namespaces, somehow, radius was done differently and all the apps were thrown inside `radius` namespace.

While this was working so far, it's not working now that I need to change `urlconf` for a url and it can't find urls with previous structure.

So, I've changed to more idiomatic django code and now `urlconf` can read the unique namespaces for different apps.

No change shall be required in ansible-openwisp2.

nemesisdesign (commented):
I will test this asap, I am delayed but I will get here too, give me some more time please, thank you for your patience.

atb00ker:
No problems, I am underload till August end as well! 😄

openwisp_radius/tests/test_models.py
nemesisdesign: The problem Ajay is that these subfolders in openwisp radius are not django apps.
OpenWISP RADIUS is a single app with different features (API, social login, saml login, etc), but it's only one django app, therefore I don't think we should add more namespaces because it would be confusing.
Can you please give pratical examples of how the single namespace is affecting you negatively so we can look for solutions?

At the moment to me it's still not clear why we need to add new namespaces.

openwisp_radius/urls.py
```diff
@@ -13,31 +12,18 @@ def get_urls(api_views=None, social_views=None, saml_views=None):
...
-    return urls
 
 
-app_name = 'radius'
-urlpatterns = get_urls()
+# Cannot used 'app_name' here because when the radius
+# urls are not registered, yet we want to get url reverse
+# using urlconf we still want radius to be our namespace.
+namespace = 'radius'
+urlpatterns = [path('', include((get_urls(), namespace), namespace=namespace))]
```
atb00ker, line 29: So, here is the solution, I've come up with.
I don't use `app_name` instead, I explicitly mention the namespace.
Hence, when radius urls are not registered, I'll still get the `radius:` namespace when using the urls with the help of `urlconf=openwisp.urls`.

I need the radius namespace because some `reverse` requests are trying to look for urls in that namespace which gets registered when the registered with the help of `app_name`!

atb00ker (commented):
LGTM! 👍🏻

nemesisdesign (approved):

PR #280, opened 2021-08-12
[feature] Added management command and celery task to delete unverified users #279

### LINKED ISSUE #279: [feature] Add management command and celery task to delete unverified users
This was a feature that was previously used in OpenWISP1, certain type of unverified users were removed automatically.

We should replicate the same, following something similar to what we did previously, see:
https://github.com/openwisp/openwisp-radius/blob/211c30cd1d249acb73619c2a39a77a2f6324d500/openwisp_radius/tasks.py

The new command shall look for users which have a RegisteredUser relation and RegisteredUser.is_verified=False and have been created for more than X days ago, where days is is a parameter defaulting to 1 day (24 hours).

The command shall accept an `exclude_methods` parameter which can contain a list of verification methods to exclude from the query, so that certain verification methods can be excluded and not deleted anyway.

The command shall be documented as the other ones.

pandafy:
Closes #279

### DIFF
--- a/docs/source/user/management_commands.rst
+++ b/docs/source/user/management_commands.rst
@@ -85,6 +85,26 @@ This command deletes users that have expired (and should have been deactivated b
 
 Note that the default duration is set to 18 months.
 
+
+``delete_unverified_users``
+---------------------------
+
+This command deletes unverified users that have been registered for
+more than specified duration.
+
+.. code-block:: shell
+
+    ./manage.py delete_unverified_users --older-than-days <duration_in_days>
+
+Note that the default duration is set to 1 day.
+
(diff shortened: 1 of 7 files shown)


docs/source/user/management_commands.rst
```diff
@@ -85,6 +85,26 @@ This command deletes users that have expired (and should have been deactivated b
 
 Note that the default duration is set to 18 months.
 
+
```
nemesisdesign, line 88: please remove this extra blank line

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #284, opened 2021-08-17
[feature] [feature] Convert RADIUS accounting CALLED-STATION-ID #276

### LINKED ISSUE #276: [feature] Convert RADIUS accounting CALLED-STATION-ID
The old generation of OpenWISP had a background task running regularly which was able to modify the called station ID of the radius sessions to include the access point mac address (instead of the captive portal mac address), this is useful in set ups which run a centralized captive portal, because in these setups the called-station-id attribute is always the captive portal mac address so the radius session info is kind of incomplete.

This is the old code: https://github.com/openwisp/OpenWISP-User-Management-System/blob/3a6a789637c356213c09e18a22bc4b069567bdd6/app/models/radius_accounting.rb#L229-L267.

~~We'll have to implement something equivalent, I need to think about it.~~

I think we should implement this as a management command (similar to what I described in #279) which is configurable via a django setting.

The django setting should map the organization slug to:

- the OpenVPN management port where to fetch the data.
- a list of possible values that indicate the called station ID has not been converted yet  

We need to look for radius accounting sessions which have not been converted (look if called-station-id is any of the values provided in the setting), if we find any session that has to be converted then we proceed to fetch the data from OpenVPN via the management port.

Once the data is retrieved we shall use the [openvpn-status](https://pypi.org/project/openvpn-status/) library to get the list of clients.

Then we loop over any non converted session, look if the mac address (calling station id) of the user or ip address in the session info matches any OpenVPN client and if it does we find out the mac address of the device so we finally change the called station ID to contain this mac address, converting the mac address to be like 00-00-00-00-00-00 instead of 00:00:00:00:00:00 for backward compatibility reasons.

We should make this optional (and the openvpn-status depency as well).

pandafy:
Added management command and celery task to convert/correct
called station id of RADIUS session.

Closes #276

____________________________

TODO: 
- [x] Docs
- [x]  Error handling

### DIFF
--- a/docs/source/user/management_commands.rst
+++ b/docs/source/user/management_commands.rst
@@ -130,3 +130,17 @@ by default the data is added to the first found organization, eg::
 
 .. warning::
     It is not possible to export user credential data for radiusbatch created using prefix, please manually preserve the PDF files if you want to access the data in the future.
+
+``convert_called_station_id``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This commands updates the ``called_station_id`` of RADIUS Sessions from MAC Address of captive
+portal to MAC Address of the device using information from OpenVPN. This command requires
+installing ``openvpn_status``, which can be installed using following command
+
+.. code-block:: shell
+
+    pip install openwisp-radius[openvpn_status]
+
+It is required configuring `OPENWISP_RADIUS_CALLED_STATION_ID <./settings.html#openwisp-radius-called-station-id>`_
+setting for functioning of this command.
--- a/docs/source/user/management_commands.rst
+++ b/docs/source/user/management_commands.rst
@@ -138,14 +138,14 @@ If an installation uses a centralized captive portal, the value of "Called Stati
 RADIUS Session will show the MAC address of the captive portal instead of access points.
 This command will update the "Called Station ID" to reflect the MAC address of the access points
 using information from OpenVPN. It requires installing ``openvpn_status``,
- which can be installed using the following command
+which can be installed using the following command
 
 .. code-block:: shell
 
     pip install openwisp-radius[openvpn_status]
 
 In order to work, this command requires to be configured via the
-`OPENWISP_RADIUS_CALLED_STATION_ID <./settings.html#openwisp-radius-called-station-id>`_ setting.
+`OPENWISP_RADIUS_CALLED_STATION_IDS <./settings.html#openwisp-radius-called-station-ids>`_ setting.
 
 .. note::
 
--- a/docs/source/user/management_commands.rst
+++ b/docs/source/user/management_commands.rst
@@ -134,13 +134,21 @@ by default the data is added to the first found organization, eg::
 ``convert_called_station_id``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-This commands updates the ``called_station_id`` of RADIUS Sessions from MAC Address of captive
-portal to MAC Address of the device using information from OpenVPN. This command requires
-installing ``openvpn_status``, which can be installed using following command
+If an installation uses a centralized captive portal, the value of "Called Station ID" of
+RADIUS Session will show the MAC address of the captive portal instead of access points.
+This command will update the "Called Station ID" to reflect the MAC address of the access points
+using information from OpenVPN. It requires installing ``openvpn_status``,
+ which can be installed using the following command
 
 .. code-block:: shell
 
     pip install openwisp-radius[openvpn_status]
 
-It is required configuring `OPENWISP_RADIUS_CALLED_STATION_ID <./settings.html#openwisp-radius-called-station-id>`_
-setting for functioning of this command.
+In order to work, this command requires to be configured via the
+`OPENWISP_RADIUS_CALLED_STATION_ID <./settings.html#openwisp-radius-called-station-id>`_ setting.
+
+.. note::
+
+    If you encounter ``ParseError`` for datetime data, you can set the datetime format
+    of the parser using `OPENWISP_RADIUS_OPENVPN_DATETIME_FORMAT <./settings.html#openwisp-radius-openvpn-datetime-format>`_
+    setting.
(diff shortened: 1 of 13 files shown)


docs/source/user/settings.rst
```diff
@@ -200,6 +200,38 @@ This setting can be used to set the maximum size limit for firmware images, eg:
     The numeric value represents the size of files in bytes.
     Setting this to ``None`` will mean there's no max size.
 
+
```
nemesisdesign, line 203: avoid this extra blank line please

docs/source/user/settings.rst
```diff
@@ -200,6 +200,38 @@ This setting can be used to set the maximum size limit for firmware images, eg:
...
+This setting should contain information of OpenVPN Management Interface for fetching
+information of client to update ``called_station_id`` of RADIUS sessions using
+`convert_called_station_id <management_commands.html#convert-called-station-id>`_ command.
+
+The dictionary should contain configuration in following format:
+
+.. code-block:: python
+
+    OPENWISP_RADIUS_CALLED_STATION_IDS = {
+        'organization_slug': {
```
nemesisdesign, line 218: should `organization_slug` contain the actual organization slug?

docs/source/user/settings.rst
```diff
@@ -200,6 +200,38 @@ This setting can be used to set the maximum size limit for firmware images, eg:
...
+            'openvpn_config': [
```
nemesisdesign, line 219: what is `openvpn_config` exactly?

openwisp_radius/management/commands/base/convert_called_station_id.py
```diff
@@ -0,0 +1,111 @@
...
+                else:
+                    radius_session.save()
+
+
+# monkey patching for openvpn_status begins
+def parse_virtual_address(virtual_address):
+    return openvpn_status.utils.parse_vaddr(virtual_address.split('@')[0])
+
+
+openvpn_status.utils.DATETIME_FORMAT_OPENVPN = u'%Y-%m-%d %H:%M:%S'
```
nemesisdesign, line 105: this is quite brutal, because it would mean this command would only work on pfsense. I think we should add a way to specify this via configuration so that if a specific flag is enabled, this is done.

docs/source/user/management_commands.rst
```diff
@@ -111,3 +111,17 @@ by default the data is added to the first found organization, eg::
...
+This commands updates the ``called_station_id`` of RADIUS Sessions from MAC Address of captive
+portal to MAC Address of the device using information from OpenVPN. This command requires
+installing ``openvpn_status``, which can be installed using following command
+
+.. code-block:: shell
+
+    pip install openwisp-radius[openvpn_status]
+
+It is required configuring `OPENWISP_RADIUS_CALLED_STATION_ID <./settings.html#openwisp-radius-called-station-id>`_
+setting for functioning of this command.
```
nemesisdesign, line 127: it sounds awkard, what about something like "In order to work, this command requires to be configured via the X setting"

docs/source/user/management_commands.rst
```diff
@@ -111,3 +111,17 @@ by default the data is added to the first found organization, eg::
 
 .. warning::
     It is not possible to export user credential data for radiusbatch created using prefix, please manually preserve the PDF files if you want to access the data in the future.
+
+``convert_called_station_id``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This commands updates the ``called_station_id`` of RADIUS Sessions from MAC Address of captive
+portal to MAC Address of the device using information from OpenVPN. This command requires
+installing ``openvpn_status``, which can be installed using following command
```
nemesisdesign, line 120: we should explain why it's needed (with centralized captive portals setup the called station ID is always the CP while we want to have the mac address of the AP).

docs/source/user/settings.rst
```diff
@@ -200,6 +200,38 @@ This setting can be used to set the maximum size limit for firmware images, eg:
     The numeric value represents the size of files in bytes.
     Setting this to ``None`` will mean there's no max size.
 
+
+``OPENWISP_RADIUS_CALLED_STATION_ID``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Default**: ``{}``
+
+This setting should contain information of OpenVPN Management Interface for fetching
```
nemesisdesign, line 209: This setting allows to specify the parameters to connect to the different OpenVPN management interfaces available for an organization.

Or something along these lines.

nemesisdesign (commented):

docs/source/user/settings.rst
pandafy: yes, I have update the example to be more explicit

nemesisdesign (changes requested):
There are some conflicts to resolve here.

pandafy:
> PS: do you confirm the code is usable without password? It looks like it but I just wanted to double check with you.

Yes the code works without a password.

docs/source/user/settings.rst
```diff
@@ -200,6 +200,51 @@ This setting can be used to set the maximum size limit for firmware images, eg:
     The numeric value represents the size of files in bytes.
     Setting this to ``None`` will mean there's no max size.
 
+``OPENWISP_RADIUS_CALLED_STATION_ID``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesisdesign, line 204: the setting seems to be called `OPENWISP_RADIUS_CALLED_STATION_IDS` (final s)

docs/source/user/management_commands.rst
```diff
@@ -130,3 +130,25 @@ by default the data is added to the first found organization, eg::
...
+This command will update the "Called Station ID" to reflect the MAC address of the access points
+using information from OpenVPN. It requires installing ``openvpn_status``,
+ which can be installed using the following command
+
+.. code-block:: shell
+
+    pip install openwisp-radius[openvpn_status]
+
+In order to work, this command requires to be configured via the
+`OPENWISP_RADIUS_CALLED_STATION_ID <./settings.html#openwisp-radius-called-station-id>`_ setting.
```
nemesisdesign, line 148: same here (anchor text and target anchor)

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Found 2 minor issues in the README

nemesisdesign (approved):

PR #285, opened 2021-08-20
[change/fix] Update username when phone number is changed #282

### LINKED ISSUE #282: [change/fix] Update username when phone number is changed
In the following view:

https://github.com/openwisp/openwisp-radius/blob/2465d6f0b367aa5fba70eb67b2e7d905ac66ef8f/openwisp_radius/api/views.py#L580-L588

There's a corner case to handle.
When using mobile phone verification, the username is set to be the same as the phone number.
So when the phone number is updated right now the username remains the old phone number.

We should add some logic to handle this corner case, it's not the cleanest solution but we can improve this in the future.
We can look if the old phone number is the same as the username and in that case I would update the username as well.

pandafy:
If user has registered with only phone number, the user will have
only phone number as username. If the user updates the phone_number,
then the username of the user will also get updated.

Closes #282

### DIFF
--- a/openwisp_radius/api/views.py
+++ b/openwisp_radius/api/views.py
@@ -580,6 +580,9 @@ class ValidatePhoneTokenView(DispatchOrgMixin, GenericAPIView):
             user.registered_user.is_verified = True
             user.registered_user.method = 'mobile_phone'
             user.is_active = True
+            # Update username if phone_number is used as username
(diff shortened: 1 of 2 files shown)


nemesisdesign (approved):

nemesisdesign (approved):
Thanks :+1:

PR #293, opened 2021-09-01
[fix] Used openwisp-users auth backend to filter user in FreeradiusAPIAuthentication #292

### LINKED ISSUE #292: [bug] FreeradiusApiAuthentication does not honor authentication backend
Recently we have introduced an [authentication backend in openwisp-users to allow users to log in using their username, email address or phone number](https://github.com/openwisp/openwisp-users/#authentication-backend).

However, our `RadiusToken` authentication here doesn't respect that, it only looks for `User.username`.

https://github.com/openwisp/openwisp-radius/blob/5947e2022254f2b084f91d677918db1a1054a950/openwisp_radius/api/freeradius_views.py#L100-L119

Here's how to replicate it:

- register a user through the API
- ensure the username is different than the email and phone number
- try to use the freeradius authorize API setting the phone number or email as username

**Expected result:**

Successful result.

**Actual result:**

``` 
Forbidden 403
{"detail":"Radius token does not exist. Obtain a new radius token or provide the organization UUID and API token."}
```

### More details

The `AuthorizeView` respects `openwisp_users.backends.UsersAuthenticationBackend`:

https://github.com/openwisp/openwisp-radius/blob/5947e2022254f2b084f91d677918db1a1054a950/openwisp_radius/api/freeradius_views.py#L207-L210

We should simply do something similar in `_radius_token_authenticate`, instead of calling `RadiusToken.objects.get(user__username=username)` directly.

We will need a failing test for this.

codesankalp:
Used authentication backend of openwisp-users to filter users with the username,
email and mobile phone.
Also, added a failing test for this.

Fixes #292

### DIFF
--- a/openwisp_radius/api/freeradius_views.py
+++ b/openwisp_radius/api/freeradius_views.py
@@ -106,7 +106,9 @@ class FreeradiusApiAuthentication(BaseAuthentication):
             return self._check_client_ip_and_return(request, cached_orgid)
         else:
             try:
-                radtoken = RadiusToken.objects.get(user__username=username)
+                radtoken = RadiusToken.objects.get(
+                    user=auth_backend.get_users(username).first()
+                )
             except RadiusToken.DoesNotExist:
                 if username:
                     message = _(
--- a/openwisp_radius/tests/test_api/test_freeradius_api.py
+++ b/openwisp_radius/tests/test_api/test_freeradius_api.py
@@ -132,16 +132,38 @@ class TestFreeradiusApi(AcctMixin, ApiTokenMixin, BaseTestCase):
         self.assertEqual(r.status_code, 200)
         self.assertEqual(r.data, {'control:Auth-Type': 'Accept'})
 
+    def _test_authorize_without_auth_helper(self, username, password):
+        r = self._authorize_user(username=username, password=password)
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(r.data, {'control:Auth-Type': 'Accept'})
+
     def test_authorize_with_user_auth(self):
+        user = self._create_user(
+            username='tester2',
+            email='tester2@gmail.com',
+            phone_number='+237675679232',
+            password='tester',
+        )
+
+        self._create_org_user(organization=self._get_org(), user=user)
+
+        with self.subTest('Test authorize with username'):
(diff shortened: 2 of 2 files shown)


openwisp_radius/api/freeradius_views.py
```diff
@@ -106,7 +106,9 @@ def _radius_token_authenticate(self, request):
             return self._check_client_ip_and_return(request, cached_orgid)
         else:
             try:
-                radtoken = RadiusToken.objects.get(user__username=username)
+                radtoken = RadiusToken.objects.get(
+                    user=auth_backend.get_users(username).first()
```
nemesisdesign, line 110: what if `first()` returns `None`, could you add a test for this case and ensure it's handled well?

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -132,16 +132,38 @@ def _test_authorize_with_user_auth_helper(self, username, password):
...
+        self._create_org_user(organization=self._get_org(), user=user)
+
+        with self.subTest('Test authorize with username'):
+            self._test_authorize_with_user_auth_helper(user.username, 'tester')
+
+        with self.subTest('Test authorize with email'):
+            self._test_authorize_with_user_auth_helper(user.email, 'tester')
+
+        with self.subTest('Test authorize with phone_number'):
+            self._test_authorize_with_user_auth_helper(user.phone_number, 'tester')
```
nemesisdesign, line 157: here you could add:

```python
with self.subTest('Test authorization failure'):
            self._test_authorize_with_user_auth_helper('thisuserdoesnotexist', 'tester')
```

nemesisdesign (changes requested):

openwisp_radius/tests/test_api/test_freeradius_api.py
codesankalp: Ok, will add it.

openwisp_radius/api/freeradius_views.py
codesankalp: It will throw does not exist exception if the user is unknown so it was already handled.

nemesisdesign (approved):
Great, it works, thanks! :+1:

PR #296, opened 2021-09-10
[change] Improved REST API to change password #289

### LINKED ISSUE #289: [change] Improve REST API to change password
The current API endpoint to change the password does not ask for the old password.
If an attacker is able to steal the rest token of a user in some way, then they can authenticate as them and change the password.

An [API endpoint to change the password is being added to OpenWISP Users](https://github.com/openwisp/openwisp-users/pull/264), we may be able to inherit that view here.

codesankalp:
Inherited PasswordChangeView of openwisp-users to add support for current-password in password change
view.

Closes #289

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -248,6 +248,7 @@ class GroupSerializer(serializers.ModelSerializer):
     class Meta:
         model = Group
         fields = '__all__'
+        ref_name = 'radius_user_group_serializer'
 
 
 class UserSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


codesankalp:
Failing build due to:
```
drf_yasg.errors.SwaggerGenerationError: Schema for
<class 'openwisp_radius.api.serializers.GroupSerializer'>
would override distinct serializer 
<class 'openwisp_users.api.serializers.GroupSerializer'> 
because they implicitly share the same ref_name; explicitly set the 
ref_name attribute on both serializers' Meta classes
```

openwisp_radius/api/serializers.py
```diff
@@ -231,14 +231,14 @@ class Meta:
         read_only_fields = ('organization',)
 
 
-class GroupSerializer(serializers.ModelSerializer):
+class UserGroupSerializer(serializers.ModelSerializer):
```
codesankalp, line 234: I tried to change this but another error is showing again in the CI build.

nemesisdesign (approved):
Thanks @codesankalp :+1:

PR #300, opened 2021-09-20
[change] Allowed users to sign up to multiple organizations #297

### LINKED ISSUE #297: [feature/change] Allow users to sign up to multiple organizations
If a user already has an account with an organization and tries to register to a different organization, the registration will fail, but no way will be provided to the user to acquire membership of the service of the organization they're interested in.

A possible solution to this issue would be the following.

### Override validation to communicate a conflict

We can override the validation process in `RegisterSerializer` so that the [run_validation method](https://github.com/encode/django-rest-framework/blob/3.12.4/rest_framework/serializers.py#L409-L427) would execute a special validation method which we add.

The special method can handle the following cases:

1. user with username/email/phone number passed does not exist, do nothing
2. user with username/email/phone already exists for this organization, do nothing, the current validation handles this
3. user with  username/email/phone already exists for another organization but not the one with which the user is registering to

In the third case, we should prepare a special response which can be easily recognized by client applications so user interfaces
can provide an appropriate solution for the user.

I think we should reply with a 409 HTTP status (conflict) and the following:

```json
{
  "detail": "A user like the one being registered already exists.",
  "organizations": [
    {
      "slug": "...",
      "name": "..."
    }
  ]
}
```

The ``organizations`` member shall include the organizations which the user is member of, so these can be shown in the UI to help the user understand what is going on, since they may not know that these services are sharing the database.

### Change ObtainAuthTokenView to avoid membership check

In `ObtainAuthTokenView`, we have to override `validate_membership` so that instead of rejecting users if they're not members of the current org, users will be automatically signed up to the new org.

### Conclusion

With the changes proposed above, frontend apps can show the right error message during sign up, then propose users to sign in with their existing credentials and the rest should work.

pandafy:
Closes #297

### DIFF
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -411,62 +411,24 @@ Registering to Multiple Organizations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 An **HTTP 409** response will be returned if an existing user tries to register
-on a URL of different organization. The response will contain a list of organizations
-with which the user has already registered to the system. Following is an example
-response:
+on a URL of a different organization (because the account already exists).
+The response will contain a list of organizations with which the user has already
+registered to the system which may be shown to the user in the UI. E.g.:
 
-.. code-block:: shell
+.. code-block:: json
 
-    $ curl -X POST \
-    http://127.0.0.1:8000/api/v1/radius/organization/openwisp/account/ \
-    -H 'content-type: application/json' \
-    -d '{
-        "username": "openwisp_user",
-        "email": "user@openwisp.io",
-        "password1": "<your-super-secret-password>",
-        "password2": "<your-super-secret-password>"
-    }'
 
-    HTTP/1.1 409 Conflict
-    Date: Tue, 21 Sep 2021 15:13:38 GMT
-    Server: WSGIServer/0.2 CPython/3.8.10
-    Content-Type: application/json
-    Vary: Accept, Origin
-    Allow: POST, OPTIONS
-    X-Frame-Options: DENY
-    Content-Length: 120
-    X-Content-Type-Options: nosniff
-    Referrer-Policy: same-origin
-
-    {"details":"A user like the one being registered already exists.","organizations":[{"slug":"default","name":"default"}]}
+    {
+        "details": "A user like the one being registered already exists.",
+        "organizations":[
+            {"slug":"default","name":"default"}
+        ]
+    }
 
 The existing user can register with a new organization using the
 `login <#login-obtain-user-auth-token>`_ endpoint. The user will also get
-membership of the new organization.
-
-.. code-block:: shell
-
-    $ curl -X POST -i \
-    >   http://127.0.0.1:8000/api/v1/radius/organization/openwisp/account/token/ \
-    >   -H 'cache-control: no-cache' \
-    >   -H 'content-type: application/json' \
-    >   -d '{
-    > "username": "openwisp_user",
-    > "password": "password"
-    > }'
-
-    HTTP/1.1 200 OK
-    Date: Tue, 21 Sep 2021 15:25:54 GMT
-    Server: WSGIServer/0.2 CPython/3.8.10
-    Content-Type: application/json
-    Allow: POST, OPTIONS
-    X-Frame-Options: DENY
-    Content-Length: 294
-    Vary: Origin
-    X-Content-Type-Options: nosniff
-    Referrer-Policy: same-origin
-
-    {"username":"openwisp","email":"user@openwisp.io","phone_number":null,"first_name":"","last_name":"","birth_date":null,"location":"","is_active":true,"is_verified":false,"method":"","radius_user_token":"XxbGRU3KV7yQOgLZucj8MVvqX9zdzGb9rAM9LjdO","key":"2f26d26fca112d6081e4c142e66480bf9e6729f7"}
+membership of the new organization only if the organization has
+`user registration enabled <settings.html#openwisp-radius-registration-api-enabled>`_.
 
 Reset password
 --------------
@@ -571,10 +533,10 @@ anyway but using the HTTP status code 401, this way consumers can recognize
 these users and trigger the appropriate response needed (eg: reject them
 or initiate account verification).
 
-If an existing user account tries to log in with the URL for a different organization,
-then the user is also added as a member of the new organization.
-Please refer to the `"Registering to Multiple Organizations" <#registering-to-multiple-organizations>`_
-section of this documentation for details.
+If an existing user account tries to authenticate to an organization of which
+they're not member of, then they would be automatically added as members
+(if registration is enabled for that org). Please refer to
+`"Registering to Multiple Organizations" <#registering-to-multiple-organizations>`_.
 
 Parameters:
 
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -407,6 +407,67 @@ simple, but can be `enabled through configuration <./settings.html#openwisp-radi
 if identity verification is disabled for a particular org, an empty string
 will be acceptable.
 
+Registering to Multiple Organizations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An **HTTP 409** response will be returned if an existing user tries to register
+on a URL of different organization. The response will contain a list of organizations
+with which the user has already registered to the system. Following is an example
+response:
+
+.. code-block:: shell
+
+    $ curl -X POST \
+    http://127.0.0.1:8000/api/v1/radius/organization/openwisp/account/ \
+    -H 'content-type: application/json' \
+    -d '{
+        "username": "openwisp_user",
+        "email": "user@openwisp.io",
+        "password1": "<your-super-secret-password>",
+        "password2": "<your-super-secret-password>"
+    }'
+
+    HTTP/1.1 409 Conflict
+    Date: Tue, 21 Sep 2021 15:13:38 GMT
+    Server: WSGIServer/0.2 CPython/3.8.10
+    Content-Type: application/json
+    Vary: Accept, Origin
+    Allow: POST, OPTIONS
+    X-Frame-Options: DENY
+    Content-Length: 120
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: same-origin
+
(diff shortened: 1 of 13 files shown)


openwisp_radius/tests/test_api/test_rest_token.py
```diff
@@ -90,6 +90,20 @@ def test_user_auth_token_400_organization(self):
         self.assertEqual(r.status_code, 400)
         self.assertIn('is not member', r.json()['non_field_errors'][0])
```
pandafy, line 91: A user will be able to register to a new organization only if it is already a member of an existing organization.

openwisp_radius/api/views.py
```diff
@@ -294,16 +295,23 @@ def get_user(self, serializer, *args, **kwargs):
 
     def validate_membership(self, user):
         if not (user.is_superuser or user.is_member(self.organization)):
-            if OrganizationUser.objects.filter(user=user).exists():
+            try:
+                org_registration_enabled = (
+                    self.organization.radius_settings.get_registration_enabled()
+                )
+            except OrganizationRadiusSettings.DoesNotExist:
+                org_registration_enabled = app_settings.REGISTRATION_API_ENABLED
```
nemesisdesign, line 303: let's avoid this duplication between here and the changes in openwisp_radius/api/permissions.py, see how something similar is done with a function: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/utils.py#L43-L52.

openwisp_radius/api/views.py
```diff
@@ -294,16 +295,23 @@ def get_user(self, serializer, *args, **kwargs):
...
                 OrganizationUser.objects.create(
                     user=user, organization=self.organization
                 )
             else:
                 message = _(
-                    'The user {username} is not member of any organization.'
-                ).format(username=user.username)
-                logger.warning(message)
-                raise serializers.ValidationError({'non_field_errors': [message]})
+                    'Registration is disabled for {organization} organization'.format(
```
nemesisdesign, line 310: Please change this message to `{organization} does not allow self registration of new accounts.`

openwisp_radius/tests/test_api/test_rest_token.py
```diff
@@ -81,15 +82,6 @@ def test_user_auth_token_400_credentials(self):
...
 
-    @capture_any_output()
-    def test_user_auth_token_400_organization(self):
-        url = self._get_url()
-        self._get_org_user()
-        OrganizationUser.objects.all().delete()
-        r = self.client.post(url, {'username': 'tester', 'password': 'tester'})
-        self.assertEqual(r.status_code, 400)
-        self.assertIn('is not member', r.json()['non_field_errors'][0])
-
```
nemesisdesign, line 92: I'd keep the test for the case in which the user has no org, just in case, but you can add this as a subtest in the new test below

nemesisdesign (changes requested):

nemesisdesign (changes requested):
Looks good! I am doing some manual testing now.

See my comments below.

Please update the API docs to briefly mention this behavior:

- https://openwisp-radius.readthedocs.io/en/latest/user/api.html#user-registration (explain 403 and mention the error message, point out users can sign up to the same org by using obtain auth token endpoint)
- https://openwisp-radius.readthedocs.io/en/latest/user/api.html#login-obtain-user-auth-token (explain what happens when user tries to log in to an org they're not member yet of)

openwisp_radius/api/serializers.py
```diff
@@ -413,6 +416,62 @@ def validate_optional_fields(self, field_name, field_value, org):
...
+            'username' in error_dict
+            and 'A user with that username already exists.' in error_dict['username']
+        ):
+            user_lookup |= Q(user__username=data['username'])
+        if (
+            'email' in error_dict
+            and 'A user is already registered with this e-mail address.'
+            in error_dict['email']
+        ):
+            user_lookup |= Q(user__email=data['email'])
```
nemesisdesign, line 435: if present, we should check also the phone number

nemesisdesign (changes requested):
I tried with wifi-login-pages but I was not able to trigger the registration 409 conflict, please test there as well.

openwisp_radius/api/serializers.py
```diff
@@ -413,6 +416,62 @@ def validate_optional_fields(self, field_name, field_value, org):
...
 
+    def validate_cross_org_registration(self, error, data):
+        error_dict = error.detail
+        if 'username' not in error_dict or 'email' not in error_dict:
+            raise error
+
+        user_lookup = Q()
+        if (
+            'username' in error_dict
+            and 'A user with that username already exists.' in error_dict['username']
```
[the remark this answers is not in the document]
pandafy: Removing checks for these error strings will make below checks execute for errors unrelated to cross organization registration. E.g. registering using phone_number when it is not allowed.

docs/source/user/api.rst
```diff
@@ -407,6 +407,67 @@ simple, but can be `enabled through configuration <./settings.html#openwisp-radi
...
+    Vary: Accept, Origin
+    Allow: POST, OPTIONS
+    X-Frame-Options: DENY
+    Content-Length: 120
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: same-origin
+
+    {"details":"A user like the one being registered already exists.","organizations":[{"slug":"default","name":"default"}]}
+
+The existing user can register with a new organization using the
```
nemesisdesign, line 443: I'd clear this behavior could be disabled if needed, eg: `with a new organization (provided registration is enabled for that organization)`

docs/source/user/api.rst
```diff
@@ -407,6 +407,67 @@ simple, but can be `enabled through configuration <./settings.html#openwisp-radi
...
+    Date: Tue, 21 Sep 2021 15:13:38 GMT
+    Server: WSGIServer/0.2 CPython/3.8.10
+    Content-Type: application/json
+    Vary: Accept, Origin
+    Allow: POST, OPTIONS
+    X-Frame-Options: DENY
+    Content-Length: 120
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: same-origin
+
```
nemesisdesign, line 440: I would include only the response JSON to keep it short and maintainable

docs/source/user/api.rst
```diff
@@ -407,6 +407,67 @@ simple, but can be `enabled through configuration <./settings.html#openwisp-radi
 if identity verification is disabled for a particular org, an empty string
 will be acceptable.
 
+Registering to Multiple Organizations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An **HTTP 409** response will be returned if an existing user tries to register
+on a URL of different organization. The response will contain a list of organizations
+with which the user has already registered to the system. Following is an example
+response:
```
nemesisdesign, line 416: `Following is an example response:` > `EG`:

Let's keep it short please.

docs/source/user/api.rst
```diff
@@ -407,6 +407,67 @@ simple, but can be `enabled through configuration <./settings.html#openwisp-radi
 if identity verification is disabled for a particular org, an empty string
 will be acceptable.
 
+Registering to Multiple Organizations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An **HTTP 409** response will be returned if an existing user tries to register
+on a URL of different organization. The response will contain a list of organizations
```
nemesisdesign, line 414: of **a** different organization **(because the account already exists)**.
The response will contain a list of organizations with which the user has already registered to the system which may be shown to the user in the UI.

docs/source/user/api.rst
```diff
@@ -407,6 +407,67 @@ simple, but can be `enabled through configuration <./settings.html#openwisp-radi
...
+    Server: WSGIServer/0.2 CPython/3.8.10
+    Content-Type: application/json
+    Allow: POST, OPTIONS
+    X-Frame-Options: DENY
+    Content-Length: 294
+    Vary: Origin
+    X-Content-Type-Options: nosniff
+    Referrer-Policy: same-origin
+
+    {"username":"openwisp","email":"user@openwisp.io","phone_number":null,"first_name":"","last_name":"","birth_date":null,"location":"","is_active":true,"is_verified":false,"method":"","radius_user_token":"XxbGRU3KV7yQOgLZucj8MVvqX9zdzGb9rAM9LjdO","key":"2f26d26fca112d6081e4c142e66480bf9e6729f7"}
```
nemesisdesign, line 469: I would avoid this example because is redundant and we want to keep this document easily maintainable because everyone (including me) keeps on forgetting to update it, so we shall add only the minimum required for developers to understand they key concepts of the API and nothing else, this makes it easier to maintain.

docs/source/user/api.rst
```diff
@@ -510,6 +571,11 @@ anyway but using the HTTP status code 401, this way consumers can recognize
 these users and trigger the appropriate response needed (eg: reject them
 or initiate account verification).
 
+If an existing user account tries to log in with the URL for a different organization,
+then the user is also added as a member of the new organization.
+Please refer to the `"Registering to Multiple Organizations" <#registering-to-multiple-organizations>`_
+section of this documentation for details.
```
nemesisdesign, line 577: I would rephrase this: different organization doesn't make sense because different to what?

I would talk about trying to authenticate to an organization of which they're not member of and in that case they would be automatically added as members (if registration is enabled for that org).

```
Please refer to the `"Registering to Multiple Organizations" <#registering-to-multiple-organizations>`_
section of this documentation for details.
```

I suggest shorten it:
```
Please refer to `"Registering to Multiple Organizations" <#registering-to-multiple-organizations>`_.```

nemesisdesign (changes requested):
Testing went well :+1:.

See my comments below to improve the docs.

docs/source/user/api.rst
pandafy: We haven't added a dedicated setting for this. Did you mean `provided registration is **disabled** for that organization)` ?

nemesisdesign: I mean this (see the last field):

Screenshot from 2021-09-22 10-03-13

nemesisdesign (changes requested):
We found a couple of critical bugs, see below.

nemesisdesign (approved):
Thanks! :+1: 

Follow up work will be: https://github.com/openwisp/openwisp-radius/issues/304.

PR #303, opened 2021-09-25
[bug] Make 'is_sms_verification_enabled' respect global settings#301

### LINKED ISSUE #301: [bug] Make  "is_sms_verification_enabled" respect global settings
I have done some similar work in https://github.com/openwisp/openwisp-radius/pull/300 for `registration_enabled` which can be used as reference. 

https://github.com/openwisp/openwisp-radius/blob/602f3ecf357c3b4ecba0a60dabedd6f9c28f1319/openwisp_radius/base/models.py#L1142-L1145

https://github.com/openwisp/openwisp-radius/blob/602f3ecf357c3b4ecba0a60dabedd6f9c28f1319/openwisp_radius/api/utils.py#L55-L62

iamvaishnav:
closes #301


nemesisdesign (changes requested):
@iamvaishnav the build is faling for commit message guideline issues, please read that part of the docs more carefully, it's linked in the build failure.

Codewise looks good, I wonder if an additional test here may be useful or not, @pandafy if you could look into this as well it would be helpful, thanks.

openwisp_radius/base/models.py
```diff
@@ -1139,6 +1139,11 @@ def allowed_mobile_prefixes_list(self):
             mobile_prefixes = self.allowed_mobile_prefixes.split(',')
         return mobile_prefixes
 
+    def get_sms_verification_enabled(self):
+        if self.sms_verification is None:
+            return app_settings.SMS_DEFAULT_VERIFICATION
+        return self.sms_verification
+
     def clean(self):
         if self.sms_verification and not self.sms_sender:
```
pandafy, line 1148: This needs to be updated to use `get_sms_verification()` method.

openwisp_radius/base/models.py
```diff
@@ -1139,6 +1139,11 @@ def allowed_mobile_prefixes_list(self):
             mobile_prefixes = self.allowed_mobile_prefixes.split(',')
         return mobile_prefixes
 
+    def get_sms_verification_enabled(self):
```
pandafy, line 1142: Since the field name is `sms_verification`, it will be better to rename this to `get_sms_verification`.

pandafy (changes requested):
I think a explicit unit test for `is_sms_verification_enabled` will be helpful to catch any bug that may appear in future.  

@nemesisdesign I found a discrepancy in the field definition for `sms_verification` and `registration_enabled`

https://github.com/openwisp/openwisp-radius/blob/04cb6754a35682f24d20de214605bea585a439c8/openwisp_radius/base/models.py#L1048-L1055

https://github.com/openwisp/openwisp-radius/blob/04cb6754a35682f24d20de214605bea585a439c8/openwisp_radius/base/models.py#L1116-L1118

Check `default` and `null` parameter in both of the fields. 

With current field definition, the work done by @iamvaishnav with never get executed, since the sms_verification field will never be `None`.

openwisp_radius/base/models.py
iamvaishnav: I will make the requested changes.

openwisp_radius/base/models.py
iamvaishnav: Are you suggesting to just return `self_verification` as it will never be `None`?

nemesisdesign: I guess he meant `if self.get_sms_verification() and not self.sms_sender:`, right @pandafy?

pandafy: Yes, you are right @nemesisdesign 
```suggestion
        if self.get_sms_verification() and not self.sms_sender:
```

iamvaishnav:
> Check `default` and `null` parameter in both of the fields.

Should I add `null`  in `sms_verifications` field too? If so, how do I make migrations? Sorry if it sounds silly.

nemesisdesign (commented):
I see a few things that are not very consistent with the rest of the codebase here which would be great to fix in this PR:

- I would rename `SMS_DEFAULT_VERIFICATION` to `SMS_VERIFICATION_ENABLED`
- I would allow the field to be `None` as the other fields, if the field is None it means the gobal setting should be used ( SMS_VERIFICATION_ENABLED`)
- I would document this setting in https://github.com/openwisp/openwisp-radius/blob/master/docs/source/user/settings.rst 

How does this sound @iamvaishnav @pandafy?

iamvaishnav:
> I see a few things that are not very consistent with the rest of the codebase here which would be great to fix in this PR:
> 
> * I would rename `SMS_DEFAULT_VERIFICATION` to `SMS_VERIFICATION_ENABLED`
> * I would allow the field to be `None` as the other fields, if the field is None it means the gobal setting should be used (`SMS_VERIFICATION_ENABLED`)
> * I would document this setting in https://github.com/openwisp/openwisp-radius/blob/master/docs/source/user/settings.rst
> 
> How does this sound @iamvaishnav @pandafy?

Sounds good to me.

nemesisdesign:
@iamvaishnav is this ready for review yet?

iamvaishnav:
I don't know why it still fails.

codesankalp:
> I don't know why it still fails.

It needs to be updated with the master branch.

nemesisdesign (changes requested):
Great progress @iamvaishnav, there's a few inconsistencies to adjust from a visual point of view.

Let's see:

Screenshot from 2021-11-03 19-08-56

- I think we should move the sms option after "Needs identity verification:".
- The allowed options of the select should be identical to the fields above, while now are very different.

I think it's because this kind of code is missing:

https://github.com/openwisp/openwisp-radius/blob/ee2d4f37ad3859aec08efc02d4ad3d8531863dc6/openwisp_radius/admin.py#L588-L603

I don't see any change in the admin.

iamvaishnav:
> I don't see any change in admin.py, but I think it's necessary to do a bit more work to achieve consistency, can you do it please?

Will do.

nemesisdesign:
> > I don't see any change in admin.py, but I think it's necessary to do a bit more work to achieve consistency, can you do it please?
> 
> Will do.

Cool, ping us when ready @iamvaishnav :+1:

iamvaishnav:
@nemesisdesign Could you review it?

openwisp_radius/migrations/0025_sms_verification.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 3.1.13 on 2021-11-12 15:01
```
pandafy, line 1: Why there are two migrations files? Can you merge them into one?

openwisp_radius/admin.py
```diff
@@ -617,6 +618,22 @@ class OrganizationRadiusSettingsForm(AlwaysHasChangedMixin, forms.ModelForm):
         help_text=_IDENTITY_VERIFICATION_ENABLED_HELP_TEXT,
         fallback='',
```
pandafy, line 619: @nemesisdesign shouldn't the fallback value be set to `False` here?

pandafy (changes requested):
Good work @iamvaishnav, we are near completion of this PR. 
I requested a change below. After that it should be ready to merge. 

Can you also rebase on the current master branch?

openwisp_radius/migrations/0025_sms_verification.py
iamvaishnav: Will do.

openwisp_radius/admin.py
nemesisdesign: @pandafy why should it be?

pandafy: You are right, this is correct.

nemesisdesign (changes requested):
@iamvaishnav looks good! Please can you upgrade your version of black and run `openwisp-qa-format`? Then let's just make sure the build passes and we're done.

iamvaishnav:
> Please can you upgrade your version of black and run `openwisp-qa-format`

`openwisp-qa-format` is missing in this repo.

nemesisdesign:
> > Please can you upgrade your version of black and run `openwisp-qa-format`
> 
> `openwisp-qa-format` is missing in this repo.

Please read this section carefully: https://openwisp.io/docs/developer/contributing.html#python-code-conventions.
Install the latest master of openwisp-utils in your virtualenv.

nemesisdesign (changes requested):
The build is now failing because of the migration in sample_radius, please can you fix it?
Are you running `./run-qa-checks` locally before pushing as explained in https://openwisp.io/docs/developer/contributing.html#python-code-conventions?

iamvaishnav:
> Are you running `./run-qa-checks` locally before pushing as explained in

Yes

nemesisdesign (commented):
Build passes, @pandafy let's do one final manual check and then once all is confirmed to be good we can merge :+1:.

pandafy (approved):
I did a manual testing and everything works as expected. 

LGTM 👍🏼

nemesisdesign (approved):
Thanks @iamvaishnav @pandafy :+1:

Issue #304, opened 2021-09-27
[change] Users signup should respect the stricter verification settings of org


codesankalp:
After https://github.com/openwisp/openwisp-radius/issues/297

Suppose two organizations (further abbreviated as org.):
1. Org A needs no identity verification for its users.
2. Org B needs identity verification for its users.

If a user registered with org A tries to login into org B then there is no verification happens of that user and the user is authenticated and authorized to use other things.

So, if the org needs verification then the user should be verified before adding that user to the group of that org.

nemesisdesign:
I think that to discern users who are verified and not who we should add a model method which checks whether the registered_user.is_verified is True and registered_user.method is not unspecified or email. Having this as a model method is useful because it will allow the method to be overridden if needed.

pandafy:
I have a little confusion regarding the flow of cross-organization registration. 

**The problem:**

1. User has already registered to organization A. Organization A does not require identity verification, user is unverified. 
2. User tries to register to organization B, where he gets notified that he has already registered to organization A (409 conflict) 
3. User tries to login via URL for organization B (`radius/organization/org-b/account/token/`), but organization B requires identity verification.

**Questions:** 

1. What should we return in response to step 3? Should we return that the user needs to verify identity? 
This information can be used by wifi-login-pages to render the identity verification rule. 

2. How to decide which verification method to use (SMS or other)? 
Ideally, users should have the freedom to choose any verification method. 
I am not sure about this, but we might need to add another view in wifi-login-pages.

nemesisdesign:
> I have a little confusion regarding the flow of cross-organization registration.
> 
> **The problem:**
> 
> 1. User has already registered to organization A. Organization A does not require identity verification, user is unverified.
> 2. User tries to register to organization B, where he gets notified that he has already registered to organization A (409 conflict)
> 3. User tries to login via URL for organization B (`radius/organization/org-b/account/token/`), but organization B requires identity verification.

I think for this release we can simply return 403 if this happens to avoid a security breach.

> **Questions:**
> 
> 1. What should we return in response to step 3? Should we return that the user needs to verify identity?
>    This information can be used by wifi-login-pages to render the identity verification rule.
> 2. How to decide which verification method to use (SMS or other)?
>    Ideally, users should have the freedom to choose any verification method.
>    I am not sure about this, but we might need to add another view in wifi-login-pages.

[...]

PR #315, opened 2021-10-07
[fix] Unverified registered users cannot register cross organization #304

pandafy:
Unverified registered users cannot register to organizations that
need identity verification using cross-organization registration.
An HTTP 403 (PermissionDenied) response will be reverted to such requests.

Related to #304

### DIFF
--- a/docs/source/user/settings.rst
+++ b/docs/source/user/settings.rst
@@ -530,8 +530,9 @@ For example:
     additional registration methods which are supported by multiple
     custom modules).
 
-    Pass ``strong_identity`` as ``True`` to flag users registering
-    using that method as verified.
+    Pass ``strong_identity`` as ``True`` to to indicate that users who
+    register using that method have indirectly verified their identity
+    (eg: SMS verification, credit card, national ID card, etc).
 
 .. warning::
 


codesankalp:
I tested this with **OpenWISP WiFi Login Pages**, I tried these things:

I have two organizations:
1. `org1`: with simple email registration.
2. `org2`: with SMS verification enabled.

Steps I did in openwisp-wifi-login-pages:

1. **Registration**: Registered the users in `org1` and `org2` using the registration page i.e. `org1_user` and `org2_user`.
2. **Login**: Tested login for both the users and was able to log in successfully in both the organization.
3. **Cross Organization Login**: `org2_user` was easily able to login in the `org1`. `org1_user` when tried to login in the `org2` then error toast of permission denied shows up as expected because the `org1_user` is unverified.
        img
4. **Login after changing verification method**: In the admin dashboard I edited details of `org1_user` i.e. changed registration_method to Mobile phone verification and it was already unverified. After this, I tried to log in again in `org2` and it shows permission denied toast again. Also, login in `org1` is still successful by `org1_user`.
5. **Login after adding OrganizationUser related to mobile verification org**: After adding an organization user from admin with registration method as mobile verification and organization as `org2`, I was able to login successfully in wifi login pages.

PR #328, opened 2021-10-18
[fix] Migration's dependencies for extended openwisp-users apps #325

### LINKED ISSUE #325: [bug] Dependencies reference nonexistent parent node
Hello, 
I want to use the extended version of OpenWisp-users in my Django project. Currently, I'm using OpenWisp-users with of OpenWisp-radius. 
I followed [this document](https://github.com/openwisp/openwisp-users#extend-openwisp-users) for extending the users app and do all steps as said in the document. My problem is at s[tep 9 ](https://github.com/openwisp/openwisp-users#9-create-database-migrations), when i execute commad "makemigrations", it gives erros message as below: 
`django.db.migrations.exceptions.NodeNotFoundError: Migration openwisp_radius.0004_default_permissions dependencies reference nonexistent parent node ('openwisp_users', '0004_default_groups')
`
So i check migrations file with name '0004_default_permissions' of OpenWisp-radius app and as is shown, is really needed dependency from OpenWisp-users, While I disabled this app in INSTALLED_APPS in settings.py: 

```python
class Migration(migrations.Migration):
    dependencies = [
        ('openwisp_users', '0004_default_groups'),
        ('openwisp_radius', '0003_default_radius_groups'),
    ]
    operations = [
        migrations.RunPython(
            assign_permissions_to_groups, reverse_code=migrations.RunPython.noop
        )
    ]
```
Can anyone tell me what is the problem? Is it possible that may migrations are wrong?
Thanks.

psk007:
Add conditions to the migrations that has dependency from OpenWisp-users when OpenWisp is extended to a new app.
Closes #325


psk007:
> 
> Hi @psk007, thanks for submitting this bug report! +1
> 
> Looking at the patch provided in #328, I see that openwisp-radius references specific migrations in openwisp-users, that means that using plain openwisp-radius with an extended/customized version of openwisp-users would most probably not work and requires fixing, let me ask you, is this your specific case?
> 
> Do you have an extended version of OpenWISP Users and using plain (not extended) OpenWISP RADIUS?
> 
> As a last resort in some cases we could point to `__latest__` if doable with the swappable dependency.
>

Hello @nemesisdesign, 
Yes you are true, In my case I'm using an extended version of Users and plain version of Radius, and this bug only occurs in this case :).
About my solution, I agree with you that is not the definitive solution and more accurate solutions is needed but currently I have no idea and used this temporary solution. There are many cases that each of them needs specific solution and we can't provide one solution of all of them.

Since in the extended version developer is creating  its own version of `Users-app` and we don't know about that what he/she want to do, I think that is better to remove dependencies that referenced to OenpWISP-Users and leave it to developers for adopting their code to `OpenWISP-Radius` and to be sure that they know this hint add a section for `Using an extended version of OpenWISP-USERS with plain version of OpenWISP-RADIUS` in [this](https://github.com/openwisp/openwisp-users/#extend-openwisp-users) or [this](https://openwisp-radius.readthedocs.io/en/latest/developer/how_to_extend.html) document.

openwisp_radius/migrations/0002_initial_openwisp_radius.py
```diff
@@ -28,12 +29,14 @@ class Migration(migrations.Migration):
...
     dependencies = [
-        ('openwisp_users', '0003_default_organization'),
         migrations.swappable_dependency(settings.AUTH_USER_MODEL),
         ('openwisp_radius', '0001_initial_freeradius'),
     ]
 
     operations = [
+        migrations.RunPython(
+            set_default_organization_uuid, reverse_code=migrations.RunPython.noop
+        ),
```
nemesisdesign, line 39: @codesankalp: this can cause issues on PostgreSQL, please read:
https://docs.djangoproject.com/en/3.1/ref/migration-operations/#runpython

This has already happened a few times to me and other contributors (eg: @atb00ker, @pandafy).

I think we have two options:

1. find a way to run the migrations of OpenWISP RADIUS only once all the migrations of OpenWISP Users have been run, this can be achieved by referring to `__latest__`, but we won't be able to use `migrations.swappable_dependency` because it points to `__first__` (for good reasons), see the [django source code](https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/db/migrations/migration.py#L216-L218) for reference. I think we should implement an alternative to the function just linked which does the same but points to `__latest__`; 
2. run the data migration in https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/migrations/0003_default_radius_groups.py, which doesn't do schema changes.

I think we should look for a generic solution that we can reuse in other modules, since the problem will appear there too. That's why I think option 1 may be the best for our use case since once we make it work we just have to use it in the same way in other modules.
Referring to `__latest__` can have drawbacks, however I believe that in our case we will be fine because we only add foreign keys to models like `User` and `Organization` and do not refer to anything else which may change abrutly. We can consider these two models as stable.

nemesisdesign (changes requested):

nemesisdesign (approved):
Thanks!

Follow up: https://github.com/openwisp/openwisp-radius/issues/347.

PR #332, opened 2021-10-19
[fix] Handled SAML authentication by a logged in user

pandafy:
Bug:
If for some reason a user stays authenticated even after logging out,
and then tries to authenticate via SAML, they will get redirected to the
"RelayState" query parameter present in the URL. This might not play
well with frontend applications like openwisp-wifi-login-pages.

Fix:
Logout the user before proceeding with SAML login flow


devkapilbansal:
Thanks for the explanation :raised_hands: 
It is clear now :sweat_smile:

PR #334, opened 2021-11-02
[feature] Add support for RADIUS group checks and replies (sqlcounter) #164

### LINKED ISSUE #164: [feature] Allow enforcing session limits (sqlcounter) set in group checks through REST API
It may be possible and useful to implement the session limits defined in group checks directly in the REST API without the need of configuring the `sqlcounter` in freeradius.

The advantages of doing so could be:

- configuration of freeradius would result simpler
- we could make the API view extensible and customizable, opening up to scenarios that now are not even thinkable with the freeradius `sqlcounter` module
- it would support multi-tenancy out of the box

This is just a proposal for the future, I'll leave it here in case somebody is interested.

nemesisdesign:
Implements and closes #164.

### DIFF
--- a/docs/source/user/saml.rst
+++ b/docs/source/user/saml.rst
@@ -36,7 +36,7 @@ a SSO.
 Setup
 -----
 
-Install required system dependencies:
+Install required system dependencies::
 
     sudo apt install xmlsec1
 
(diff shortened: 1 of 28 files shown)


nemesisdesign:
Only test coverage improvement and docs are remaining.

nemesisdesign:
This is ready! :rocket:

docs/source/user/saml.rst
```diff
@@ -36,7 +36,7 @@ a SSO.
 Setup
 -----
 
-Install required system dependencies:
+Install required system dependencies::
```
codesankalp, line 39:
```suggestion
Install required system dependencies:
```

nemesisdesign: the two semi-colons are needed to format the next line as a code block

PR #337, opened 2021-11-06
[ux:change] Send password reset emails using HTML theme #103

### LINKED ISSUE #103: [ux:change] Send password reset emails using HTML theme
Depends on openwisp/openwisp-utils#259.

Let's see if we can find a way to send the password reset emails using the HTML theme from openwisp/openwisp-utils#259.

Looking at the code, it seems the serializer uses django-allauth under the hood:

https://github.com/openwisp/openwisp-radius/blob/c8e628d6133121c0c0c8841e4d03a18a0bf4e856/openwisp_radius/api/serializers.py#L347-L362

Which in turn calls:

https://github.com/openwisp/openwisp-radius/blob/c8e628d6133121c0c0c8841e4d03a18a0bf4e856/openwisp_radius/templates/custom_password_reset_email.html#L1-L14

So it should be similar to what @devkapilbansal is doing in https://github.com/openwisp/openwisp-users/pull/295.

codesankalp:
Used openwisp-utils html theme to add support for html template
for resetting password.

Closes #103


openwisp_radius/settings.py
```diff
@@ -139,3 +139,12 @@ def get_settings_value(option, default):
...
 )
+
+OPENWISP_EMAIL_LOGO = getattr(
+    settings,
+    'OPENWISP_EMAIL_LOGO',
+    'https://raw.githubusercontent.com/openwisp/openwisp-utils/master/openwisp_utils/'
+    'static/openwisp-utils/images/openwisp-logo.png',
+)
+
+OPENWISP_HTML_EMAIL = getattr(settings, 'OPENWISP_HTML_EMAIL', True)
```
nemesisdesign, line 150: aren't these settings already defined in openwisp-utils? Why redefine them here

openwisp_radius/templates/html_password_reset_email.html
```diff
@@ -0,0 +1,35 @@
...
+        {% endblocktrans %}
+    </p>
+    <p>
+        {% trans "Please go to the following page and choose a new password:" %}
+    </p>
+</div>
+
+<div class="call-to-action">
+    {# Password reset confirm url #}
+    <a href="{{password_reset_url}}" target="_blank" class="btn">Reset Password</a>
```
nemesisdesign, line 26: `Reset Password` is not flagged as translatable, please also double check if there's already a string to use from django or django-allauth so we don't need to add translations for it again.

nemesisdesign (commented):

nemesisdesign (changes requested):
It seems to me that a test is missing here, we should check whether the email sent is using the HTML template.

openwisp_radius/templates/html_password_reset_email.html
```diff
@@ -0,0 +1,35 @@
+{% extends 'openwisp_utils/email_template.html' %}
+{% load i18n %}
+{% load l10n %}
+
+{% block mail_header %}
+{{ block.super }}
+<div class="subject">
+{% blocktrans %}Password reset on {{ site_name }}!{% endblocktrans %}
+</div>
+{% endblock %}
```
nemesisdesign, line 10: The generated output doesn't look right:

```html
      <div class="logo-container">
        <img src="" alt="Logo" id="logo" class="logo">
      </div>
      <div class="subject">
        
      </div>
      
<div class="subject">
Password reset on example.com!
</div>
```

I would remove the exclamation mark.

Can't you pass this string in the context argument of `send_email` above?

openwisp_radius/templates/html_password_reset_email.html
```diff
@@ -0,0 +1,35 @@
...
+        {% endblocktrans %}
+    </p>
+    <p>
+        {% trans "Please go to the following page and choose a new password:" %}
+    </p>
+</div>
+
+<div class="call-to-action">
+    {# Password reset confirm url #}
+    <a href="{{password_reset_url}}" target="_blank" class="btn">{% trans "Reset My Password" %}</a>
```
nemesisdesign, line 26: Reset My Password > Reset password

openwisp_radius/templates/html_password_reset_email.html
```diff
@@ -0,0 +1,35 @@
...
+
+{% block mail_body %}
+<div class="msg">
+    <p>
+        {% blocktrans %}
+        You're receiving this email because you requested a password reset for your user account at {{ site_name }}.
+        {% endblocktrans %}
+    </p>
+    <p>
+        {% trans "Please go to the following page and choose a new password:" %}
```
nemesisdesign, line 20: Let's change this: Please click on the button below to open a page where you can choose a new password.

openwisp_radius/api/serializers.py
```diff
@@ -347,6 +349,22 @@ class Meta:
...
+        to_email,
+        html_email_template_name=None,
+    ):
+        subject = loader.render_to_string(subject_template_name, context)
+        # Email subject *must not* contain newlines
+        subject = ''.join(subject.splitlines())
+        body = loader.render_to_string(email_template_name, context)
+        if html_email_template_name is not None:
+            body_html = loader.render_to_string(html_email_template_name, context)
+        send_email(subject, body, body_html, [to_email], context)
```
nemesisdesign, line 366: I would try to do this differently, I would define a template only for the body message and pass everything else as extra context.

```python
# here I would only put the HTML to render the message and nothing else
body_html = loader.render_to_string(html_email_template_name, context)
# this should
send_email(subject, body, body_html, [to_email], extra_context={
    'subject': '.......',
    'call_to_action_url': '....',
    'call_to_action_text': _('Reset password')
})
```

Can you try this?

openwisp_radius/templates/html_password_reset_email.html
codesankalp: @nemesisdesign `Reset password` translations are not present in django or allauth but `Password Reset` is present, should I change to that?

openwisp_radius/tests/test_api/test_api.py
```diff
@@ -703,7 +703,15 @@ def test_api_password_reset(self):
...
-
+        email = mail.outbox.pop()
+        self.assertIn(
+            "<p> Please click on the button below to open a page where you can",
+            ' '.join(email.alternatives[0][0].split()),
+        )
+        self.assertIn(
+            "class=\"btn\"> Reset password </a>",
+            ' '.join(email.alternatives[0][0].split()),
+        )
```
nemesisdesign, line 714: Can you also add an assertion like:

```python
self.assertNotIn('<img src=""', email.alternatives[0][0])
```

openwisp_radius/tests/test_api/test_api.py
```diff
@@ -703,7 +703,15 @@ def test_api_password_reset(self):
         reset_payload = {'email': 'test@email.com'}
         response = self.client.post(password_reset_url, data=reset_payload)
         self.assertEqual(len(mail.outbox), mail_count + 1)
-
+        email = mail.outbox.pop()
+        self.assertIn(
+            "<p> Please click on the button below to open a page where you can",
+            ' '.join(email.alternatives[0][0].split()),
```
nemesisdesign, line 709: why the join and splitting BTW?

nemesisdesign (commented):
Can you try depending on this branch of openwisp-users to see if the build gets fixed?
https://github.com/openwisp/openwisp-users/tree/issues/285-html-email-template

I still see `Logo` in my tests, I am not sure why, because the code looks correct, can you replicate this issue?

openwisp_radius/tests/test_api/test_api.py
codesankalp: because the Html contains spaces and `\n`

nemesisdesign (commented):
Locally it works fine, however when I started trying it in a prod-like environment, an excepion is raised:

```
Internal Server Error: /api/v1/radius/organization/{org}/account/password/reset/
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/openwisp_radius/api/views.py", line 127, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "/opt/openwisp2/env/lib/python3.7/site-packages/rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/openwisp_radius/api/views.py", line 463, in post
    return super().post(request, *args, **kwargs)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/dj_rest_auth/views.py", line 250, in post
    serializer.save()
  File "/opt/openwisp2/env/lib/python3.7/site-packages/dj_rest_auth/serializers.py", line 259, in save
    self.reset_form.save(**opts)
  File "/opt/openwisp2/env/lib/python3.7/site-packages/dj_rest_auth/forms.py", line 44, in save
    args=[user_pk_to_url_str(user), temp_key],
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/urls/base.py", line 87, in reverse
    return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))
  File "/opt/openwisp2/env/lib/python3.7/site-packages/django/urls/resolvers.py", line 685, in _reverse_with_prefix
    raise NoReverseMatch(msg)
django.urls.exceptions.NoReverseMatch: Reverse for 'password_reset_confirm' not found. 'password_reset_confirm' is not a valid view function or pattern name.
```
I investigated but couldn't find the exact cause.

Do you have any idea @codesankalp?

nemesisdesign (commented):
I found more info of the problem mentioned in my previous comment:

- https://dj-rest-auth.readthedocs.io/en/latest/faq.html#faq
- https://github.com/iMerica/dj-rest-auth/blob/2.1.11/demo/demo/urls.py#L40-L43

It seems we are missing something for this feature to work 100% but somehow I only noticed this now. I am pretty sure I did test this before and I am not sure why this error popped up recently, I also tested the same functionality on another instance which doesn't have this patch installed and I get the same error, hence it's not being caused by these changes.

Can you help me debug this and make it work please?

openwisp_radius/api/serializers.py
```diff
@@ -359,7 +360,11 @@ def save(self):
             'from_email': getattr(settings, 'DEFAULT_FROM_EMAIL'),
             'email_template_name': ('custom_password_reset_email.html'),
             'request': request,
-            'extra_email_context': {'password_reset_url': password_reset_url},
+            'extra_email_context': {
+                'subject': _(f'Password reset on {get_current_site(request).name}'),
```
nemesisdesign, line 364: This will not work.. f strings do not work with gettext translations, please refer to the last paragraph of this django docs section https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#standard-translation.
Please convert it as the django docs recommends and then please add the translations for this message to all the supported locales (using gtranslate).

nemesisdesign (changes requested):
Ignore my previous comments! I found out a misconfiguration on both systems I was testing.

The only remaining problem is that the subject is not translated:

Password reset on {site_name}

nemesisdesign (commented):
Looks good, waiting until openwisp-utils is merged so we can update the dependency and merge this too.

nemesisdesign (approved):

PR #345, opened 2021-11-09
[feature] Added `modified` field to Registered User #340

### LINKED ISSUE #340: [feature] Add "modified" (last modified date) field to RegisteredUser
This info is useful to know when something changed regarding the verification of the user.

PS: I think we can simply make this model inherit from `TimestampEditableModel` like all other models for consistency.
Then we can hide the "added" field from the admin and change the label of "modified" to say something like "Last verification changed at". The problem with  `TimestampEditableModel` is that it tries to add also an id field using UUIDField, but we do not need that in this model, since the primary-key is the foreign key to the organization.

We must double check the modified field is automatically updated when the verification status changes in either direction.

codesankalp:
Closes #340


nemesisdesign (approved):
Thanks @codesankalp :+1:

PR #349, opened 2021-11-15
[change] Changed delete_unverified_users to avoid deleting users which have accounting sessions #348

### LINKED ISSUE #348: [change] Change delete_unverified_users to avoid deleting users which have accounting sessions
In some cases, users are flagged as unverified after a certain amount of time to force them to update their verification details and we can't delete these users automatically.

We should therefore update this functionality to look whether the user has any radius session related to them and only if they do not have any radius session we can delete their account automatically.

https://github.com/openwisp/openwisp-radius/blob/fc4a1abec802e2a755092b1c2eaa2a4982f9a201/openwisp_radius/management/commands/base/delete_unverified_users.py#L30-L34

Here we should loop over the available result instead of deleting them, and check whether any RadiusAccounting with their username exists (using the `exists()` of the django ORM).

We should also update the docs to reflect this and to clarify that this feature is needed to automatically delete users who register and never verify their account.

codesankalp:
Closes #348

### DIFF
--- a/docs/source/user/management_commands.rst
+++ b/docs/source/user/management_commands.rst
@@ -89,7 +89,8 @@ Note that the default duration is set to 18 months.
 ---------------------------
 
 This command deletes unverified users that have been registered for
-more than specified duration.
+more than specified duration. This feature is needed to automatically
+delete users who register and never verify their accounts.
 
 .. code-block:: shell
 
--- a/docs/source/user/management_commands.rst
+++ b/docs/source/user/management_commands.rst
@@ -89,8 +89,9 @@ Note that the default duration is set to 18 months.
 ---------------------------
 
 This command deletes unverified users that have been registered for
-more than specified duration. This feature is needed to automatically
-delete users who register and never verify their accounts.
+more than specified duration and have no associated radius session.
+This feature is needed to delete users who have registered but never
+completed the verification process.
 
 .. code-block:: shell
 
--- a/openwisp_radius/management/commands/base/delete_unverified_users.py
+++ b/openwisp_radius/management/commands/base/delete_unverified_users.py
@@ -30,11 +30,15 @@ class BaseDeleteUnverifiedUsersCommand(BaseCommand):
     def handle(self, *args, **options):
         days = now() - timedelta(days=int(options['older_than_days']))
         exclude_methods = str(options['exclude_methods']).split(',')
-        for user in User.objects.filter(
-            date_joined__lt=days,
-            registered_user__isnull=False,
-            registered_user__is_verified=False,
-        ).exclude(registered_user__method__in=exclude_methods):
+        for user in (
+            User.objects.filter(
+                date_joined__lt=days,
+                registered_user__isnull=False,
+                registered_user__is_verified=False,
+            )
+            .exclude(registered_user__method__in=exclude_methods)
+            .iterator()
+        ):
             if not RadiusAccounting.objects.filter(username=user.username).exists():
                 user.delete()
 
--- a/openwisp_radius/management/commands/base/delete_unverified_users.py
+++ b/openwisp_radius/management/commands/base/delete_unverified_users.py
@@ -4,7 +4,10 @@ from django.contrib.auth import get_user_model
 from django.core.management import BaseCommand
 from django.utils.timezone import now
 
+from openwisp_radius.utils import load_model
+
 User = get_user_model()
+RadiusAccounting = load_model('RadiusAccounting')
 
 
 class BaseDeleteUnverifiedUsersCommand(BaseCommand):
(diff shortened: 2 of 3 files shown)


openwisp_radius/management/commands/base/delete_unverified_users.py
```diff
@@ -27,11 +30,14 @@ def add_arguments(self, parser):
...
-        User.objects.filter(
+        for user in User.objects.filter(
             date_joined__lt=days,
             registered_user__isnull=False,
             registered_user__is_verified=False,
-        ).exclude(registered_user__method__in=exclude_methods).delete()
+        ).exclude(registered_user__method__in=exclude_methods):
+            if not RadiusAccounting.objects.filter(username=user.username).exists():
+                user.delete()
+
```
codesankalp, line 40: I followed the issue description, but can we do this?
```suggestion
User.objects.filter(
            date_joined__lt=days,
            registered_user__isnull=False,
            registered_user__is_verified=False,
        ).exclude(registered_user__method__in=exclude_methods, username__in=RadiusAccounting.objects.all().values_list('username')).delete()

```

pandafy: `RadiusAccounting.objects.all().values_list('username')` this is a very costly database query. 
If I am not wrong, It will fetch username of all RadiusAccounting objects in the database first and then execute the other query. You can check the raw SQL queries to verify this.

openwisp_radius/tests/test_commands.py
```diff
@@ -163,7 +163,7 @@ def test_prefix_add_users_command(self):
         with self.assertRaises(SystemExit):
             self._call_command('prefix_add_users', **options)
 
-    @capture_stdout()
+    # @capture_stdout()
```
pandafy, line 166: Looks like a leftover

openwisp_radius/tests/test_commands.py
```diff
@@ -225,6 +225,27 @@ def _create_old_users():
...
+                user=user,
+                method='email',
+                is_verified=False,
+            )
+            opts = dict(
+                username=user.username,
+                unique_id='2',
+                nas_ip_address='127.0.0.1',
+                session_id='1',
+            )
```
pandafy, line 241:
```suggestion
            opts = options = _RADACCT.copy()
            options['username'] = user.username
```
Consider using above since it is how other tests are written.

pandafy (changes requested):
Good work @codesankalp! Some minor changes requested.

openwisp_radius/tests/test_commands.py
codesankalp: Thanks for pointing it out.

pandafy (approved):
LGTM 👍🏼

openwisp_radius/management/commands/base/delete_unverified_users.py
```diff
@@ -27,11 +30,14 @@ def add_arguments(self, parser):
     def handle(self, *args, **options):
         days = now() - timedelta(days=int(options['older_than_days']))
         exclude_methods = str(options['exclude_methods']).split(',')
-        User.objects.filter(
+        for user in User.objects.filter(
             date_joined__lt=days,
             registered_user__isnull=False,
             registered_user__is_verified=False,
-        ).exclude(registered_user__method__in=exclude_methods).delete()
+        ).exclude(registered_user__method__in=exclude_methods):
```
pandafy, line 37:
```suggestion
        ).exclude(registered_user__method__in=exclude_methods).iterator():
```
I missed this in my earlier reviews. We should consider using `iterator()` whenever we are looping over a queryset. Read about iterator here: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#iterator

docs/source/user/management_commands.rst
```diff
@@ -89,7 +89,8 @@ Note that the default duration is set to 18 months.
 ---------------------------
 
 This command deletes unverified users that have been registered for
-more than specified duration.
+more than specified duration. This feature is needed to automatically
+delete users who register and never verify their accounts.
```
nemesisdesign, line 93:
```suggestion
more than specified duration and have no associated radius session.
This feature is needed to delete users who have registered but never
completed the verification process.
```

openwisp_radius/management/commands/base/delete_unverified_users.py
nemesisdesign: Yes please.

openwisp_radius/tests/test_commands.py
```diff
@@ -225,6 +225,24 @@ def _create_old_users():
...
+            opts = _RADACCT.copy()
+            opts['unique_id'] = 1
+            opts['username'] = user.username
+            self._create_radius_accounting(**opts)
+            self.assertEqual(User.objects.count(), 4)
+            call_command(
+                'delete_unverified_users',
+            )
+            self.assertEqual(User.objects.count(), 1)
```
nemesisdesign, line 244: please assert that the remaining user is the one with the radius accounting associated to it.

nemesisdesign (changes requested):

openwisp_radius/tests/test_commands.py
```diff
@@ -242,6 +242,12 @@ def _create_old_users():
                 'delete_unverified_users',
             )
             self.assertEqual(User.objects.count(), 1)
+            self.assertEqual(
+                RadiusAccounting.objects.filter(
+                    username=User.objects.first().username
```
nemesisdesign, line 247: isn't the username already in `opts['username']`?

nemesisdesign (changes requested):

nemesisdesign (approved):
I fixed https://github.com/openwisp/openwisp-radius/pull/349#pullrequestreview-807746535 in https://github.com/openwisp/openwisp-radius/pull/349/commits/494878829d6727f72348b0d9ade38d55da04b67f.

PR #350, opened 2021-11-15
[feature] Emit signal in freeradius accounting view #341

### LINKED ISSUE #341: [feature] Emit signal when requests to the freeradius accounting view are processed successfully
Before a successful response is returned from the freeradius accounting view, we need to emit a signal so we can connect to it from other parts of the application or from third party apps.

The signal should not be emitted if there was any error.

The signal could be called `radius_accounting_success`, we should pass as argument the radius accounting object instance and the request object.

We shall add tests and docs for this signal. As for the docs, I think we'll need a new page called `signals.rst` in [docs/source/developer](https://github.com/openwisp/openwisp-radius/tree/master/docs/source/developer), the signal should be documented as we document signals in other modules (eg: openwisp-controller).

devkapilbansal:
Closes #341

### DIFF
--- a/docs/source/developer/signals.rst
+++ b/docs/source/developer/signals.rst
@@ -9,6 +9,8 @@ Signals
 
 **Arguments**:
 
-- ``accounting_data``: accounting information
+- ``sender`` (``str``): the view that sends the signal
+- ``accounting_data`` (``dict``): accounting information
 
-This signal is emitted every time accounting information is added or updated.
+This signal is emitted every time the accounting REST API endpoint
+completes successfully, just before the response is returned.
--- a/docs/source/developer/signals.rst
+++ b/docs/source/developer/signals.rst
@@ -9,9 +9,9 @@ Signals
 
 **Arguments**:
 
-- ``sender`` : instance of ``AccountingView``
+- ``sender`` : class of ``AccountingView`` instance.
 - ``accounting_data`` (``dict``): accounting information
-- ``view``: view object which calls the signal
+- ``view``: instance of ``AccountingView``
 
 This signal is emitted every time the accounting REST API endpoint
 completes successfully, just before the response is returned.
--- a/docs/source/developer/signals.rst
+++ b/docs/source/developer/signals.rst
@@ -1,0 +1,14 @@
+=======
+Signals
+=======
+
+``radius_accounting_success``
+-----------------------------
+
+**Path**: ``openwisp_radius.signals.radius_accounting_success``
+
+**Arguments**:
+
+- ``accounting_data``: accounting information
+
+This signal is emitted every time accounting information is added or updated.
--- a/docs/source/developer/signals.rst
+++ b/docs/source/developer/signals.rst
@@ -11,6 +11,10 @@ Signals
 
 - ``sender`` : instance of ``AccountingView``
 - ``accounting_data`` (``dict``): accounting information
+- ``view``: view object which calls the signal
 
 This signal is emitted every time the accounting REST API endpoint
 completes successfully, just before the response is returned.
+
+The ``view`` argument can also be used to access the ``request``
+object i.e. ``view.request``.
--- a/docs/source/developer/signals.rst
+++ b/docs/source/developer/signals.rst
@@ -9,7 +9,7 @@ Signals
 
 **Arguments**:
 
-- ``sender`` (``str``): the view that sends the signal
+- ``sender`` : instance of ``AccountingView``
 - ``accounting_data`` (``dict``): accounting information
 
 This signal is emitted every time the accounting REST API endpoint
--- a/docs/source/developer/signals.rst
+++ b/docs/source/developer/signals.rst
@@ -9,7 +9,7 @@ Signals
 
 **Arguments**:
 
-- ``sender`` : class of ``AccountingView`` instance.
+- ``sender`` : ``AccountingView``
 - ``accounting_data`` (``dict``): accounting information
 - ``view``: instance of ``AccountingView``
 
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -378,6 +378,20 @@ stop_time           Stop time (less or equal to)
 is_open             If stop_time is null
 ==================  ====================================
 
+Signals
++++++++
+
+``radius_accounting_success``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Path**: ``openwisp_radius.signals.radius_accounting_success``
+
+**Arguments**:
+
+- ``accounting_data``: accounting information
+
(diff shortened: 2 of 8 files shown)


openwisp_radius/signals.py
```diff
@@ -0,0 +1,3 @@
+from django.dispatch import Signal
+
+radius_accounting_success = Signal(providing_args=['account_data'])
```
nemesisdesign, line 3: Please chagne `account_data` to `accounting_data` here and anywhere else, because account and accounting are two different concepts which must not be confused.

nemesisdesign (commented):
It looks good, please add the docs.

docs/source/user/api.rst
```diff
@@ -378,6 +378,20 @@ stop_time           Stop time (less or equal to)
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Path**: ``openwisp_radius.signals.radius_accounting_success``
+
+**Arguments**:
+
+- ``accounting_data``: accounting information
+
+This signal is emitted every time accounting information is added or updated.
+
```
nemesisdesign, line 394: Please let's add a new page as requested in the [issue description](https://github.com/openwisp/openwisp-radius/issues/341) because this will likely grow in the near future.

nemesisdesign (changes requested):

docs/source/developer/signals.rst
```diff
@@ -0,0 +1,14 @@
...
+``radius_accounting_success``
+-----------------------------
+
+**Path**: ``openwisp_radius.signals.radius_accounting_success``
+
+**Arguments**:
+
+- ``accounting_data``: accounting information
+
+This signal is emitted every time accounting information is added or updated.
```
nemesisdesign, line 14: Let's make sure the following concepts are extra clear:

- the signal is emitted every time the accounting REST API endpoint completes successfully
- the signal is emitted right before the response is returned
- accounting_data is a `dict`

nemesisdesign (changes requested):

openwisp_radius/api/freeradius_views.py
```diff
@@ -445,6 +448,11 @@ def _data_to_acct_model(self, valid_data):
         valid_data['organization'] = acct_org
         return valid_data
 
+    def send_radius_accounting_signal(self, accounting_data):
+        radius_accounting_success.send(
+            sender=self.__class__, accounting_data=accounting_data
```
nemesisdesign, line 453: The sender is the view, I noticed only now, it makes sense, but we haven't specified this in the docs, so users would have to dig in the code to find out, which is means the documentation is not good enough. Please fix this.

nemesisdesign (changes requested):

openwisp_radius/api/freeradius_views.py
```diff
@@ -431,12 +432,14 @@ def post(self, request, *args, **kwargs):
...
             serializer.create(acct_data)
             headers = self.get_success_headers(serializer.data)
+            self.send_radius_accounting_signal(acct_data)
             return Response(None, status=201, headers=headers)
         else:
             serializer = self.get_serializer(instance, data=data, partial=False)
             serializer.is_valid(raise_exception=True)
             acct_data = self._data_to_acct_model(serializer.validated_data.copy())
             serializer.update(instance, acct_data)
+            self.send_radius_accounting_signal(acct_data)
```
codesankalp, line 442: @devkapilbansal Instead of `acct_data`, can we pass `serializer.data`?
Because while passing this data to celery I am getting:
`kombu.exceptions.EncodeError: Object of type Organization is not JSON serializable`

`_data_to_acct_model` removes `status_type` and adds `organization`. Also, `status_type` is more required in detecting the radius status i.e. `Start`.

nemesisdesign: Makes sense.

openwisp_radius/api/freeradius_views.py
nemesisdesign: I thought that it could be useful to pass the view instance (self) instead of the class, because that allows accessing a lot of information about the request and it can be useful to developers who may want to use this signal. Can you please update it here and in the docs?

nemesisdesign (changes requested):

nemesisdesign (approved):
This looks good now :+1: 

I will do some manual testing and then merge it.

openwisp_radius/api/freeradius_views.py
```diff
@@ -445,6 +448,9 @@ def _data_to_acct_model(self, valid_data):
         valid_data['organization'] = acct_org
         return valid_data
 
+    def send_radius_accounting_signal(self, accounting_data):
+        radius_accounting_success.send(sender=self, accounting_data=accounting_data)
```
codesankalp, line 452: When we are passing `self` from here it is a view object which we can't specify as sender while connecting to receivers.
We must use `__class__` to get the class view and not the view object.
cc: @nemesisdesign

nemesisdesign: @codesankalp we know it's a view instance, the change was done on purpose (django also passes view instances), what is the problem you're trying to describe? I don't understand.

nemesisdesign: Oh I think I understand now, in this case we don't need to take into account the sender though, because only this class will send this. Let's look at how django implements some of their request cycle signals and do it similarly (let's look if they use the view instance or class as sender).

openwisp_radius/signals.py
```diff
@@ -0,0 +1,3 @@
+from django.dispatch import Signal
+
+radius_accounting_success = Signal(providing_args=['accounting_data'])
```
nemesisdesign, line 3: `providing_args` doesn't play well with django 4 in which it was removed, please see the release notes of django 4 (lookfor `providing_args`) and update this code according to what is suggested there.

openwisp_radius/api/freeradius_views.py
nemesisdesign: @devkapilbansal Sankalp is right, let's do the following which should be more ortodox:

- restore the previous sender (the class)
- pass the request and view instance as an additional argument
- update docs

nemesisdesign (changes requested):

setup.py
```diff
@@ -36,7 +36,7 @@
     zip_safe=False,
     install_requires=[
         'django>=3.0,<3.2',
-        'swapper @ git+git://github.com/openwisp/django-swappable-models.git@master',
+        'swapper~=1.3.0',
```
nemesisdesign, line 39: please remove this because as I just said in the dev chat it's not needed anymore since it's defined in openwisp-utils.

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #351, opened 2021-11-16
[change] Save user preferred language in obtain and validate token views #342

### LINKED ISSUE #342: [change] Save user preferred language in obtain token and validate token views
Depends on https://github.com/openwisp/openwisp-users/issues/261.

We need to check whether the user preferred language stored is the same as what we receive from the request in these views and if not we shall update it.

We need this to be able to send emails out the system's own intiative with the right language.

codesankalp:
Closes #342


openwisp_radius/api/views.py
```diff
@@ -159,6 +160,23 @@ def get(self, request, *args, **kwargs):
 download_rad_batch_pdf = DownloadRadiusBatchPdfView.as_view()
 
 
+class UserLanguageMixin(object):
+    def save_language(self, user):
+        languages = self._parse_language_header(self.request)
```
codesankalp, line 165: I tried to use https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.translation.get_language_from_request, but it always returns en-gb even if language headers passed is different.

nemesisdesign: Instead of duplicating Django's internal code, which is never an acceptable solution, please use ipdb or any other debugging tool to find out what's going on when the django function is called: https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/utils/translation/trans_real.py#L515-L555

There must be an explanation for this.

openwisp_radius/api/views.py
```diff
@@ -159,6 +160,23 @@ def get(self, request, *args, **kwargs):
...
+        settings_languages = list(zip(*settings.LANGUAGES))[0]
+        for language in languages:
+            if language == user.language:
+                break
+            elif language in settings_languages:
+                user.language = language
+                user.save()
+                break
```
nemesisdesign, line 173: I'd like to avoid looping on all the languages in settings.LANGUAGES since there's many entries there, it looks like django's `get_language_from_request` already does this work in the most possible efficient way and we should find a way to leverage that instead of reimplementing it.

openwisp_radius/api/views.py
```diff
@@ -159,6 +160,23 @@ def get(self, request, *args, **kwargs):
...
+class UserLanguageMixin(object):
+    def save_language(self, user):
+        languages = self._parse_language_header(self.request)
+        settings_languages = list(zip(*settings.LANGUAGES))[0]
+        for language in languages:
+            if language == user.language:
+                break
+            elif language in settings_languages:
+                user.language = language
+                user.save()
```
nemesisdesign, line 172: we should always ensure:

- the save query is executed only if the language changed
- we call `user.save(update_fields=['language'])` to ensure the query only writes this specific field

Please keep in mind this operation should be as efficient as possible because will be executed often.

nemesisdesign (changes requested):
Good progress @codesankalp!
It requires polishing but at least we should be able to start using it to implement other parts.

Please update the API docs of these 2 endpoints to make clear the user language preference is updated.

docs/source/user/api.rst
```diff
@@ -577,6 +577,9 @@ they're not member of, then they would be automatically added as members
 (if registration is enabled for that org). Please refer to
 `"Registering to Multiple Organizations" <#registering-to-multiple-organizations>`_.
 
+This endpoint updates the user language preference from the `Accept-Language` HTTP
+header.
```
nemesisdesign, line 581:
```suggestion
This endpoint updates the user language preference field according
to the ``Accept-Language`` HTTP header.
```

Same below.

nemesisdesign (changes requested):
A lot better, I'll test asap, in the meanwhile see my comment below.

nemesisdesign (changes requested):
I did some testing and it works fine! :+1: 

After having looked at the code, I see that we are doing something very similar with `RadiusTokenMixin.update_user_last_login`, which already executes a save query to update the `last_login` field for the user.

I see that `RadiusTokenMixin` is also inherited from in `RegisterView`, which makes sense, because after registration the user is logged in automatically, so I think we should update the user lang field there too.

In order to avoid to execute 2 different queries to update 2 different fields and to maintain consistency, I think we should do the following:

- remove `update_user_last_login` from `RadiusTokenMixin` because it's not really related to the radius token functionality
- remove the `save_language`
- rename `UserLanguageMixin` to `UserDetailsUpdaterMixin`
- create a method called `update_user_details`, properly commented with a docstring, which executes both operations of `update_user_last_login` and `save_language` in one query
- add `UserLanguageMixin` to `RegisterView`

What do you think @codesankalp?

setup.py
```diff
@@ -39,7 +39,10 @@
         'swapper~=1.1.0',
         # Needed for the new authentication backend in openwisp-users
         # TODO: remove when the new version of openwisp-users is released
-        'openwisp-users @ https://github.com/openwisp/openwisp-users/tarball/master',
+        (
+            'openwisp-users @ https://github.com/openwisp/openwisp-users/'
+            'tarball/issue/261-user-preferred-lang'
```
nemesisdesign, line 44: you can now point to master again

nemesisdesign (changes requested):

codesankalp:
It will pass after https://github.com/openwisp/openwisp-radius/pull/359 is merged.

nemesisdesign (changes requested):
The migrations need to be updated, see the failing build (`django.db.migrations.exceptions.NodeNotFoundError: Migration sample_users.0005_user_language dependencies reference nonexistent parent node ('sample_users', '0004_user_notes')`).

codesankalp:
> The migrations need to be updated, see the failing build (`django.db.migrations.exceptions.NodeNotFoundError: Migration sample_users.0005_user_language dependencies reference nonexistent parent node ('sample_users', '0004_user_notes')`).

Resolved.

openwisp_radius/api/views.py
```diff
@@ -159,6 +160,17 @@ def get(self, request, *args, **kwargs):
...
 
+class UserDetailsUpdaterMixin(object):
+    def update_user_details(self, user):
+        language = get_language_from_request(self.request)
+        update_fields = ['last_login']
+        if user.language != language:
+            user.language = language
+            update_fields.append('language')
+        user.last_login = timezone.now()
+        user.save(update_fields=['last_login', 'language'])
```
nemesisdesign, line 171: you forgot to pass `update_fields` here

nemesisdesign (changes requested):

openwisp_radius/api/views.py
codesankalp: yes :sweat_smile:

nemesisdesign (approved):

nemesisdesign (approved):
:+1:

PR #355, opened 2021-11-19
[feature] Send email when radius accounting session starts #343 

### LINKED ISSUE #343: [feature] Send email when radius accounting session starts
This depends on the following features:

- https://github.com/openwisp/openwisp-radius/issues/342
- https://github.com/openwisp/openwisp-users/pull/290
- https://github.com/openwisp/openwisp-radius/issues/341

We shall send an email with a one time login link to the user once the radius accounting session starts.

The flow should be something like the following:

- if the radius accounting has just been created (eg: accounting type start)
- initialize a celery task in the background
- the celery task shall check whether there's any user in the organization of the radius accounting with the username specified in the radius accounting, if yes, it shall send the user an email with a text that communicates that a new session has been started and provide a one-time link to the user where they can close the session (log out from the networking service) and/or manage their account
- it would be great to make sure that if the one-time link has expired, the user sees the wifi login page (instead of the page which says the link has expired)
- I think we should add new fields to `OrganizationRadiusSettings` where we can let admins specify the login page URL (where users log in to the wifi service) and status page URL (where users can log out or manage their account)
- we have to make sure the language used for the translations is the user preferred language described in https://github.com/openwisp/openwisp-users/issues/261, because the radius accounting is started by freeradius calling our API, therefore we do not have the information of the preferred language coming from the browser, for that reason we have to get the preferred language from the DB and activate the translations for that language ([see translation.activate() in the django docs](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#explicitly-setting-the-active-language))

codesankalp:
Closes #343


codesankalp:
Should I create a new section in docs for this feature?

codesankalp:
I haven't updated the translations yet because text changes may be required.
I am working on writing tests.

openwisp_radius/tasks.py
```diff
@@ -39,3 +49,38 @@ def delete_unverified_users(older_than_days=1, exclude_methods=''):
...
+    User = get_user_model()
+    username = accounting_data.get('username', None)
+    org_uuid = accounting_data.get('organization')
+    user = User.objects.filter(username=username).first()
+    if user:
+        organization = (
+            user.openwisp_users_organization.all().filter(id=org_uuid).first()
+        )
+        if organization:
+            from sesame.utils import get_query_string
```
nemesisdesign, line 65: why this import here and not above?

openwisp_radius/tasks.py
```diff
@@ -39,3 +49,38 @@ def delete_unverified_users(older_than_days=1, exclude_methods=''):
...
 def convert_called_station_id(unique_id=None):
     management.call_command('convert_called_station_id', unique_id=unique_id)
+
+
+@shared_task
+def send_login_email(accounting_data):
+    User = get_user_model()
+    username = accounting_data.get('username', None)
+    org_uuid = accounting_data.get('organization')
+    user = User.objects.filter(username=username).first()
```
nemesisdesign, line 59: I think a try/except here using `get()` instead of filter would result more readable and with less indented code, we'll have to test the case in which no user corresponds to the radius accounting anyway because it can happen when using realms or when using radius checks

openwisp_radius/tasks.py
```diff
@@ -39,3 +49,38 @@ def delete_unverified_users(older_than_days=1, exclude_methods=''):
...
+                subject = _('New radius accounting session started')
+                context = {
+                    'user': user,
+                    'subject': subject,
+                    'call_to_action_url': one_time_login_url,
+                    'call_to_action_text': _('Manage Session'),
+                }
+                body_html = loader.render_to_string(
+                    'radius_accounting_start.html', context
+                )
```
nemesisdesign, line 81: use django's  strip_tags to prepare the textual message (`body_text`)

openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -0,0 +1,9 @@
+{% load i18n %}{% load l10n %}
+<div class="msg">
+<p>
+{% trans "New session has been started for your account with username:" %} {{ user.get_username }}
```
nemesisdesign, line 4: try to use:

``` 
<p>.....</p>
<p>......</p>
```
To help strip_tags to do its job without leaving too many whitespace.

nemesisdesign (changes requested):
@codesankalp great progress, see my comments below.
@pandafy your review here would be helpful!

openwisp_radius/tasks.py
codesankalp: The app is not loading if imported above.

openwisp_radius/tasks.py
codesankalp: Strip tag is already used for `body_text` in the send_mail function of openwisp_utils so if I pass the HTML then it strips the tag by default if the email type is text.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -681,24 +684,37 @@ def test_accounting_start_200(self):
...
         self._login_and_obtain_auth_token()
         data = self._prep_start_acct_data()
         data.update(username='tester')
         self.assertEqual(RadiusAccounting.objects.count(), 0)
-        response = self.client.post(
-            self._acct_url,
-            data=json.dumps(data),
-            content_type='application/json',
-        )
+        with catch_signal(radius_accounting_success) as handler:
```
pandafy, line 711: We should also assert that the celery task responsible for sending email is called.

codesankalp: I tried that by mocking it but it was failing for the sample app because the delay method was not present for the sample app.

pandafy: That is strange. Can you please add that code and push the changes. We'll find out the reason for failing build after that.

codesankalp: @pandafy I have added the tests.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -677,28 +680,45 @@ def test_accounting_no_token_403(self):
         )
 
     @freeze_time(START_DATE)
-    def test_accounting_start_200(self):
+    @mock.patch('openwisp_radius.receivers.send_login_email')
```
pandafy, line 683:
```suggestion
    @mock.patch('openwisp_radius.receivers.send_login_email.delay')
```

It is required to mock the exact method that is called.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -677,28 +680,45 @@ def test_accounting_no_token_403(self):
...
+    def test_accounting_start_200(self, send_login_email):
         self.assertEqual(RadiusAccounting.objects.count(), 0)
         ra = self._create_radius_accounting(**self._acct_initial_data)
         data = self._prep_start_acct_data()
-        response = self.post_json(data)
+        with catch_signal(radius_accounting_success) as handler:
+            response = self.post_json(data)
+        handler.assert_called_once()
+        send_login_email.delay.assert_called_once()
```
pandafy, line 691:
```suggestion
        send_login_email.assert_called_once()
```

pandafy (changes requested):

openwisp_radius/tests/test_api/test_freeradius_api.py
codesankalp: It still fails.

pandafy: In `openwisp_radius.apps`, `radius_accounting_success` is connected to `openwisp_radius.api.freeradius_views.AccountingView`

https://github.com/openwisp/openwisp-radius/blob/6facba5814022303c70ab88ed4a382979af5202f/openwisp_radius/apps.py#L69-L73

But the sample app inherits the base view and creates a new class: 

https://github.com/openwisp/openwisp-radius/blob/6facba5814022303c70ab88ed4a382979af5202f/tests/openwisp2/sample_radius/api/views.py#L32-L33

The sender of the signal is this new class for sample application. We need to connect this signal in apps.py of sample app.

openwisp_radius/apps.py
```diff
@@ -62,6 +64,13 @@ def connect_signals(self):
         RadiusToken = load_model('RadiusToken')
         RadiusAccounting = load_model('RadiusAccounting')
         User = get_user_model()
+        from openwisp_radius.api.freeradius_views import AccountingView
+
+        radius_accounting_success.connect(
+            radius_accounting_success_handler,
+            sender=AccountingView,
+            dispatch_uid='radius_accounting_success',
```
nemesisdesign, line 72: `radius_accounting_success` is too generic, we should use something more specific, eg: `send_email_on_new_accounting`

openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n %}{% load l10n %}
+<div class="msg">
+<p>{% trans "New session has been started for your account with username:" %} {{ user.get_username }}</p>
+<p>{% trans "Please click on the button below to manage your session:" %}</p>
```
nemesisdesign, line 4:
```suggestion
<p>{% trans "You can manage your account or terminate this session any time by clicking on the button below." %}</p>
```

nemesisdesign (changes requested):
Great progress @codesankalp! :+1: 

In order to test this more easily, can you please change the captive portal login and logout mock views (and then update their relative docs) to send HTTP requests to the accounting view to perform the accounting start and stop? Right now the record is created or modified directly with the ORM, but it doesn't trigger this feature.

See also my comments below.

setup.py
```diff
@@ -39,7 +39,10 @@
         'swapper~=1.1.0',
         # Needed for the new authentication backend in openwisp-users
         # TODO: remove when the new version of openwisp-users is released
-        'openwisp-users @ https://github.com/openwisp/openwisp-users/tarball/master',
+        (
+            'openwisp-users @ https://github.com/codesankalp/openwisp-users/'
+            'tarball/dev'
```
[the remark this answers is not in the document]
codesankalp: Actually there were two PRs (language field + passwordless authentication). So I have to merge those commits into a single branch.

tests/openwisp2/sample_radius/apps.py
```diff
@@ -1,7 +1,19 @@
...
+
+    def connect_signals(self):
+        from .api.views import AccountingView
+
+        radius_accounting_success.connect(
+            radius_accounting_success_handler,
+            sender=AccountingView,
+            dispatch_uid='radius_accounting_success',
+        )
+        return super().connect_signals()
```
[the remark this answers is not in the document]
codesankalp: To test the signal is calling the reciever I added the test, but in sample_app it was failing because the actual view method changes which calls the signal. So to make it pass I have to define it here as per [this comment](https://github.com/openwisp/openwisp-radius/pull/355#discussion_r754318693).

codesankalp:
> In order to test this more easily, can you please change the captive portal login and logout mock views (and then update their relative docs) to send HTTP requests to the accounting view to perform the accounting start and stop? Right now the record is created or modified directly with the ORM, but it doesn't trigger this feature.

I tried this but I was not able to change the request data object so that it can call the accounting view with updated request.data.

nemesisdesign:
> > In order to test this more easily, can you please change the captive portal login and logout mock views (and then update their relative docs) to send HTTP requests to the accounting view to perform the accounting start and stop? Right now the record is created or modified directly with the ORM, but it doesn't trigger this feature.
> 
> I tried this but I was not able to change the request data object so that it can call the accounting view with updated request.data.
> 
> ```python
> request.POST = request.POST.copy()
> request.POST = {'username': 'codesankalp} # data
> accounting(request) # the data is not being passed
> ```
> 
> I tried other methods also, but in all I cannot update this request object.

I'd send an HTTP request (using the requests library) to the AccountingView endpoint so that it would work very similarly to how it works in prod.

openwisp_radius/tasks.py
```diff
@@ -39,3 +51,41 @@ def delete_unverified_users(older_than_days=1, exclude_methods=''):
...
+    org_uuid = accounting_data.get('organization')
+    try:
+        user = User.objects.get(username=username)
+        organization = Organization.objects.get(id=org_uuid)
+        if organization.is_member(user):
+            from sesame.utils import get_query_string
+
+            org_radius_settings = organization.radius_settings
+            login_url = org_radius_settings.login_url
+            if login_url:
```
nemesisdesign, line 70: let's do:

```python
if not login_url:
    return
```

To avoid unnecessary indentation which makes this harder to read.

openwisp_radius/tasks.py
```diff
@@ -39,3 +51,41 @@ def delete_unverified_users(older_than_days=1, exclude_methods=''):
...
+@shared_task
+def send_login_email(accounting_data):
+    User = get_user_model()
+    Organization = swapper.load_model('openwisp_users', 'Organization')
+    username = accounting_data.get('username', None)
+    org_uuid = accounting_data.get('organization')
+    try:
+        user = User.objects.get(username=username)
+        organization = Organization.objects.get(id=org_uuid)
+        if organization.is_member(user):
```
nemesisdesign, line 65: same here:

```python
if not organization.is_member(user):
    return
```

openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -0,0 +1,5 @@
+{% load i18n %}{% load l10n %}
+<div class="msg">
+<p>{% trans "New session has been started for your account with username:" %} {{ user.get_username }}</p>
```
nemesisdesign, line 3: A new session has been started for your account.

openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -0,0 +1,5 @@
...
+<p>{% trans "You can manage your account or terminate this session any time by clicking on the button below." %}</p>
```
nemesisdesign, line 4: You can review your session to find out how much time and/or traffic has been used or you can terminate the session at any time by clicking on the button below.

nemesisdesign (changes requested):
It's taking shape!

See my comments below.

I suggested improvements regarding the email text and subject, please also add the translations for these.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -13,6 +14,7 @@
 from django.utils.timezone import now
 from freezegun import freeze_time
 
+from openwisp_radius.api.freeradius_views import AccountingView
```
pandafy, line 17: QA checks are failing due to this line

nemesisdesign (changes requested):
@codesankalp please fix the build when you can.

pandafy (approved):

tests/openwisp2/integrations/tests.py
```diff
@@ -60,7 +61,7 @@ def test_captive_portal_login_mock(self, *args, **kwargs):
             self.assertIsNone(ra.stop_time)
             self.assertEqual(ra.username, user.username)
 
-        return radius_token, ra
+        cache.delete(f'rt-{radius_token.user.username}')
```
codesankalp, line 64: So, the tests are failing due to cache of token. I had spent many hours in trying to resolve this :sweat_smile:.
Thanks for fixing.

pandafy: Yes, it took me couple of hours to understand what was happening.

nemesisdesign (changes requested):
Have you tested the translations too?
I tried switching language on OW login pages but I still get the email in English.

I also get this error when I try to open the sesame link in an incognito window:
Screenshot from 2022-01-07 12-34-24

codesankalp:
> Have you tested the translations too? I tried switching language on OW login pages but I still get the email in English.
> 

I will test again and try to fix this.

codesankalp:
I tried it also, and after changing 
https://github.com/openwisp/openwisp-radius/blob/1265c88b631b7384d88e6a03c953a6c9611e9257/tests/openwisp2/settings.py#L184
to `True` and after running `./tests/manage.py compilemessages` it worked.

Sesame login is also working:
demo2-2022-01-10_10 23

nemesisdesign (approved):
Great work @codesankalp @pandafy :+1: 
Finally we can merge.

PR #359, opened 2021-11-27
[Fix] Fixed Inconsistent migration history #358

### LINKED ISSUE #358: [bug] Fix django.db.migrations.exceptions.InconsistentMigrationHistory
Let's fix it with using https://github.com/openwisp/django-swappable-models/commit/7a636a8dc3567ebc029b76735eca74e9a232baa1 and https://github.com/openwisp/openwisp-users/commit/993f7ee64d389558cd3cb8dfcaf4a352de00afe9

codesankalp:
Closes #358


tests/openwisp2/sample_radius/migrations/0003_default_groups_and_permissions.py
```diff
@@ -25,7 +25,7 @@ class Migration(migrations.Migration):
     dependencies = [
         swapper.dependency('openwisp_radius', 'RadiusToken'),
         (model_app_label, '0002_initial_openwisp_app'),
-        (users_model_app_label, '0002_default_groups_and_permissions'),
+        (users_model_app_label, '0004_default_groups'),
```
nemesisdesign, line 28: I suggest changing this code to use `swapper.dependency`, since it basically does the same thing.

eg:
```python
swapper.dependency(
    *swapper.split(rad_model), version='0002_initial_openwisp_app'
)
```

And the same for users.

tests/openwisp2/sample_users/migrations/0004_default_groups.py
```diff
@@ -15,9 +15,7 @@ class Migration(migrations.Migration):
 
     org_model = swapper.get_model_name('openwisp_users', 'organization')
     model_app_label = swapper.split(org_model)[0]
-    dependencies = [
-        (model_app_label, '0001_initial'),
-    ]
+    dependencies = [(model_app_label, '0001_initial')]
```
nemesisdesign, line 18: same here.

nemesisdesign (changes requested):
Looks good, see my comments below.

setup.py
```diff
@@ -36,7 +36,7 @@
     zip_safe=False,
     install_requires=[
         'django>=3.0,<3.2',
-        'swapper~=1.2.0',
+        'swapper @ git+git://github.com/openwisp/django-swappable-models.git@master',
```
codesankalp, line 39: Should I remove it from here?

.github/workflows/ci.yml
```diff
@@ -44,6 +44,7 @@ jobs:
         run: |
           pip install -U "pip==20.2.4" wheel setuptools
           pip install ${{ matrix.django-version }}
+          pip install swapper@git+git://github.com/openwisp/django-swappable-models.git@master
```
codesankalp, line 47: I added this here because openwisp-users is installing swapper v1.2 and when installing openwisp-radius it is not updating the swapper.

nemesisdesign (approved):

PR #366, opened 2022-01-20
[fix] Fix url csv from admin portal

atb00ker:
I tested the API but the API was referenced from admin portal as well in `radiusbatch/<slug>/change/`, I realized this a bit after the PR was merged!

### DIFF
--- a/openwisp_radius/settings.py
+++ b/openwisp_radius/settings.py
@@ -131,7 +131,7 @@ except AssertionError:  # pragma: no cover
     )
 
 # Path of urls that need to be refered in migrations files.
-CSV_URL_PATH = 'radiusbatch/csv/'
+CSV_URL_PATH = 'api/v1/radiusbatch/csv/'
 CALLED_STATION_IDS = get_settings_value('CALLED_STATION_IDS', {})
 CONVERT_CALLED_STATION_ON_CREATE = get_settings_value(
     'CONVERT_CALLED_STATION_ON_CREATE', False
--- a/openwisp_radius/tests/test_admin.py
+++ b/openwisp_radius/tests/test_admin.py
@@ -660,6 +660,14 @@ class TestAdmin(
         self.assertEqual(OrganizationUser.objects.all().count(), 3)
         for u in OrganizationUser.objects.all():
             self.assertEqual(u.organization, RadiusBatch.objects.first().organization)
+        # Test CSV URL in change form
+        rad_batch = RadiusBatch.objects.first()
+        response = self.client.get(
+            reverse(f'admin:{self.app_label}_radiusbatch_change', args=[rad_batch.id])
+        )
+        self.assertContains(
+            response, reverse('radius:serve_private_file', args=[rad_batch.csvfile])
+        )
 
     def _create_multitenancy_test_env(
         self, usergroup=False, groupcheck=False, groupreply=False
--- a/openwisp_radius/urls.py
+++ b/openwisp_radius/urls.py
@@ -16,7 +16,7 @@ def get_urls(api_views=None, social_views=None, saml_views=None):
     """
     return [
         path('api/v1/', include(get_api_urls(api_views))),
-        path('api/v1/', include(get_private_store_urls())),
+        path('', include(get_private_store_urls())),
         path('radius/social-login/', include(get_social_urls(social_views))),
         path('radius/saml2/', include(get_saml_urls(saml_views))),
     ]


openwisp_radius/settings.py
```diff
@@ -131,7 +131,7 @@ def get_settings_value(option, default):
     )
 
 # Path of urls that need to be refered in migrations files.
-CSV_URL_PATH = 'radiusbatch/csv/'
+CSV_URL_PATH = 'api/v1/radiusbatch/csv/'
```
atb00ker, line 134: Added this here because `CSV_URL_PATH` is used in the API and the admin page and the url needs to be the same at both places.

Please note that the old url did not contain `api/v1`, but I have added it because all the other API endpoints have it, this is a breaking change, please let me know if I should remove it!

openwisp_radius/urls.py
```diff
@@ -16,7 +16,7 @@ def get_urls(api_views=None, social_views=None, saml_views=None):
     """
     return [
         path('api/v1/', include(get_api_urls(api_views))),
-        path('api/v1/', include(get_private_store_urls())),
+        path('', include(get_private_store_urls())),
```
nemesisdesign, line 19: not sure I understood this? Can you explain why this is being changed?

atb00ker: Okay, so `CSV_URL_PATH` variable is used for both the [API urls](https://github.com/openwisp/openwisp-radius/blob/6ea0fcd54b54d653678892c0d310412e064fb1cc/openwisp_radius/private_storage/urls.py#L12) and [admin dashboard urls](https://github.com/openwisp/openwisp-radius/blob/60a8e145da615fb45249d30520a6199b17df058c/openwisp_radius/base/models.py#L819).
If I add `/api/v1` here. The path in admin dashboard is `api/v1/<rest-of-the-path>` but the UI is pointing to `api/v1/api/v1/<rest-of-the-path>`.

pandafy: Can we use `reverse_lazy` here so that something like this never happens again? 

https://github.com/openwisp/openwisp-radius/blob/60a8e145da615fb45249d30520a6199b17df058c/openwisp_radius/base/models.py#L817-L820

pandafy: > Can we use `reverse_lazy` here so that something like this never happens again?

No, we cannot. I tried it locally.

nemesisdesign (changes requested):
@atb00ker @pandafy a test which fails without the fix is missing for this patch.

openwisp_radius/tests/test_admin.py
```diff
@@ -660,6 +660,14 @@ def test_radiusbatch_org_user(self):
         self.assertEqual(OrganizationUser.objects.all().count(), 3)
         for u in OrganizationUser.objects.all():
             self.assertEqual(u.organization, RadiusBatch.objects.first().organization)
+        # Test CSV URL in change form
+        rad_batch = RadiusBatch.objects.first()
+        response = self.client.get(
+            reverse(f'admin:{self.app_label}_radiusbatch_change', args=[rad_batch.id])
```
atb00ker, line 666: I am not sure if it will test what we want it to test, because we are using reverse, it will always resolve some path, if the path is expected or not is not tested.
In other words, I don't think this test will fail without the patch either.

About testing it, I think if we want to test it, we need to resolve the path and compare it with `CSV_URL_PATH`.

I'll check it out and try to do this approach in a couple of hours! :smile:

pandafy: The goal of this test is to ensure that the URL used in admin UI is correct, i.e. it does not return HTTP 404.
In the current master, it is not possible to download the uploaded CSV file from the admin.

atb00ker: Oh, if you've checked it works, then we are good, I just assumed based on my thought that we need to compare it with `CSV_URL_PATH`! 

Thanks! :smile:

nemesisdesign (approved):
Thank you @pandafy for explaining me the issue which I didn't understand initially and sending the failing test.

There's one thing which baffles me:

```
./manage.py show_urls | grep batch | grep api
/api/v1/radius/batch/	openwisp_radius.api.views.BatchView	radius:batch
/api/v1/radius/organization/<slug:slug>/batch/<uuid:pk>/pdf/	openwisp_radius.api.views.DownloadRadiusBatchPdfView	radius:download_rad_batch_pdf
/api/v1/radiusbatch/csv/<path:csvfile>	openwisp_radius.private_storage.views.RadiusBatchCsvDownloadView	radius:serve_private_file
```

Shouldn't this CSV url be as follows?

`/api/v1/radius/organization/<slug:slug>/batch/<uuid:pk>/csv/`.

We have an inconsistency here, I will create a new issue for this, another problem is that this URL does not show up in the API docs, so it would be cool to find a way to show it there too.

Issue #365, opened 2022-01-20
[bug] invalid username allowed


atb00ker:
Create a batch of users (`/radiusbatch/add/`) using prefix method, and set the name and prefix as `Test#1` and save the batch.
Notice that the username `Test#1` is not allowed hence the request fails.

We should add a check in the input boxes to ensure only valid characters for username are allowed.

bsande6:
@atb00ker Hey, I am new to open source and would like to contribute. Can I give this issue a shot?

atb00ker:
@bsande6 feel free to start with it, permission not required! :smile:

bsande6:
@atb00ker Great! May I be assigned to this issue?

atb00ker:
Hi @bsande6 please start working on it, no assignment, permission or any process such is required to contribute.
When you create the pull request, please tag me for review! :smile:

bsande6:
@atb00ker I have followed the setup steps and have been able to launch the server on localhost however any changes I make in the code are not being displayed on the localhost. Do you have any ideas on where I have gone wrong?

atb00ker:
@bsande6 
Please ask questions in the development channel : https://gitter.im/openwisp/development
From the information provided, I have no clue.
Please add more details in the gitter channel about the change that you are making and it's not reflecting! 😄

PR #371, opened 2022-01-28
[change] Adds check for valid prefix in batch creation #365

bsande6:
@atb00ker This commit adds a check such that the prefix in a batch user creation only accepts letters, numbers and the special characters @/./+/-/_

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -901,7 +901,8 @@ class AbstractRadiusBatch(OrgMixin, TimeStampedEditableModel):
                     raise ValidationError(
                         {
                             'prefix': _(
-                                'This value may contain only letters, numbers, and @/./+/-/_ characters.'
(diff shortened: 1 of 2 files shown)


atb00ker (commented):
@devkapilbansal It's required because currently any character is allowed in the radius batch user creation form, but users can have only limited characters.

@bsande6 please:
1. Link the following the pr description and commit: https://github.com/openwisp/openwisp-radius/issues/365
2. Add a test to ensure that if an illegal character is entered, the form rejects the input, please checkout the existing tests for help: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/tests/test_batch_add_users.py
3. Tests are failing, please check them as well! 

If you need further information, please reach out on the gitter development channel!

atb00ker (approved):
Thanks for adding the tests! 👍🏻

PR #372, opened 2022-01-30
[docs] Fix internal references #339

### LINKED ISSUE #339: [docs] Fix internal references
In the docs we have quite a few internal references which point to HTML pages, but that doesn't work well when the documentation is converted to non HTML formats (eg: PDF, ePub). We should change those references to use the [ReStructuredText format for internal references](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#use-a-custom-anchor), eg:

```
:ref:`configure-rest-module`
```

Aryamanz29:
- Fixed various `internal references` in docs

Closes #339

Screenshot from 2022-01-31 02-19-01


Aryamanz29:
Changed commit message.

nemesisdesign (approved):

PR #374, opened 2022-02-01
[change!] Allowed username and phone_number in password reset API

pandafy:
**Backward incompatible change**: The password reset API endpoint now
accepts "input" parameter instead of "email".


nemesisdesign (changes requested):
`swagger_auto_schema` needs to be updated, `The email field is required` > `The input field is required`.

Let's update also the docstring of the post method, which is shown in the API docs, to indicate that input can be an email, an username or a phone number (if mobile phone verification is in use).

nemesisdesign (changes requested):
Something I had forgot to specify (sorry for that) is that we need to make sure we can recognize phone numbers also if the user doesn't enter the international prefix (which is the 99% of cases), we can do so by leveraging the get_user method of the openwisp-users authentication backend which already implements this: https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/backends.py#L25-L31.

Please add a failing test for this case and then use the users authentication backend to fix it, see also how we already use it in this module:
https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/api/freeradius_views.py#L46.

nemesisdesign (approved):
:+1:

PR #381, opened 2022-02-08
[change] Updated accounting start email to indicate duration #379

### LINKED ISSUE #379: [change] Update accounting start email to indicate duration
The email message defined in the following template does not mention the validity of the the link which is configurable by `SESAME_MAX_AGE` setting. 

https://github.com/openwisp/openwisp-radius/blob/81cefafa44ef13b2756f2d2598839f03d5c539fc/openwisp_radius/templates/radius_accounting_start.html#L1-L5

The default value of `SESAME_MAX_AGE` is `None`, i.e. valid indefinitely. Therefore, we should only show the additional text mentioning validity of the link only if `SESAME_MAX_AGE` is not `None`. 

The value of `SESAME_MAX_AGE` is in seconds. It should be converted into more readable format, e.g. `This link is valid for **1 hour**`.

Aryamanz29:
Updated `radius_accounting_start` email to indicate duration of validity of link
only if `SESAME_MAX_AGE` is not `None`. Closes #379


openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -2,4 +2,5 @@
 <div class="msg">
 <p>{% trans "A new session has been started for your account:" %} {{ user.get_username }}.</p>
 <p>{% trans "You can review your session to find out how much time and/or traffic has been used or you can terminate the session at any time by clicking on the button below." %}</p>
+{% if sesame_max_age %}<p>{% trans "NOTE: This link is valid only for" %} {{ sesame_max_age }} hrs</p> {% endif %}
```
nemesisdesign, line 5: please use this form:
https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#blocktranslate-template-tag

Different languages may place the `sesame_max_age` value before or after specific words, languages of our planet are very different from one another, we can't assume the value will be always displayed at the end of the phrase.

nemesisdesign (changes requested):

openwisp_radius/templates/radius_accounting_start.html
Aryamanz29: > please use this form: https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#blocktranslate-template-tag
> 
> Different languages may place the `sesame_max_age` value before or after specific words, languages of our planet are very different from one another, we can't assume the value will be always displayed at the end of the phrase.

You are right, I'll read docs and fix this 👍

openwisp_radius/tasks.py
```diff
@@ -99,5 +101,9 @@ def send_login_email(accounting_data):
             'call_to_action_url': one_time_login_url,
             'call_to_action_text': _('Manage Session'),
         }
+        if hasattr(settings, 'SESAME_MAX_AGE'):
+            context.update(
+                {'sesame_max_age': str(timedelta(seconds=settings.SESAME_MAX_AGE))}
```
codesankalp, line 106: Instead of passing in this format you can use [`timesince`](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#timesince) django template tag which is more readable than this.

codesankalp (changes requested):

openwisp_radius/tasks.py
Aryamanz29: Yes but that filter works with date instance, Here we only getting `SESAME_MAX_VALUE` as integer 🤔

Aryamanz29: > Yes but that filter works with date instance, Here we only getting `SESAME_MAX_VALUE` as integer thinking

This works :point_down: 

```python
# SESAME_MAX_AGE =  60 * 60
# SESAME_MAX_AGE =  60 * 60 * 60
sesame_max_age = datetime.datetime.utcnow() - datetime.timedelta(seconds=settings.SESAME_MAX_AGE)
{{sesame_max_age | timesince}} # In templates
```
Screenshot from 2022-02-09 12-44-59

Screenshot from 2022-02-09 12-45-17

codesankalp: Another solution is `naturaltime` but it depends on `django.contrib.humanize`

openwisp_radius/tasks.py
```diff
@@ -99,5 +101,12 @@ def send_login_email(accounting_data):
             'call_to_action_url': one_time_login_url,
             'call_to_action_text': _('Manage Session'),
         }
+        if hasattr(settings, 'SESAME_MAX_AGE'):
+            context.update(
+                {
+                    'sesame_max_age': timezone.localtime()
```
Aryamanz29, line 107: Changed `datetime.utcnow()` to `timezone.localtime()` because we're using `TIME_ZONE = 'America/Asuncion'` in settings, Which is not in UTC and time zone is user dependent.

pandafy:
```suggestion
                    'sesame_max_age': timezone.now()
```

openwisp_radius/tasks.py
```diff
@@ -99,5 +101,12 @@ def send_login_email(accounting_data):
...
+                    - timedelta(seconds=settings.SESAME_MAX_AGE)
```
pandafy, line 108:
```suggestion
                    + timedelta(seconds=settings.SESAME_MAX_AGE)
```

The expiration time will be in future. We should add the time here.

openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -2,4 +2,5 @@
 <div class="msg">
 <p>{% trans "A new session has been started for your account:" %} {{ user.get_username }}.</p>
 <p>{% trans "You can review your session to find out how much time and/or traffic has been used or you can terminate the session at any time by clicking on the button below." %}</p>
+{% if sesame_max_age %}<p>{% blocktrans with sesame_max_age|timesince as sesame_max_age %}NOTE: This link is valid only for {{sesame_max_age}}{% endblocktrans %}.</p>{% endif %}
```
pandafy, line 5: I think using [`naturaltime`](https://docs.djangoproject.com/en/4.0/ref/contrib/humanize/#naturaltime) will be better here. Since, it renders time as `an hour from now.`

openwisp_radius/tests/test_tasks.py
```diff
@@ -173,6 +174,10 @@ def test_send_login_email(self, translation_activate, logger):
                 ' and/or traffic has been used or you can terminate the session',
                 ' '.join(email.alternatives[0][0].split()),
             )
+            self.assertIn(
+                'NOTE: This link is valid only for 2 hours',
+                ' '.join(email.alternatives[0][0].split()),
+            )
             translation_activate.assert_called_with(user.language)
```
pandafy, line 181: We should add a separate subtest for sesame token. In the current subtest, we should check that the expiration text is not present if `SESAME_MAX_AGE` variable is not set.

pandafy (changes requested):
I have requested some minor changes below. After that, it will be good to merge.

openwisp_radius/templates/radius_accounting_start.html
```diff
@@ -1,5 +1,7 @@
 {% load i18n %}{% load l10n %}
+{% load humanize %}
 <div class="msg">
 <p>{% trans "A new session has been started for your account:" %} {{ user.get_username }}.</p>
 <p>{% trans "You can review your session to find out how much time and/or traffic has been used or you can terminate the session at any time by clicking on the button below." %}</p>
+{% if sesame_max_age %}<p>{% blocktrans with sesame_max_age|naturaltime as sesame_max_age %}NOTE: This link is valid only for {{sesame_max_age}}{% endblocktrans %}.</p>{% endif %}
```
pandafy, line 6: @Aryamanz29 did you tested email translation manually? I tested it in my local environment but it didn't work. Can you double check this? 

```html
<div class="msg">
<p>È iniziata una nuova sessione per il tuo account: user_pandafy.</p>
<p>Puoi monitorare il consumo di tempo e traffico della tua sessione o terminare la sessione in qualsiasi momento cliccando sul collegamento seguente.</p>
<p>NOTE: This link is valid only for tra 59 minuti.</p>
</div>
</div>
```

pandafy (changes requested):
Good work @Aryamanz29! We are almost there with this PR.

openwisp_radius/templates/radius_accounting_start.html
Aryamanz29: > @Aryamanz29 did you tested email translation manually? I tested it in my local environment but it didn't work. Can you double check this?
> 
> ```
> <div class="msg">
> <p>È iniziata una nuova sessione per il tuo account: user_pandafy.</p>
> <p>Puoi monitorare il consumo di tempo e traffico della tua sessione o terminare la sessione in qualsiasi momento cliccando sul collegamento seguente.</p>
> <p>NOTE: This link is valid only for tra 59 minuti.</p>
> </div>
> </div>
> ```

@pandafy 
I've used the below setting while testing 🤔
```python
    with override_settings(SESAME_MAX_AGE=2*60*60, LANGUAGE_CODE='it'):
        with self.subTest(
                'it should check expiration text is present when SESAME_MAX_AGE is set'
            ):
            # Test code
```
It is showing translated text in my console :point_down: 
```html
<div class="msg">
<p>È iniziata una nuova sessione per il tuo account: tester.</p> 
<p>Puoi monitorare il consumo di tempo e traffico della tua sessione o terminare la sessione in qualsiasi momento cliccando sul collegamento seguente.</p> 
<p>NOTA: Questo link è valido solo per tra un&#x27;ora.</p> </div> </div> <div class="call-to-action"> <a href="https://wifi.openwisp.org/default/login/?sesame=e9Gy-IW1SF6EmS8wvv6wHlwHeqwgNu1f78M" class="btn">Gestisci sessione</a> 
</div>
```

openwisp_radius/tests/test_tasks.py
```diff
@@ -173,6 +172,25 @@ def test_send_login_email(self, translation_activate, logger):
...
+                ' '.join(email.alternatives[0][0].split()),
+            )
+            translation_activate.assert_called_with(user.language)
+
+        translation_activate.reset_mock()
+
+        with self.subTest(
+            'it should check expiration text is present when SESAME_MAX_AGE is set'
+        ):
+            setattr(settings, 'SESAME_MAX_AGE', 2 * 60 * 60)
```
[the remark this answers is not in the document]
Aryamanz29: > I would mock this settings as following instead of using `setattr`.
> 
> ```
> mock.patch('sesame.settings.MAX_AGE')
> ```
> 
> If it is not possible to mock this setting, then let's unset this after completion of the test. Otherwise, it could affect other tests as well.

Can we also do something like this? 🤔

```python
from django.test.utils import override_settings

    with override_settings(SESAME_MAX_AGE=2*60*60):
        with self.subTest(
                'it should check expiration text is present when SESAME_MAX_AGE is set'
            ):
            # Test code
```
https://docs.djangoproject.com/en/4.0/topics/testing/tools/#overriding-settings

openwisp_radius/templates/radius_accounting_start.html
pandafy: @Aryamanz29 the translation should work regardless of `LANGUAGE_CODE` setting. We translate the email context according to user preference (according to `accept-language` header of the user's requests.

Aryamanz29: Tested with `wifi-login-pages` :point_down: 
image

nemesisdesign (commented):
Looks good, I need more time to test it, we'll have to update the ansible and docker projects to add `'django.contrib.humanize',` in order for this to work.

openwisp_radius/locale/de/LC_MESSAGES/django.po
```diff
@@ -136,6 +136,10 @@ msgstr "Eine neue Session für Ihren Account wird gestartet:"
 msgid "You can review your session to find out how much time and/or traffic has been used or you can terminate the session at any time by clicking on the button below."
 msgstr "Um die Dauer und den Datenverbrauch Ihrer Session zu überwachen oder um Ihre Session in jedem Moment zu beenden, klicken Sie auf den folgenden Link."
 
+#: openwisp_radius/templates/radius_accounting_start.html:5
+msgid "NOTE: This link is valid only for %(sesame_max_age)s"
```
nemesisdesign, line 140: `NOTE: This ` > `Note: this`, I am handling this change.

nemesisdesign (approved):

PR #382, opened 2022-02-09
[change] Move password reset url to db #344

### LINKED ISSUE #344: [change] Move password reset URL to DB
As https://github.com/openwisp/openwisp-radius/issues/343, we shall move the password reset URL to the DB so it's editable by organization administrators, for consistency and ease of use. Docs and test shall be updated.

This shall be done once #343 is completed.

totallynotvaishnav:
closes #344


openwisp_radius/api/utils.py
```diff
@@ -63,3 +63,15 @@ def is_registration_enabled(org):
         raise APIException(
             _('Could not complete operation because of an internal misconfiguration')
         )
+
+
+def get_password_reset_url(org):
```
nemesisdesign, line 68: it looks to me that placing this function here does not make sense.

openwisp_radius/api/utils.py
```diff
@@ -63,3 +63,15 @@ def is_registration_enabled(org):
...
+    try:
+        return org.radius_settings.password_reset_url
+    except ObjectDoesNotExist:
+        logger.exception(
+            f'Got exception while accessing radius_settings for {org.name}'
+        )
+        raise APIException(
```
nemesisdesign, line 75: raising APIException outside of a views.py file also doesn't make a lot of sense.

openwisp_radius/api/views.py
```diff
@@ -487,18 +492,19 @@ def get_serializer_context(self):
...
-            organization_pk, organization_slug = None, None  # pragma: no cover
+            organization_slug = None  # pragma: no cover
         else:
-            organization_pk = self.organization.pk
             organization_slug = self.organization.slug
-        password_reset_url = password_reset_urls.get(str(organization_pk), default_url)
-        password_reset_url = password_reset_url.format(
-            organization=organization_slug, uid=uid, token=token, site=domain
-        )
+        password_reset_url = get_password_reset_url(self.organization)
```
nemesisdesign, line 500: I would make `get_password_reset_url` a method of this class or a function in this file. I think it makes sense as a method of this class.

PR #389, opened 2022-02-24
[feature] Added organization level setting to configure registration methods #308

### LINKED ISSUE #308: [change] Allow enabling/disabling registration methods (SAML, Social login, SMS)
We should add a way to allow users to specify if they want to enable specific registration/login methods like SAML, social login, SMS verification, etc. for their organization.

We shall also check if a method it's enabled before allowing users to log in / sign up with it or otherwise return 403.

pandafy:
Added settings for enabling or disabling SAML and social login
at organization level.

Closes #308


openwisp_radius/base/models.py
```diff
@@ -1199,6 +1217,21 @@ def get_sms_verification(self):
...
+        if self.saml_registration_enabled is None:
+            return app_settings.SAML_LOGIN_ENABLED
+        return self.saml_registration_enabled
+
+    def get_social_registration_enabled(self):
+        print(
+            'social',
+            self.social_registration_enabled,
+            app_settings.SOCIAL_LOGIN_ENABLED,
+        )
```
nemesisdesign, line 1230: this seems like a leftover

openwisp_radius/saml/utils.py
```diff
@@ -1,8 +1,26 @@
...
     return parsed_url.path
+
+
+def is_saml_authentication_enabled(org):
+    try:
+        return org.radius_settings.get_saml_registration_enabled()
+    except ObjectDoesNotExist:
+        logger.exception(
+            f'Got exception while accessing radius_settings for {org.name}'
+        )
```
nemesisdesign, line 23: I am not sure this is needed here, the exception will fire anyway below

openwisp_radius/social/utils.py
```diff
@@ -0,0 +1,18 @@
...
+logger = logging.getLogger()
+
+
+def is_social_authentication_enabled(org):
+    try:
+        return org.radius_settings.get_social_registration_enabled()
+    except ObjectDoesNotExist:
+        logger.exception(
+            f'Got exception while accessing radius_settings for {org.name}'
+        )
```
nemesisdesign, line 15: same here

nemesisdesign (changes requested):
Great work @pandafy! 

I see that basically these two features are enabled by default but I think it's not the right thing to do, can you proceed to change the default settings so that they're both disabled? Please also make the default settings configurable and documented in the settings page as it's done for [OPENWISP_RADIUS_REGISTRATION_API_ENABLED](https://openwisp-radius.readthedocs.io/en/latest/user/settings.html#openwisp-radius-registration-api-enabled).

I think the docs have to be updated:

- saml page needs to refer (eg: link) how to enable this feature
- social login page needs to (eg: link) how to enable this feature

nemesisdesign (approved):

PR #390, opened 2022-03-01
[fix] Fixed handling of interim-updates for closed sessions #322

### LINKED ISSUE #322: [bug] RadiusAccounting: accounting on closed session should re-open it
If a session is flagged as closed but interim updates keep coming, it means the session was not really closed (which in itself is another issue which users will have to deal depending on their specific installation), on the openwisp radius side we should ensure the accounting request does not fail.

**Here's how to replicate it:**

- create a radius session with its own unique_id, flag it as terminated
- send an interim update request, eg:

```
curl -X POST "http://localhost:8000/api/v1/freeradius/accounting/" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{\"status_type\": \"Interim-Update\", \"session_id\": \"405448b5b78861da\", \"unique_id\": \"5b24f4f88545927940cf1f12e9dc70a2\", \"username\": \"username-here\", \"realm\": \"NULL\", \"nas_ip_address\": \"127.0.0.1\", \"nas_port_id\": \"2128\", \"nas_port_type\": \"Ethernet\", \"session_time\": \"3127\", \"authentication\": \"RADIUS\", \"input_octets\": \"2506645\", \"output_octets\": \"33801977\", \"called_station_id\": \"00:00:00:00:00:11:gw\", \"calling_station_id\": \"00:00:00:00:00:22\", \"terminate_cause\": \"\", \"service_type\": \"Login-User\", \"framed_protocol\": \"\", \"framed_ip_address\": \"172.28.121.95\"}"
```

**Expected result:**

200 ok, request input/output packets and time shall be updated, it shall be flagged as still open.

**Current Result:**

```
{"unique_id":["accounting with this accounting unique ID already exists."]}
```

It seems it's trying to create a new radius accounting record instead of updating the existing one.

pandafy:
Added handling of "Interim-Updates" for RadiusAccounting sessions
that are closed by OpenWISP when user logs into another organization.

Fixes #322

### DIFF
--- a/openwisp_radius/api/freeradius_views.py
+++ b/openwisp_radius/api/freeradius_views.py
@@ -436,22 +436,8 @@ class AccountingView(ListCreateAPIView):
             try:
                 serializer.is_valid(raise_exception=True)
             except ValidationError as error:
-                # Handles "Interim-Updates" for RadiusAccounting sessions
-                # that are closed by OpenWISP when user logs into
-                # another organization.
-                unique_id_errors = error.detail.get('unique_id', [])
-                if len(unique_id_errors) == 1:
-                    error_detail = unique_id_errors.pop()
-                    if (
-                        str(error_detail)
-                        == 'accounting with this accounting unique ID already exists.'
-                        and error_detail.code == 'unique'
-                    ):
-                        rad = RadiusAccounting.objects.only(
-                            'unique_id', 'organization_id'
-                        ).get(unique_id=data.get('unique_id'))
-                        if rad.organization_id != request.auth:
-                            return Response(None)
+                if self._is_interim_update_corner_case(error, data):
+                    return Response(None)
                 raise error
             acct_data = self._data_to_acct_model(serializer.validated_data.copy())
             serializer.create(acct_data)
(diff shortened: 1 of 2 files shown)


openwisp_radius/api/freeradius_views.py
```diff
@@ -428,7 +433,26 @@ def post(self, request, *args, **kwargs):
...
+                    if (
+                        str(error_detail)
+                        == 'accounting with this accounting unique ID already exists.'
+                        and error_detail.code == 'unique'
+                    ):
+                        rad = RadiusAccounting.objects.only(
+                            'unique_id', 'organization_id'
+                        ).get(unique_id=data.get('unique_id'))
+                        if rad.organization_id != request.auth:
+                            return Response(None)
```
nemesisdesign, line 454: can you please make this whole block a private method of the view called `_is_interim_update_corner_case()` and then just do:

```python
if self._is_interim_update_corner_case(error):
    return Response(None)
raise error
```

This way we keep the `post` method leaner, it's always the best thing to maintain this kind of complex logic, especially for corner cases, separate.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1043,6 +1043,48 @@ def test_accounting_400_validation_error(self, mocked_logger):
...
+
+        # Send Interim-Update request for "default" organization
+        # with Authorization header containing "org2" credentials.
+        response = self.client.post(
+            self._acct_url,
+            data=json.dumps(data),
+            HTTP_AUTHORIZATION=f'Bearer {org2.pk} {org2.radius_settings.token}',
+            content_type='application/json',
+        )
+        self.assertEqual(response.status_code, 200)
```
nemesisdesign, line 1086: please assert the response is `None` as well.

nemesisdesign (changes requested):

openwisp_radius/api/freeradius_views.py
```diff
@@ -442,6 +452,25 @@ def post(self, request, *args, **kwargs):
             self.send_radius_accounting_signal(serializer.validated_data)
             return Response(None)
 
+    def _is_interim_update_corner_case(self, error, data):
+        # Handles "Interim-Updates" for RadiusAccounting sessions
+        # that are closed by OpenWISP when user logs into
+        # another organization.
```
nemesisdesign, line 458: please convert this into a docstring

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #399, opened 2022-04-01
[Fix] Flag user as verified in batch user creation #397

### LINKED ISSUE #397: [bug] Batch user creation does not flag users as verified
Batch user creation does not flag users as verified, this will most probably not work in organizations which require identity verification, the users created won't be able to log in.

Here's what I think we have to do:

- add a functional test in [test_batch_add_users.py](https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/tests/test_batch_add_users.py) with an organization which requires identity verification, generate 1 user and then try the authorization method with that user expecting an auth type accept response (`{'control:Auth-Type': 'Accept'}` and 200 HTTP status code), I am pretty sure this test should fail
- change the code which generates the users so that a RegisteredUser model is also created automatically for each user, using the "Manually created" option, then, if the organization requires identity verification, also flag the verified field as true
- reflect these changes in the documentation
- override the template of the form used in the admin for this model to add an help text at the beginning of the form like this one:

https://github.com/openwisp/openwisp-utils/blob/7a86dffbcc72c72392a283a3db3fca9a0e021ce7/openwisp_utils/admin_theme/templates/admin/edit_inline/help_text_stacked.html#L29-L38

The help text should say: "Users imported or generated through this form will be flagged as verified if the organization requires identity verification, otherwise the generated users would not be able to log in. If this organization requires identity verification, make sure the identity of the users is verified before importing or before giving out the credentials.".

codesankalp:
Fixes #397

### DIFF
--- a/openwisp_radius/admin.py
+++ b/openwisp_radius/admin.py
@@ -386,6 +386,7 @@ class RadiusPostAuthAdmin(OrganizationFirstMixin, BasePostAuth):
 @admin.register(RadiusBatch)
 class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
     change_form_template = 'openwisp-radius/admin/rad_batch_users_change_form.html'
+    add_form_template = 'openwisp-radius/admin/rad_batch_users_add_form.html'
     list_display = [
         'name',
         'organization',
@@ -412,6 +413,20 @@ class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
     ]
     search_fields = ['name']
     form = RadiusBatchForm
+    help_text = {
+        'text': _(
+            'Users imported or generated through this form will '
+            'be flagged as verified if the organization requires '
+            'identity verification, otherwise the generated users '
+            'would not be able to log in. If this organization '
+            'requires identity verification, make sure the identity '
+            'of the users is verified before importing or before '
+            'giving out the credentials.'
+        ),
+        'documentation_url': (
+            'https://openwisp-radius.readthedocs.io/en/latest/user/importing_users.html'
+        ),
+    }
 
     class Media:
         js = [
@@ -468,6 +483,11 @@ class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
             form_url,
             extra_context=extra_context,
         )
+
+    def add_view(self, request, form_url='', extra_context=None):
+        extra_context = extra_context or {}
+        extra_context['help_text'] = self.help_text
+        return super().add_view(request, form_url, extra_context)
 
     def get_actions(self, request):
         actions = super().get_actions(request)
--- a/openwisp_radius/admin.py
+++ b/openwisp_radius/admin.py
@@ -415,12 +415,11 @@ class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
     form = RadiusBatchForm
     help_text = {
         'text': _(
-            'Users imported or generated through this form will '
-            'be flagged as verified if the organization requires '
-            'identity verification, otherwise the generated users '
-            'would not be able to log in. If this organization '
-            'requires identity verification, make sure the identity '
-            'of the users is verified before importing or before '
+            'Users imported or generated through this form will be flagged '
+            'as verified if the organization requires identity verification, '
+            'otherwise the generated users would not be able to log in. '
+            'If this organization requires identity verification, make sure '
+            'the identity of the users is verified before before '
             'giving out the credentials.'
         ),
         'documentation_url': (
@@ -433,7 +432,12 @@ class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
             'admin/js/jquery.init.js',
             'openwisp-radius/js/strategy-switcher.js',
         ]
-        css = {'all': ('openwisp-radius/css/radiusbatch.css',)}
+        css = {
+            'all': (
+                'openwisp-radius/css/radiusbatch.css',
+                'admin/css/help-text-stacked.css',
+            )
+        }
 
     def number_of_users(self, obj):
         return obj.users.count()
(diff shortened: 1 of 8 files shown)


openwisp_radius/base/models.py
```diff
@@ -1007,7 +1007,13 @@ def get_or_create_user(self, row, users_list, password_length):
 
     def save_user(self, user):
         OrganizationUser = swapper.load_model('openwisp_users', 'OrganizationUser')
+        RegisteredUser = swapper.load_model('openwisp_radius', 'RegisteredUser')
         user.save()
+        registered_user = RegisteredUser(user=user)
+        registered_user.save()
```
nemesisdesign, line 1013: I suggest to move this line just before the `self.users.add(user)`, so you won't need the other save call below.

openwisp_radius/base/models.py
```diff
@@ -1007,7 +1007,13 @@ def get_or_create_user(self, row, users_list, password_length):
...
+        if self.organization.radius_settings.needs_identity_verification:
+            registered_user.is_verified = True
+            registered_user.save(update_fields=['is_verified'])
```
nemesisdesign, line 1016: this shouldn't be needed if you follow my suggestion above

openwisp_radius/admin.py
```diff
@@ -469,6 +482,11 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
             extra_context=extra_context,
         )
 
+    def add_view(self, request, form_url='', extra_context=None):
+        extra_context = extra_context or {}
+        extra_context['help_text'] = self.help_text
+        return super().add_view(request, form_url, extra_context)
```
codesankalp, line 488: This is going into the template for add_view but it is not rendering:

image

image

openwisp_radius/admin.py
```diff
@@ -412,6 +413,20 @@ class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
...
     form = RadiusBatchForm
+    help_text = {
+        'text': _(
+            'Users imported or generated through this form will '
+            'be flagged as verified if the organization requires '
+            'identity verification, otherwise the generated users '
+            'would not be able to log in. If this organization '
+            'requires identity verification, make sure the identity '
+            'of the users is verified before importing or before '
+            'giving out the credentials.'
```
nemesisdesign, line 424: I think here and above (in the docs) we can remove "before importing or"

openwisp_radius/base/models.py
```diff
@@ -1007,7 +1007,12 @@ def get_or_create_user(self, row, users_list, password_length):
 
     def save_user(self, user):
         OrganizationUser = swapper.load_model('openwisp_users', 'OrganizationUser')
+        RegisteredUser = swapper.load_model('openwisp_radius', 'RegisteredUser')
         user.save()
+        registered_user = RegisteredUser(user=user)
+        if self.organization.radius_settings.needs_identity_verification:
+            registered_user.is_verified = True
```
nemesisdesign, line 1014: in the issue description I asked to use the "Manually created" option, see https://github.com/openwisp/openwisp-radius/blob/2235fe185bfb8e13118a3f9edc4e28126882ddc3/openwisp_radius/registration.py#L8-L13.

openwisp_radius/admin.py
```diff
@@ -386,6 +386,7 @@ class RadiusPostAuthAdmin(OrganizationFirstMixin, BasePostAuth):
 @admin.register(RadiusBatch)
 class RadiusBatchAdmin(MultitenantAdminMixin, TimeStampedEditableAdmin):
     change_form_template = 'openwisp-radius/admin/rad_batch_users_change_form.html'
+    add_form_template = 'openwisp-radius/admin/rad_batch_users_add_form.html'
```
nemesisdesign, line 389: looks like this CSS is missing: https://github.com/openwisp/openwisp-utils/blob/master/openwisp_utils/admin.py#L163.

See how it looks: https://github.com/openwisp/openwisp-utils#openwisp-utils-admin-helptextstackedinline.

openwisp_radius/tests/test_tasks.py
```diff
@@ -103,7 +103,9 @@ def test_delete_unverified_users(self):
         management.call_command('batch_add_users', **options)
         User.objects.update(date_joined=now() - timedelta(days=3))
         for user in User.objects.all():
-            RegisteredUser.objects.create(user=user, method='email', is_verified=False)
+            user.registered_user.is_verified = False
+            user.registered_user.method = 'email'
+            user.registered_user.save(update_fields=['is_verified', 'method'])
```
nemesisdesign, line 108: the same thing has to be done in `openwisp_radius.tests.test_commands.TestCommands.test_unverified_users_command`

nemesisdesign (changes requested):
Good progress @codesankalp, see my comments below.

openwisp_radius/tests/test_batch_add_users.py
```diff
@@ -108,3 +111,31 @@ def test_valid_username_special_char(self):
...
+        )
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(
+            response.json(),
+            {
+                'control:Auth-Type': 'Accept',
+                'Session-Timeout': 10800,
+                'ChilliSpot-Max-Total-Octets': 3000000000,
+            },
+        )
```
nemesisdesign, line 141: Let's also add an assertion which ensures the users have a RegisteredUser relation with is_verified=True and the manually created method please.

nemesisdesign (changes requested):

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_add_form.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'admin/change_form.html' %}
+{% load i18n %}
+{% block field_sets %}
+{% if help_text %}
+<div class="ow-help-text">
+    <p>{{ help_text.text }}
```
codesankalp, line 6: I removed the image tag from here because it is not looking good with long text.

image

nemesisdesign (approved):
Thanks :+1:

PR #400, opened 2022-04-01
[feature] Added setting to avoid updating username from SAML #398

### LINKED ISSUE #398: [bug/regression] Signing-in with SAML changes user's username
If a user signs up with email or phone number and then tries to log-in with SAML, then the user's username will get updated if the email address of the user is same in both IdP and OpenWISP. This is because, `djangosaml2` uses email addresses 

We should prevent username field to get changed in this scenario.

pandafy:
Fixes #398


openwisp_radius/saml/backends.py
```diff
@@ -0,0 +1,28 @@
...
+                        attribute_mapping.pop(key, None)
+                    else:
+                        attribute_mapping[key] = []
+                        for attr in value:
+                            if attr != 'username':
+                                attribute_mapping[key].append(attr)
+
+            except ObjectDoesNotExist:
+                pass
+        return super()._update_user(user, attributes, attribute_mapping, force_save)
```
nemesisdesign, line 28: I think something is in how this works.

I believe if the user is created with SPID, we shall always update the username no matter what.
If the user was not created with SPID, we shall not update their username unless the setting is set to true.
The docs should reflect this difference.

I'd rename the setting something like `SAML_UPDATES_PRE_EXISTING_USERNAME` to make it more explicit.

docs/source/user/saml.rst
```diff
@@ -142,3 +142,25 @@ Settings
...
+a database query is made to check whether such a user already exists.
+This lookup is done using the value of ``SAML_DJANGO_USER_MAIN_ATTRIBUTE`` setting.
+If a match is found, the details of the user are updated with the
+information received from SAML Identity Provider.
+
+If a user (who has registered on OpenWISP with a different method from SAML)
+logs into OpenWISP with SAML, then the details of this user will also get updated.
+This operation could also change the username of the user which
+will render old credentials useless. To avoid updating the username in such scenarios,
+set :ref:`OPENWISP_RADIUS_SAML_UPDATES_USERNAME <openwisp_radius_saml_updates_username>` to ``False``.
```
nemesisdesign, line 166: Here this part is written as if the user has to do something, it's misleading.

We should instead communicate that this is the default behavior, the user doesn't have to do anything unless they want to change the default behavior.

docs/source/user/saml.rst
```diff
@@ -142,3 +142,25 @@ Settings
...
+
+FAQs
+----
+
+.. _preventing_change_in_username_of_registered_user:
+
+Preventing change in username of a registered user
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``djangosaml2`` requires configuring the ``SAML_DJANGO_USER_MAIN_ATTRIBUTE``
```
nemesisdesign, line 154: Startig with

```
The ``djangosaml2`` library
```

sounds better

nemesisdesign (changes requested):

openwisp_radius/saml/backends.py
pandafy: > I believe if the user is created with SPID, we shall always update the username no matter what.
If the user was not created with SPID, we shall not update their username unless the setting is set to true.
The docs should reflect this difference.

The code functions like this in the current state. I have updated name of the setting.

PR #404, opened 2022-04-12
[Fix] Unique attribute for check or reply #335

### LINKED ISSUE #335: [bug] It should not be possible to define the same check or reply attributes multiple times
I believe it shouldn't be possible to define the same check or reply multiple times.

Right now it seems to me that a group check could define the same attribute multiple times for a group with different values.
Same applies for group replies.
But this doesn't make sense.

With checks and replies it's similar, there should be only one check/reply with a specific combination of organization_id, user_id (or username) and attribute. 

If the conditions above are broken we should raise a ValidationError.

codesankalp:
Fixes #335

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -198,24 +198,56 @@ class AutoGroupnameMixin(object):
             )
 
 
-class UniqueAtrributeMixin(object):
+class AttributeValidationMixin(object):
+    def _get_validation_queryset_kwargs(self):
+        raise NotImplementedError
+
+    def _get_error_message(self):
+        raise NotImplementedError
+
     def clean(self):
         """
         checks if the check or reply attribute is unique
         """
         model = type(self).__name__
-        if 'RadiusGroup' in model:
-            options = dict(group=self.group)
-        else:
-            options = dict(organization=self.organization, user=self.user)
-        options.update(attribute=self.attribute)
-        if load_model(model).objects.filter(**options).exclude(pk=self.pk).exists():
-            raise ValidationError({'attribute': _('This attribute is already in use.')})
+        if (
+            load_model(model)
+            .objects.filter(**self._get_validation_queryset_kwargs())
+            .exclude(pk=self.pk)
+            .exists()
+        ):
+            raise ValidationError({'attribute': self._get_error_message()})
         return super().clean()
 
 
+class UserAttributeValidationMixin(AttributeValidationMixin):
+    def _get_validation_queryset_kwargs(self):
+        return dict(
+            organization=self.organization, user=self.user, attribute=self.attribute
+        )
+
+    def _get_error_message(self):
+        object_name = 'check' if 'Check' in type(self).__name__ else 'reply'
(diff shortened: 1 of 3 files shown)


openwisp_radius/base/models.py
```diff
@@ -197,7 +198,25 @@ def clean(self):
...
+class UniqueAtrributeMixin(object):
+    def clean(self):
+        """
+        checks if the check or reply attribute is unique
+        """
+        model = type(self).__name__
+        if 'RadiusGroup' in model:
+            options = dict(group=self.group)
+        else:
+            options = dict(organization=self.organization, user=self.user)
```
codesankalp, line 210: Should I separate this conditional logic into separate mixin for groups?

nemesisdesign: This can be returned by a method, you can then create two mixins, one for group check/replies and one for non group check/replies, which return the right value.

eg:

```python
GroupAttributeValidationMixin(AttributeValidationMixin):
    def __get_validation_queryset_kwargs(self):
        return dict(group=self.group)
```

openwisp_radius/base/models.py
```diff
@@ -197,7 +198,25 @@ def clean(self):
...
+        options.update(attribute=self.attribute)
+        if load_model(model).objects.filter(**options).exclude(pk=self.pk).exists():
+            raise ValidationError({'attribute': _('This attribute is already in use.')})
```
nemesisdesign, line 213: The error message should be improved.

If it's a group check/reply should be:
`Another {object_name} for the same group and with the same attribute already exists`.

Otherwise:
`Another {object_name} for the same user and with the same attribute already exists`.

Where object_name is either group check, group reply, check or reply.

Also in this case, the error message could be returned by a method, eg:

eg:

```python
GroupAttributeValidationMixin(AttributeValidationMixin):
    def __get_validation_queryset_kwargs(self):
        return dict(group=self.group)

    def __get_error_message(self):
        object_name = .....  # find out whether it's check or a reply
        return f'Another {object_name} for the same group and with the same attribute already exists'  
```

openwisp_radius/base/models.py
```diff
@@ -197,7 +198,25 @@ def clean(self):
             )
 
 
-class AbstractRadiusCheck(OrgMixin, AutoUsernameMixin, TimeStampedEditableModel):
+class UniqueAtrributeMixin(object):
```
nemesisdesign, line 201: I think `AttributeValidationMixin` is a better name

nemesisdesign (changes requested):

nemesisdesign:
For the record, I discovered a flaw which made me realize the unique constraint is a better approach in the case of the RadiusGroup check and replies, however, I think this is good enough for now, in the future with more usage data I will be able to decide if adding a unique constraint wouldn't cause other issues and would really be better and hence we can change the code accordingly.

The flaw is the following: if multiple duplicate radius group checks / radius group replies are added via the admin in one save operation, the validation doesn't spot the duplicates.

PR #384, opened 2022-02-09
[change] Move password reset url to db #344

codesankalp:
Closes #344

PR #411, opened 2022-05-09
[fix] Fixed OrganizationRadiusSetting change form not populating default values #410

### LINKED ISSUE #410: [bug] Fallback char field does not show value if default
**Expected behavior:**

if a text field in org radius settings is set to the default value, we expect to see the value.

**Actual behavior:**

The text field is shown as empty.

pandafy:
- Renamed "FallbackBooleanField" to "FallbackBooleanChoiceField"
- Renamed "FallbackCharField" to "FallbackCharChoiceField"
- Made "FallbackTextField" return fallback value if database
  contains None. This automatically populates the field in
  the admin form.

Closes #410

### DIFF
--- a/openwisp_radius/admin.py
+++ b/openwisp_radius/admin.py
@@ -540,6 +540,7 @@ UserAdmin.list_select_related = ('registered_user',)
 
 class OrganizationRadiusSettingsInline(admin.StackedInline):
     model = OrganizationRadiusSettings
+    form = AlwaysHasChangedForm
     fieldsets = (
         (
             None,
--- a/openwisp_radius/base/fields.py
+++ b/openwisp_radius/base/fields.py
@@ -14,12 +14,19 @@ class FallbackMixin(object):
         return (name, path, args, kwargs)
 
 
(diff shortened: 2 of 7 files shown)


nemesisdesign (changes requested):
I still see the following problems.

Given the settings:

```python
OPENWISP_RADIUS_FREERADIUS_ALLOWED_HOSTS = ['127.0.0.1']
OPENWISP_RADIUS_ALLOWED_MOBILE_PREFIXES = ['+44', '+237']
OPENWISP_RADIUS_PASSWORD_RESET_URLS = {
    'default': (
        'http://localhost:8080/{organization}/password/reset/confirm/{uid}/{token}'
    ),
}
```

- I create a new org, I define only the name and save
- I expect to see the values of these fields non empty:
  - Freeradius allowed hosts
  - Allowed mobile prefixes:
  - Password reset URL

But those values are empty.
If I change something like the value of the SMS sender, then the values appear, except for the Password reset URL, isn't this field supposed to work in the same way? Initially this field had a fallback field too: https://github.com/openwisp/openwisp-radius/pull/384, but now that functionality is gone.

PR #416, opened 2022-06-02
[change] Rename default value of `OPENWISP_RADIUS_PASSWORD_RESET_URLS` to `__all__` #414

### LINKED ISSUE #414: [change] Rename deafult value of OPENWISP_RADIUS_PASSWORD_RESET_URLS to __all__
In https://github.com/openwisp/openwisp-monitoring/pull/376 we have used a different dict value for the fallback value, instead of `default` we have used `__all__` which may be more explicit.

I think it's a good idea to change it here too, while maintaining backward compatibility and still support `default` too, but the documentation should mention only `__all__`.

codesankalp:
Closes #414

### DIFF
--- a/openwisp_radius/settings.py
+++ b/openwisp_radius/settings.py
@@ -25,6 +25,12 @@ def get_settings_value(option, default):
     return getattr(settings, f'OPENWISP_RADIUS_{option}', default)
 
 
+def get_default_password_reset_url(urls):
+    # default is kept for backward compatibility,
+    # will be removed in future versions
+    return urls.get('default') or urls.get('__all__')
+
+
 RADIUS_API = get_settings_value('API', True)
 RADIUS_API_BASEURL = get_settings_value('API_BASEURL', '/')
 RADIUS_API_URLCONF = get_settings_value('API_URLCONF', None)
@@ -73,11 +79,7 @@ PASSWORD_RESET_URLS = {
     # use the uuid because the slug can change
 }
 PASSWORD_RESET_URLS.update(get_settings_value('PASSWORD_RESET_URLS', {}))
-DEFAULT_PASSWORD_RESET_URL = (
-    # kept for backward compatibility, will be removed in future versions
-    PASSWORD_RESET_URLS.get('default')
-    or PASSWORD_RESET_URLS.get('__all__')
-)
+DEFAULT_PASSWORD_RESET_URL = get_default_password_reset_url(PASSWORD_RESET_URLS)
 SMS_VERIFICATION_ENABLED = get_settings_value('SMS_VERIFICATION_ENABLED', False)
 # SMS_TOKEN_DEFAULT_VALIDITY time is in minutes
 SMS_TOKEN_DEFAULT_VALIDITY = get_settings_value('SMS_TOKEN_DEFAULT_VALIDITY', 30)
--- a/openwisp_radius/tests/test_admin.py
+++ b/openwisp_radius/tests/test_admin.py
@@ -594,6 +594,30 @@ class TestAdmin(
             radsetting.refresh_from_db()
             self.assertEqual(radsetting.password_reset_url, PASSWORD_RESET_URL)
 
+    def test_backward_compatible_default_password_reset_url(self):
+        # this test is only for backward compatible change
+        # it will be removed in the future
+        default_password_reset_url = (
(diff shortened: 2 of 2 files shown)


openwisp_radius/base/models.py
```diff
@@ -1174,8 +1174,10 @@ class AbstractOrganizationRadiusSettings(UUIDModel):
         null=True,
         blank=True,
         help_text=_PASSWORD_RESET_URL_HELP_TEXT,
-        default=PASSWORD_RESET_URLS.get('default'),
-        fallback=PASSWORD_RESET_URLS.get('default'),
+        default=PASSWORD_RESET_URLS.get('__all__')
+        or PASSWORD_RESET_URLS.get('default'),
+        fallback=PASSWORD_RESET_URLS.get('__all__')
+        or PASSWORD_RESET_URLS.get('default'),
```
nemesisdesign, line 1180: why instead of doing this a zillion times we don't just do it once in the app settings file and call just that?

```python
DEFAULT_PASSWORD_RESET_URL = (
    PASSWORD_RESET_URLS.get('__all__')
    # kept for backward compatibility, will be removed in future versions
    or PASSWORD_RESET_URLS.get('default')
)
```

nemesisdesign (changes requested):

docs/source/user/settings.rst
```diff
@@ -432,7 +432,7 @@ be used.
 .. code-block:: python
 
     {
-        'default': 'https://{site}/{organization}/password/reset/confirm/{uid}/{token}'
+        '__all__': 'https://{site}/{organization}/password/reset/confirm/{uid}/{token}'
```
pandafy, line 435: We also need to update this example https://github.com/openwisp/openwisp-radius/blame/f296f237983c51cb70b18f2329dba039cda9799f/docs/source/user/settings.rst#L464-L468

tests/openwisp2/settings.py
```diff
@@ -278,7 +278,7 @@
 
 # OPENWISP_RADIUS_PASSWORD_RESET_URLS = {
 #     # fallback in case the specific org page is not defined
-#     'default': 'https://example.com/{{organization}/password/reset/confirm/{uid}/{token},
+#     '__all__': 'https://example.com/{{organization}/password/reset/confirm/{uid}/{token},
 #     # use the uuid because the slug can change
 #     # 'dabbd57a-11ca-4277-8dbb-ad21057b5ecd': 'https://org.com/{organization}/password/reset/confirm/{uid}/{token}',
 # }
```
pandafy, line 284: @nemesisdesign @codesankalp why is this commented code present here? Is it present for instructions? If so, can we move the instructions to the part where the `__all__` key is configured?

codesankalp: I think it is for [instructions](https://github.com/openwisp/openwisp-radius/commit/fb663c69a2891aadea416d63e491ba44a53fe040#diff-cf01e5d691c05467162c955fc1688cc98afdb2ad6c6640caacb2569b0735a7c7)

I will move this block above `__all__` block.

pandafy (changes requested):
If I have the setting configured as following for the project 

```python
OPENWISP_RADIUS_PASSWORD_RESET_URLS = {
    'default': (
        'http://localhost:8080/{organization}/password/reset/confirm/{uid}/{token}'
    ),
}
```
The code takes the default value configured for `__all__` key. This makes the change backward incompatible. To be backward compatible, I would expect the code to use the value for the `default` key.

codesankalp:
> Can we add a simple test which ensures that the setting keeps working with the `default` key? It will help us prevent removing this backward compatibility unknowingly.

I am facing issues in writing tests for this specific case because even after mocking, the value remains the same in test. Is there any test present in openwisp codebase related to this?

pandafy (commented):
> Also, I have a question if user enters both the values in OPENWISP_RADIUS_PASSWORD_RESET_URLS 

I don't think we need to deal with this. Using `default` along with `__all__` is not something we allow. We can add a simple check for this, but clearly this is not an intended use case. At this point, the user is just messing with us (😂).

The README asks users to use `__all__`, if they use both they'll have unexpected behaviour. The logic you wrote right now looks correct to me. We will point it out in the CHANGELOG that users should migrate from `default` to `__all__`. You can add this note to the CHANGELOG in this PR itself if you like,

pandafy (approved):
LGTM! 👍🏼 

Deferring merge to @nemesisdesign

PR #424, opened 2022-06-29
[fix] Fixed organization cache bug in SAML ACS view 

pandafy:
In some cases, it possible that the organization cache for
the user is not updated before execution of the following
code. Hence, the cache is manually updated here.

### DIFF
--- a/openwisp_radius/saml/views.py
+++ b/openwisp_radius/saml/views.py
@@ -4,6 +4,7 @@ from urllib.parse import parse_qs, urlparse
 import swapper
 from django.conf import settings
 from django.contrib.auth import logout
+from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.shortcuts import get_object_or_404, render
 from djangosaml2.views import (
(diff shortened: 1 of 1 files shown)


openwisp_radius/saml/views.py
```diff
@@ -52,6 +53,10 @@ class AssertionConsumerServiceView(
 ):
     def post_login_hook(self, request, user, session_info):
         """If desired, a hook to add logic after a user has successfully logged in."""
+        # In some cases, it possible that the organization cache for
+        # the user is not updated before execution of the following
+        # code. Hence, the cache is manually updated here.
+        cache.delete('user_{}_organizations'.format(user.pk))
```
pandafy, line 59: @nemesisdesign the change that we did in openwisp-users 1.0.2 is not useful here because `openwisp-users.apps.OpenwispUsersConfig.update_organizations_dict` expects either an `OrganizationUser` or `OrganizationOwner` object. Therefore, I have used cache.delete here.

setup.py
```diff
@@ -36,7 +36,7 @@
     zip_safe=False,
     install_requires=[
         'django>=3.0,<=4.1.0',
-        'openwisp-users~=1.0.0',
+        'openwisp-users~=1.0.2',
```
nemesisdesign, line 39: in this case we can leave this alone then, this is not required

nemesisdesign (approved):

PR #427, opened 2022-07-22
[fix] Use storage backend method for deleting RadiusBatch.csvfile

pandafy:
The previous implementation used the "os" module for deleting resisdual
csv files. This causes issues when the project uses a file storage backend
other than based on file system.

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -995,9 +995,7 @@ class AbstractRadiusBatch(OrgMixin, TimeStampedEditableModel):
 
     def _remove_files(self):
         if self.csvfile:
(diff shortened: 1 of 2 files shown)


openwisp_radius/base/models.py
```diff
@@ -994,10 +994,7 @@ def expire(self):
             u.save()
 
     def _remove_files(self):
-        if self.csvfile:
```
nemesisdesign, line 997: I think we have to maintain this if, isn't it?

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #430, opened 2022-08-17
[change] Improved cleanup stale session command #426

### LINKED ISSUE #426: [change] Clean up stale sessions improvements
The following:

https://github.com/openwisp/openwisp-radius/blob/e61d2457a1c043dc56dce82f6de5bc6f9e5b7ced/openwisp_radius/management/commands/base/cleanup_stale_radacct.py#L18-L25

Should become a model method.
Then we should edit this so that we filter either:

- sessions which have a last update time older than the specified amount of time
- sessions which do not have any last update time set and have the start time older than the specified amount of time (kind of as it is now but with the difference that we take this only if update time is not set)

pandafy:
- Added "RadiusAccounting.close_stale_sessions" class method for
  closing stale sessions
- The method takes into account the "update_time" time of the
  session
- If the "update_time" is recent, then the session is not closed
  even when the "start_time" is older than the specified time.
  The "start_time" of a session is only checked when the
  "update_time" is not set.

Closes #426

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -531,6 +531,7 @@ class AbstractRadiusAccounting(OrgMixin, models.Model):
             session.session_time = (now() - session.start_time).total_seconds()
             session.stop_time = now()
             session.update_time = session.stop_time
+            session.terminate_cause = 'Session Timeout'
             session.save()
 
 
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -15,7 +15,7 @@ from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.mail import send_mail
 from django.db import models
-from django.db.models import ProtectedError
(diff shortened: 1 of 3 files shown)


openwisp_radius/base/models.py
```diff
@@ -515,6 +515,24 @@ class Meta:
...
+        # "update_time" is not set.
+        sessions = cls.objects.filter(
+            Q(update_time__lt=older_than)
+            | (Q(update_time=None) & Q(start_time__lt=older_than))
+        )
+        for session in sessions:
+            # calculate seconds in between two dates
+            session.session_time = (now() - session.start_time).total_seconds()
+            session.stop_time = now()
+            session.update_time = session.stop_time
```
nemesisdesign, line 533: one minor request which I had forgot to ask, if the termination cause field is empty, can you please fill it with `Session Timeout`?

nemesisdesign (changes requested):
One minor change requested, the rest is good :+1:

nemesisdesign (approved):

PR #433, opened 2022-08-30
[feature] Added setting to customize SMS message #432

### LINKED ISSUE #432: [feature] Allow customising SMS message text
Allow customising SMS message text at the organization level.

pandafy:
Closes #432


docs/source/user/settings.rst
```diff
@@ -882,3 +882,19 @@ The max number of SMS tokens a single user can request within a day.
...
+The template used for sending verification code to users via SMS.
+
+.. note::
+
+    The template should always contain ``{code}`` placeholder.
+    Otherwise, the sent SMS will not contain the verification code.
+
+This value can be overridden per organization in the organization
+change page. You can skip setting this option if you intend to set
+it from organization change page for each organization.
```
nemesisdesign, line 900: I would add another bit of text to explain the following:
```
Keep in mind that the default value is translated in other languages, If the value is customized the translations will not work, so if you need this message to be translated in different languages you should either not change the default value or prepare the additional translations.
```

Please can you double check to ensure the translation continues working also after this change?

docs/source/user/settings.rst
```diff
@@ -882,3 +882,19 @@ The max number of SMS tokens a single user can request within a day.
 
 The max number of tokens which can be requested from the same IP address
 during the same day.
+
+``OPENWISP_RADIUS_SMS_MESSAGE_TEMPLATE``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Default**: ``{organization_name} verification code: {code}``
```
nemesisdesign, line 889: You changed the string and broke the translations.

```
from django.utils.translation import gettext as _
from django.utils.translation import activate, get_language_info
activate("it")
_(org.radius_settings.sms_message)
 '{organization_name} verification code: {code}'
```

Why did you change `{organization}` to `{organization_name}`?
**That is a backward incompatible change.**

nemesisdesign (changes requested):
There's a detail regarding the translation of the text which is not clear so I am suggesting to amend the docs.

Here's one problems I found during testing that seem related to the fallback field.
Please can you double check these issues? If confirmed please create a new issue for this.

If I delete the sms message entirely and save, I expect the fallback value to be used and shown, but that's not the case, I think it's because the value saved to the DB is an empty string. I guess that when the value of the text field is falsy we shall set it to `None` automatically.

The rest is good.

pandafy:
> Here's one problems I found during testing that seem related to the fallback field.
> Please can you double check these issues? If confirmed please create a new issue for this.

I can replicate this issue locally. I have opened https://github.com/openwisp/openwisp-radius/issues/435

openwisp_radius/base/models.py
```diff
@@ -1295,6 +1307,22 @@ def _clean_password_reset_url(self):
...
+                        'The SMS message must contain the "{{code}}" '
+                        'placeholder, eg: {}.'.format(app_settings.SMS_MESSAGE_TEMPLATE)
+                    )
+                }
+            )
+        if (
+            not self.sms_message
+            or self.sms_message == app_settings.SMS_MESSAGE_TEMPLATE
+        ):
+            self.sms_message = None
```
nemesisdesign, line 1324: isn't this slightly redundant with https://github.com/openwisp/openwisp-radius/pull/436?

nemesisdesign (approved):

Issue #435, opened 2022-09-01
[bug] FallbackTextField does not handle empty strings


pandafy:
If the value of the of `FallbackTextField` is an empty string (`''`), then the form field does not show the fallback value. We should check the TextField for falsy values, i.e. for both `None` and empty string (`''`)

_Originally reported in https://github.com/openwisp/openwisp-radius/pull/433#pullrequestreview-1093123537_

nemesisdesign:
I would double check this on char fields too.

PR #436, opened 2022-09-02
[fix] Fixed fallback TextField and ChartField to handle empty strings #435

pandafy:
FallbackTextField and FallbackCharField defaults to fallback value
if they contain empty string.

Fixes https://github.com/openwisp/openwisp-radius/issues/435


pandafy:
We do not use FallbackCharField in openwisp-radius, but I have tested the fix manually.

openwisp_radius/base/fields.py
```diff
@@ -16,7 +16,7 @@ def deconstruct(self):
 
 class FallbackFromDbValueMixin:
     def from_db_value(self, value, expression, connection):
-        if value is None:
+        if value in [None, '']:
             return self.fallback
```
nemesisdesign, line 20: can we add logic so that if the value is the same as the default value we set it to `None`?
I thought that's how it worked but it seems to me that's not the case, but I do not remember very well what we did decide a few months ago?

openwisp_radius/base/models.py
```diff
@@ -1317,11 +1312,6 @@ def _clean_sms_message(self):
                     )
                 }
             )
-        if (
-            not self.sms_message
-            or self.sms_message == app_settings.SMS_MESSAGE_TEMPLATE
-        ):
-            self.sms_message = None
```
nemesisdesign, line 1324: Why is this here!? I am getting super confused.

nemesisdesign (changes requested):

openwisp_radius/base/models.py
pandafy: This PR was based on #433 earlier, hence these changes appeared here.

nemesisdesign (dismissed):

nemesisdesign (approved):

nemesisdesign (approved):

PR #438, opened 2022-09-16
[fix] Fixed validation for organization's password reset URL ##423

### LINKED ISSUE #423: [bug] Site key of organization's password reset url is not formatted automatically
The default value for `OPENWISP_RADIUS_PASSWORD_RESET_URLS` contains `{site}` placeholder. The documentation hints that the value for `{site}` is taken from the Django site object. 

> {site}: site domain as defined in the [django site framework](https://docs.djangoproject.com/en/dev/ref/contrib/sites/) (defaults to example.com and an be changed through the django admin)

I couldn't find the source code that deals with this. Also, when I try to save a new organization I get the following error

Screenshot from 2022-06-28 16-14-42

The URL field validation fails because of `{site}` placeholder.

pandafy:
Fixes #423


nemesisdesign (approved):
Thanks :+1:

PR #450, opened 2022-11-04
[feature] Added support for RADIUS CoA #447

### LINKED ISSUE #447: [feature] Implement change of authorization
Info on CoA:

- https://erbugcelebi.medium.com/coa-change-of-authorization-from-free-radius-server-914e3ccac2e9
- https://datatracker.ietf.org/doc/rfc5176/ (PDF available)

First of all we need to test this with coova-chilli: https://coova.github.io/CoovaChilli/chilli.conf(5).html (`coaport`)
We have to test the CoA using radclient as in the medium blog post. If that works we can proceed with the next.

The NAS IP must give info about how OpenWISP can reach the NAS.
It seems to me that for this we have to use the `radiuslisten` option or `nasip` on coova-chilli, but this means that the IP of the management tunnel needs to be static, which in turns means that we most probably have to use Wireguard.
We need to document all of this info in a new documentation page named "Change of Authorization".

Once all the above is confirmed or an equivally good alternatrive is found, I think we can proceed as follow:

- when the radius group assigned to a user changes, we shall initiate the CoA  - **only if the user has an open radius session**
- the CoA shall be executed by a celery task, if possible we should define the queue directly in the task itself and it shall be a new dedicated queue, so that we can make sure to have a celery worker that executes this task on the hosts where freeradius is running
- the task shall execute the radclient command or any equivalent good alternative
- the task shall find the open radius accounting session for the current user being acted upon, if there's any open session it shall read the NAS IP address and send the CoA to it, handling failures (log a warning message)
- let's add a global setting and an organization radius settings field (boolean) to control whether CoA is enabled or not, if disabled the code shall not be triggered

pandafy:
Closes #447

**TODOS**

- [x] Add documentation


openwisp_radius/radclient/client.py
```diff
@@ -0,0 +1,38 @@
...
+        Returns True if CoA request was accepted by NAS.
+        Otherwise, returns False.
+        """
+        attr = self.clean_attributes(attributes)
+        request = self.client.CreateCoAPacket(**attr)
+        try:
+            response = self.client.SendPacket(request)
+        except Timeout:
+            return False
+        return response.status_code == CoAACK
```
nemesisdesign, line 38: I think we must use more logging here:

- log with info level successful coa ack
- log with info level timeouts
- log with warn level coa nack

openwisp_radius/tasks.py
```diff
@@ -111,3 +116,89 @@ def send_login_email(accounting_data):
...
+            host=session.nas_ip_address,
+            radsecret=radsecret,
+            dicts=app_settings.RADCLIENT_ATTRIBUTE_DICTIONARIES,
+        )
+        result = client.perform_change_of_authorization(attributes)
+        if not result:
+            logger.warning(
+                f'Failed to perform CoA for "{session.unique_id}"'
+                f' RadiusAccounting object of "{user}" user'
+            )
```
nemesisdesign, line 204: I see the logging here but I think it's better to do it above as I recommended so we know whether the failure is due to timeout or a coa nack,

openwisp_radius/settings.py
```diff
@@ -81,6 +81,7 @@ def get_default_password_reset_url(urls):
 PASSWORD_RESET_URLS.update(get_settings_value('PASSWORD_RESET_URLS', {}))
 DEFAULT_PASSWORD_RESET_URL = get_default_password_reset_url(PASSWORD_RESET_URLS)
 SMS_VERIFICATION_ENABLED = get_settings_value('SMS_VERIFICATION_ENABLED', False)
+COA_ENABLED = get_settings_value('COA_ENABLED', False)
```
nemesisdesign, line 84: Please start adding this to the settings page and add a draft page for coa which is linked from this setting.

nemesisdesign (changes requested):
Looks ready to be tested, see my comments below for some minor improvements.

docs/source/user/settings.rst
```diff
@@ -180,6 +180,41 @@ If this option and organization change page option are both
...
+user's current RADIUS attributes whenever the ``RadiusGroup`` of
+user is changed. This allow enforcing of rate limits on active
+RADIUS sessions without requiring users to re-authenticate. For
+more details, :ref:`read the dedicated section for configuring
+openwisp-radius and NAS for using CoA <change_of_authorization>`.
+
+This can be overridden for each organization separately
+via the organization radius settings section of the admin interface.
+
+.. image:: /images/organization_coa_enabled.png
```
nemesisdesign, line 200: The screenshot is misleading because it's saying "Default (Enabled)", which is not the real default value used here, it most likely will be misinterpreted so please make sure the image shows the real default we use in the code to avoid misleading readers.

tests/openwisp2/radclient_dictionary.txt
```diff
@@ -0,0 +1,391 @@
...
+VALUE		Packet-Type	Accounting-Response		5
+VALUE		Packet-Type	Accounting-Status		6
+VALUE		Packet-Type	Password-Request		7
+VALUE		Packet-Type	Password-Accept			8
+VALUE		Packet-Type	Password-Reject			9
+VALUE		Packet-Type	Accounting-Message		10
+VALUE		Packet-Type	Access-Challenge		11
+VALUE		Packet-Type	Status-Server			12
+VALUE		Packet-Type	Status-Client			13
+
```
nemesisdesign, line 391: do you think it makes sense to ship this in the application code to make deployment easier?

nemesisdesign (changes requested):
Almost there!

docs/source/user/settings.rst
```diff
@@ -180,6 +180,41 @@ If this option and organization change page option are both
...
+| **type**:    | ``list`` |
++--------------+----------+
+| **default**: |  ``[]``  |
++--------------+----------+
+
+List of absolute file path of RADIUS dictionaries used
+for RADIUS attribute mapping.
+
+You can refer this `example dictionary <https://github.com/openwisp/openwisp-radius/blob/master/tests/openwisp2/radclient_dictionary.txt>`_
+to get started.
```
nemesisdesign, line 216: Let's ammend here as discussed (have the default dictionary automatically included here and allow other dictionaries to be included).

nemesisdesign (changes requested):

nemesisdesign (approved):

PR #455, opened 2022-11-27
[change] Make openwisp-radius docs similar to new theme

Aryamanz29:
While reading the docs, I found that we're still using the Sphinx[ default theme,](https://sphinx-rtd-theme.readthedocs.io/en/stable/) so I took this opportunity to change it to the new [OpenWISP theme.](https://openwisp.io/docs/index.html)

### `Before`

Screenshot from 2022-11-27 17-54-19

### `After`

Screenshot from 2022-11-27 17-57-07


nemesisdesign (approved):
This is great Aryaman.

My only concern is that we also have https://github.com/openwisp/netjsonconfig which has a documentation website built with sphinx and if we do this there we will have the same CSS duplicated in multiple repositories which is not gonna be good for maintainability in the long run.

PR #465, opened 2023-01-03
[fix] Fixed RadiusCheck/RadiusReply empty org field raises 500 error …

### LINKED ISSUE #464: [bug] Saving RadiusCheck/RadiusReply without an organization raises server 500 error
**Steps to replicate**

1. Go to http://127.0.0.1:8000/admin/openwisp_radius/radiuscheck/add/
2. Selected "Guided (dropdown)" mode
3. Fill out the form as follows (leaving out the organization field)

The same happens for RadiusReply.

Screenshot from 2022-12-20 18-38-50

**Expected outcome**

The admin form complains for empty organization field 

**Actual outcome**

Server 500 error 

Screenshot from 2022-12-20 18-39-10

Aryamanz29:
- Fixed `RadiusCheck/RadiusReply` empty `org field` raises **500 error**.

Screenshot from 2023-01-03 16-31-19

    
 Fixes #464

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -234,12 +234,12 @@ class AttributeValidationMixin(object):
 
 class UserAttributeValidationMixin(AttributeValidationMixin):
     def _get_validation_queryset_kwargs(self):
+        kwargs = dict(user=self.user, attribute=self.attribute)
+        org = getattr(self, 'organization', None)
         # only add `organization` key if it exists
-        if getattr(self, 'organization_id'):
-            return dict(
-                organization=self.organization, user=self.user, attribute=self.attribute
-            )
-        return dict(user=self.user, attribute=self.attribute)
+        if org:
+            kwargs['organization'] = org
+        return kwargs
 
     def _get_error_message(self):
         return _(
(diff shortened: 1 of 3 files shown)


openwisp_radius/base/models.py
```diff
@@ -234,9 +234,12 @@ def clean(self):
...
     def _get_validation_queryset_kwargs(self):
-        return dict(
-            organization=self.organization, user=self.user, attribute=self.attribute
-        )
+        # only add `organization` key if it exists
+        if getattr(self, 'organization_id'):
+            return dict(
+                organization=self.organization, user=self.user, attribute=self.attribute
+            )
+        return dict(user=self.user, attribute=self.attribute)
```
nemesisdesign, line 242: there's a bit of duplication (redefinition of the same dict keys/values) and inefficiency (the call to `getattr(self, 'organization_id')` is wasted) here, a better way would be:

```python
kwargs = dict(user=self.user, attribute=self.attribute)
org = getattr(self, 'organization', None)
if org:
    kwargs['organization'] = org
return kwargs
```

nemesisdesign (changes requested):

openwisp_radius/base/models.py
Aryamanz29: noted.

nemesisdesign (approved):

PR #478, opened 2023-06-23
[change] Grouped SMS features in organization admin #469

### LINKED ISSUE #469: [change] Group SMS features in organization admin
The rows relating to SMS sending shall be grouped together (that is, shown one next to the other), the sms settings should be shown only when sms sending is enabled. We've been doing similar things in other modules or even in this module: https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/static/openwisp-radius/js/mode-switcher.js

pandafy:
Closes #469

### DIFF
--- a/openwisp_radius/admin.py
+++ b/openwisp_radius/admin.py
@@ -553,14 +553,10 @@ class OrganizationRadiusSettingsInline(admin.StackedInline):
                     'saml_registration_enabled',
                     'social_registration_enabled',
                     'needs_identity_verification',
-                    'sms_verification',
                     'first_name',
                     'last_name',
                     'birth_date',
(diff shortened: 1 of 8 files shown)


nemesifier (changes requested):
The current `Sms meta data` field grouping doesn't make sense (we had discussed this):

Screenshot from 2023-06-26 11-14-21

Let's also take advantage of this chance to make sure all the labels of the SMS fields are consistent in showing the uppercase "SMS" acronym:

- Sms meta data > SMS meta data
- Sms verification > SMS verification

It would be good to do a project search & replace here just to be sure we're consistent.

nemesifier (approved):

PR #485, opened 2023-07-25
[feature] Limit consecutive SMS tokens #481

### LINKED ISSUE #481: [feature] Limit consecutive SMS tokens (backend REST API)
We need to introduce changes to the REST API to reject new SMS from the same user before X seconds after the last SMS sent, where X is configurable.

The waiting time shall be configurable via the organization radius settings in the admin using a fallback field, I recommend a positive integer field or anything equivalent (the value must be higher than zero). The default value of the setting can be 10 seconds.

The REST API for sending a new SMS token shall return the amount of seconds that need to pass before being allowed to send another SMS in the response so that consumer knows how to make use of this information in the UI, another good reason to do so is that we do not need to duplicate this configuration setting in the consumer application.

Include translations for all languages, a translation service like Google Translate can be used for the first iteration, I will get the translations reviewed before merging.

pandafy:
Closes #481

**Blockers**

- [ ] https://github.com/openwisp/openwisp-utils/pull/353


.github/workflows/ci.yml
```diff
@@ -53,6 +53,7 @@ jobs:
         run: |
           pip install -e .[saml,openvpn_status]
           pip install ${{ matrix.django-version }}
+          pip install --force-reinstall --no-deps "https://github.com/openwisp/openwisp-utils/tarball/fallback-positiveintegerfield"
```
pandafy, line 56: TODO: Remove this before merging.

nemesifier: you can switch back to master

docs/source/user/settings.rst
```diff
@@ -410,6 +410,26 @@ For example:
 Using the setting above will only allow phone numbers from the UK (``+44``)
 or Cameroon (``+237``).
 
+.. note::
+
+    This setting is applicable only for organizations
+    which have :ref:`enabled the SMS verification option
+    <openwisp_radius_sms_verification_enabled>`.
+
+``OPENWISP_RADIUS_SMS_REQUEST_TIMEOUT``
```
nemesifier, line 419: Time out is not a good name for this.

I would change it to ``OPENWISP_RADIUS_SMS_COOLDOWN``.

openwisp_radius/admin.py
```diff
@@ -571,6 +571,7 @@ class OrganizationRadiusSettingsInline(admin.StackedInline):
                     'sms_sender',
                     'sms_message',
                     'allowed_mobile_prefixes',
+                    'sms_timeout',
```
nemesifier, line 574: same here

openwisp_radius/base/models.py
```diff
@@ -1109,6 +1110,16 @@ class AbstractOrganizationRadiusSettings(UUIDModel):
         ),
         fallback=app_settings.SMS_MESSAGE_TEMPLATE,
     )
+    sms_timeout = FallbackPositiveIntegerField(
```
nemesifier, line 1113: same here

openwisp_radius/tests/test_api/test_phone_verification.py
```diff
@@ -184,6 +190,30 @@ def test_create_phone_token_400_limit_reached(self):
...
+
+        with freeze_time(start_time + timedelta(seconds=15)):
+            response = self.client.post(url, HTTP_AUTHORIZATION=token_header)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data['timeout'], 15)
+        self.assertEqual(
+            response.data['non_field_errors'],
+            ['Wait before requesting another SMS token.'],
+        )
+
```
nemesifier, line 216: Please add another test to ensure the API responds with 201 after the cooldown time.

nemesifier (changes requested):
See my comments below.

docs/source/user/settings.rst
```diff
@@ -410,6 +410,26 @@ For example:
 Using the setting above will only allow phone numbers from the UK (``+44``)
 or Cameroon (``+237``).
 
+.. note::
+
+    This setting is applicable only for organizations
+    which have :ref:`enabled the SMS verification option
+    <openwisp_radius_sms_verification_enabled>`.
```
nemesifier, line 417: is this note referred to `OPENWISP_RADIUS_SMS_COOLDOWN`? If so, please move it after the heading, not before.

nemesifier (changes requested):

docs/source/user/settings.rst
pandafy: This note is for `OPENWISP_RADIUS_ALLOWED_MOBILE_PREFIXES`

nemesifier (approved):

PR #487, opened 2023-08-13
Bug Fix: Different types of phone numbers are allowed, but only mobile types should

### LINKED ISSUE #484: [bug] Different types of phone numbers are allowed, but only mobile types should
Landline numbers seem to be currently accepted, which will not work for sending SMS.

I think we have to edit this part of the code:

https://github.com/openwisp/openwisp-radius/blob/0cfe07133551d60479f85a8c4f619a39aa09495b/openwisp_radius/api/serializers.py#L430-L443

```
# assuming phone_number is a phonenumbers instance
phonenumbers.phonenumberutil.number_type(phone_number)
```

Possible return values according to [the library's code](https://github.com/daviddrysdale/python-phonenumbers/blob/c2fe6f480f24939ad64a53887e53bc6c27f7a9ab/python/phonenumbers/phonenumberutil.py#L439):

FIXED_LINE = 0
MOBILE = 1
FIXED_LINE_OR_MOBILE = 2
TOLL_FREE = 3
PREMIUM_RATE = 4
SHARED_COST = 5
VOIP = 6
PERSONAL_NUMBER = 7
PAGER = 8
UAN = 9
VOICEMAIL = 10
UNKNOWN = 99

We shall accept only 1 (MOBILE).

ujjwalkirti:
I changed the code to check whether phone_number provided is a Mobile number and raise expception if not. (#484 )
Below is the function code that I have altered. 

```
 def validate_phone_number(self, phone_number):
        org = self.context['view'].organization
        if get_organization_radius_settings(org, 'sms_verification'):
            if phonenumbers.phonenumberutil.number_type(phone_number) != 1:
                raise serializers.ValidationError(_('Landline numbers can\'t be used to send SMS.'))
            if not phone_number:
                raise serializers.ValidationError(_('This field is required.'))
            mobile_prefixes = org.radius_settings.allowed_mobile_prefixes_list
            if not self.is_prefix_allowed(phone_number, mobile_prefixes):
                raise serializers.ValidationError(
                    _('This international mobile prefix is not allowed.')
                )
            if User.objects.filter(phone_number=phone_number).exists():
                raise serializers.ValidationError(
                    _('A user is already registered with this phone number.')
                )
        else:
            # Phone number should not be stored if sms verification is disabled
            phone_number = None
        return phone_number
```

---------
Closes #484 

Edited by @pandafy

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -20,6 +20,7 @@ from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from phonenumber_field.serializerfields import PhoneNumberField
+from phonenumbers import PhoneNumberType, phonenumberutil
 from rest_framework import serializers
 from rest_framework.authtoken.serializers import (
     AuthTokenSerializer as BaseAuthTokenSerializer,
@@ -436,6 +437,14 @@ class RegisterSerializer(
             if not self.is_prefix_allowed(phone_number, mobile_prefixes):
                 raise serializers.ValidationError(
                     _('This international mobile prefix is not allowed.')
+                )
+            phone_number_type = phonenumberutil.number_type(phone_number)
+            allowed_types = [PhoneNumberType.MOBILE]
+            if app_settings.ALLOW_FIXED_LINE_OR_MOBILE:
+                allowed_types.append(PhoneNumberType.FIXED_LINE_OR_MOBILE)
+            if phone_number_type not in allowed_types:
+                raise serializers.ValidationError(
+                    _('Only mobile phone numbers are allowed.')
                 )
             if User.objects.filter(phone_number=phone_number).exists():
                 raise serializers.ValidationError(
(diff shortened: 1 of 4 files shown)


ujjwalkirti:
@nemesifier can you please review this PR for the issue #484

openwisp_radius/api/serializers.py
```diff
@@ -430,6 +430,8 @@ class RegisterSerializer(
     def validate_phone_number(self, phone_number):
         org = self.context['view'].organization
         if get_organization_radius_settings(org, 'sms_verification'):
+            if phonenumbers.phonenumberutil.number_type(phone_number) != 1:
+                raise serializers.ValidationError(_('Landline numbers can\'t be used to send SMS.'))
```
nemesifier, line 434: I'd write that only mobile phone numbers are allowed.

nemesifier (changes requested):
@ujjwalkirti thanks for contributing! We need a failing unit test which replicates the bug before we can merge a fix. Do you know how to write tests? If you look at https://github.com/openwisp/openwisp-radius/blob/master/openwisp_radius/tests/test_api/test_phone_verification.py you can find plenty of examples.

openwisp_radius/api/serializers.py
ujjwalkirti: Okay, I will make the required changes.

nemesifier (approved):

PR #488, opened 2023-08-31
[feature] Added intergration with openwisp-monitoring

pandafy:
**Blockers**
- [x] https://github.com/openwisp/openwisp-monitoring/pull/538

Closes https://github.com/openwisp/openwisp-monitoring/issues/532
Closes https://github.com/openwisp/openwisp-monitoring/issues/534


nemesifier (changes requested):
Tests are failing on Django 3.2.0 here.

nemesifier (changes requested):
This is still failing, I restarted the build to find out whether it was a temporary failure because the error looked weird (had to do with a non-existent time).

nemesifier (approved):
The build passed, let's keep testing this a bit more before merging.

nemesifier (changes requested):
Problem found during testing: the count of total users should be taken from the SQL DB because influxdb doesn't know. The goal of the total users chart is to show the total amount of users registered in the system at any given time.

nemesifier (changes requested):
Let's add more space between date filter and the pie charts.

Can we change the default colors of the RADIUS charts to differentiate between traffic charts?
They look too similar to the traffic charts and do not stand out.
Eg: (total: #3A3335, #FDF0D5, #C6D8D3, #F0544F, #D81E5B, #048A81, #343E8A)

pandafy:
We found these shortcomings during the review 

- Write total number of users registered on the systems to influxdb once a day. 
  No need to write the data point on every user registration. 
- Change the colors in the charts.

pandafy:
**Summary of Changes:**

The current implementation faced a challenge wherein the charts for "User Registration" and "Total User Registration" started from zero, displaying the number of users registered since the feature deployment rather than the total count of users registered on the system.

To address this, the proposed solution was to write the total number of registered users to InfluxDB once a day, eliminating the need for a data point write on every user registration event.

**Solution Exploration:**

In the latest iteration, I delved into the feasibility of the proposed solution. Initial investigations revealed a necessity to write metric points every hour instead of once a day, as the chart resolution for "last 7 days" draws a point for each hour.

**Implementation Details:**

To accommodate this, the following logic was implemented in Python:

```python
def write_user_registration_metrics():
    now = timezone.now()
    metric_data = []
    # Get the registration data for the past hour.
    # The query returns a tuple of organization_id, registration_method and
    # count of users who registered with that organization and method.
    registered_users = (
        RegisteredUser.objects.filter(
            user__openwisp_users_organizationuser__created__lte=now,
            user__openwisp_users_organizationuser__created__gt=now
            - timezone.timedelta(hours=1),
        )
        .values_list('user__openwisp_users_organizationuser__organization_id', 'method')
        .annotate(count=Count('user', distinct=True))
    )
    for org_id, registration_method, count in registered_users:
        registration_method = clean_registration_method(registration_method)
        metric, _ = Metric._get_or_create(
            configuration='user_signups_new',
            name='User SignUps',
            key='user_signups_new',
            object_id=None,
            content_type=None,
            extra_tags={
                'organization_id': str(org_id),
                'method': registration_method,
            },
        )
        metric_data.append((metric, {'value': count, 'time': now}))
    Metric.batch_write(metric_data)
```

**Caveats Identified:**

However, two caveats were encountered:

1. The logic wrote a metric point for each organization when a user registered with multiple organizations, causing the total user count to be misrepresented.
2. Users without a corresponding RegisteredUser object were not considered, such as manually created superusers, further contributing to data inconsistency.
Decision and Refinement:

After discussion with @nemesifier, it was concluded that the current implementation, which allows a DISTINCT operation on user IDs, resolves the first caveat effectively.

To address the second caveat, it was decided to update the logic to write a metric point for all users created. For users without a RegisteredUser object, the metric point will be assigned an "unspecified" method.

**Resolution:**

This adjustment resolves the mismatch between the total user count displayed on charts and the user's changelist page.

Furthermore, to rectify the initial problem, a celery task will be added to retroactively write metric points for old data from the database. This celery task would get triggered on the post_migration signal, thus aligning the implementation with the desired outcome.

pandafy:
I had added this task for writing RADIUS metrics retroactively from the database 

```python
@shared_task
def post_migrate_retroactive_write():
    """
    This task retroactively writes the RADIUS metrics
    to InfluxDB using the data present in the database.
    """
    metric_data = []
    for user in (
        User.objects.select_related('registered_user')
        .only('id', 'registered_user__method')
        .iterator()
    ):
        try:
            registration_method = user.registered_user.method
        except RegisteredUser.DoesNotExist:
            registration_method = 'unspecified'
        registration_method = clean_registration_method(registration_method)
        org_query = list(
            user.openwisp_users_organizationuser.values_list(
                'organization_id',
                'created',
            )
        )
        if not org_query:
            # This user is not a member of any organization.
            # Use "None" for organization_id for this user.
            org_query = [(None, user.date_joined)]
        for (org_id, time) in org_query:
            metric = _get_user_signup_metric(
                organization_id=str(org_id), registration_method=registration_method
            )
            metric_data.append(
                # (metric, {'value': sha1_hash(str(user.id)), 'time': time})
                (metric, {'value': user.username, 'time': time})
            )
    Metric.batch_write(metric_data)
```

Upon testing with realistic data, I encountered two caveats with this: 
1. The InfluxDB's retention policy will not accept any data older than the defined limits. Thus, we can not store the metrics from the beginning of time (time of installation). Moreover, InfluxDB would keep deleting older data as the time progresses. This would lead to inconsistent values for the total number of users. 
2. The chart only shows total for the selected time range. E.g. if there are total 60 users - 59 users registered in last year and one user registered in last 2 years, the chart will only show 59 users when the time range is set to "last 365 days"

docs/source/user/radius_monitoring.rst
```diff
@@ -0,0 +1,81 @@
...
+1. User registrations
+~~~~~~~~~~~~~~~~~~~~~
+
+.. image:: /images/user-registration-chart.png
+   :alt: User registration chart
+
+This chart shows number of users signed up using different registration methods.
+
+2. Total user registrations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
nemesifier, line 23: this line is longer than the title, let's also update the charts images once we are done with everything else

nemesifier (changes requested):
Please add a setting which allows to invert download and upload when writing data to influxdb, defaulting to false.

Have you moved the RADIUS sessions tab in the device page yet?

nemesifier (changes requested):
Let's fix these UI issues:

1. there's a dot which looks like it shouldn't be there:
Screenshot from 2024-03-19 14-38-43

2. Let's make the styel more consistent with the "WiFi Sessions" tab, alignment, padding, spacing and the call to action spacing.

Screenshot from 2024-03-19 14-38-50

Then, please also address thef following:

- code coverage
- fill in user registration graph, I think adding the fill here should work well

nemesifier (changes requested):
Refinements and fixes for issues found during the last round of testing:

- acct stats: never lose any stat, if there's no registered user flag it as "unspecified" in the registration method
- remove the fill from the "User Registration"
- rename "User Registration" to "User Registrations"
- rename "Unique RADIUS Session Count" to "Unique RADIUS Sessions"
- rename "Total RADIUS Sessions Traffic" to "Traffic of RADIUS Sessions"

Please update the openwisp-monitoring branch (dependent PR was merged).

openwisp_radius/integrations/monitoring/tasks.py
```diff
@@ -198,9 +198,9 @@ def post_save_radiusaccounting(
     except RegisteredUser.DoesNotExist:
         logger.warning(
```
nemesifier, line 199: I would change this to info as this can happen in different scenarios

nemesifier (changes requested):

nemesifier (changes requested):
Another point which I forgot to add here:

- Can we allow the device charts to ignore the organization if configured to do so?
- The RADIUS session tab already ignores the organization field, but this is inconsistent with the chart behavior, ideally we should have a setting to configure this, the default behavior should be to enforce the organization lookup

pandafy:
> * Can we allow the device charts to ignore the organization if configured to do so?

I was looking into the queries made to the InfluxDB when opening the device charts. I found the following relevant queries related to RADIUS sessions

```
SELECT SUM(output_octets) / 1000000000 AS upload, SUM(input_octets) / 1000000000 AS download FROM radius_acc WHERE time >= '2023-04-19'  AND content_type = 'config.device' AND object_id = '4d1a6d5d-a638-46fb-8297-6ea050388985' GROUP BY time(7d)
SELECT COUNT(DISTINCT(username)) FROM radius_acc WHERE time >= '2023-04-19'  AND content_type = 'config.device' AND object_id = '4d1a6d5d-a638-46fb-8297-6ea050388985' GROUP by time(7d), method
```

These do not filter by organization

pandafy:
When writing RADIUS accounting metric, we may not always want to lookup the device with the organization of the RadiusAccounting object. We shall add a setting, which if enabled removes the organization_id from the Device lookup in `post_save_radiusaccounting` task.

nemesifier (changes requested):
Calling-station-id needs to be hashed.

docs/source/user/settings.rst
```diff
@@ -1044,3 +1044,24 @@ the default value is translated in other languages. If the value is
...
+
+OpenWISP Monitoring integration related settings
+================================================
+
+.. note::
+
+    This settings are only used if you have enabled the
+    :ref:`integration_with_openwisp_monitoring`.
+
+``OPENWISP_RADIUS_MONITORING_DEVICE_LOOKUP_IGNORE_ORGANIZATION``
```
pandafy, line 1056: @nemesifier can you recommend a better name for this setting?

.gitignore
```diff
@@ -1,3 +1,5 @@
+tests/openwisp2/saml/*
```
nemesifier, line 1: what does this has to do with this PR?

setup.cfg
```diff
@@ -19,6 +19,8 @@ force_grid_wrap=0
 ignore = W605, W503, W504, E203
 exclude = *.egg-info,
           .git,
+          ./tests/openwisp2/saml/*
+          ./test-*
```
nemesifier, line 23: what does this has to do with this PR?

.gitignore
pandafy: These changes are related to my local environment which accidentally got added to this PR. I have removed these changes from the PR.

docs/source/user/radius_monitoring.rst
```diff
@@ -0,0 +1,83 @@
+.. _integration_with_openwisp_monitoring:
+
+Integration with OpenWISP Monitoring
+------------------------------------
+
+OpenWISP RADIUS includes an optional Django sub-app that adds integration with
+`OpenWISP Monitoring <https://github.com/openwisp/openwisp-monitoring#openwisp-monitoring>`_
```
nemesifier, line 7: udpate link

docs/source/user/radius_monitoring.rst
```diff
@@ -0,0 +1,83 @@
...
+        # ...
+    ]
+
+.. note::
+
+    Ensure your Django project is correctly configured to utilize OpenWISP Monitoring as
+    outlined in the `OpenWISP Monitoring's documentation <https://github.com/openwisp/openwisp-monitoring#setup-integrate-in-an-existing-django-project>`_.
+    For production environments, it is advisable to deploy OpenWISP using
+    `Ansible OpenWISP2 <https://github.com/openwisp/ansible-openwisp2>`_ or
+    `Docker OpenWISP <https://github.com/openwisp/docker-openwisp>`_, as they simplify
```
nemesifier, line 74: update all links to point to new docs

docs/source/user/radius_monitoring.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. image:: /images/radius-traffic-chart.png
+   :alt: RADIUS traffic chart
+
+This chart shows the RADIUS traffic generated by user sessions.
+
+Enabling RADIUS metrics in Django project
+=========================================
+
+You can enable the monitoring integration by including ``openwisp_radius.integrations.monitoring``
```
nemesifier, line 53: add the usual note with link to how to edit settings

docs/source/user/radius_monitoring.rst
```diff
@@ -0,0 +1,83 @@
...
+.. code-block:: python
+
+    # In your_project/settings.py
+
+    INSTALLED_APPS = [
+        # ...
+        'openwisp_radius',
+        'openwisp_radius.integrations.monitoring' # <--- add the app after openwisp_radius
+        # ...
+    ]
```
nemesifier, line 66: What about:
```python
INSTALLED_APPS.append('openwisp_radius.integrations.monitoring')
```
Should be easier for users to copy/paste.

openwisp_radius/integrations/monitoring/apps.py
```diff
@@ -0,0 +1,121 @@
...
+        super().ready()
+        self.register_dashboard_charts()
+        self.register_radius_metrics()
+        self.connect_signal_receivers()
+
+    def register_dashboard_charts(self):
+        register_dashboard_chart(
+            position=30,
+            config={
+                'name': _("Today's RADIUS sessions"),
```
nemesifier, line 34: quoting style

nemesifier (changes requested):

openwisp_radius/integrations/monitoring/apps.py
pandafy: The double quotes is used because there's an apostrophe present in the text. Shall change it to 
```suggestion
                'name': _('Today\'s RADIUS sessions'),
```

IMO,  this is less readable

Issue #490, opened 2023-10-02
[feature] Add mac address authentication for roaming users


nemesifier:
A prototype of this feature is available here: https://github.com/openwisp/openwisp-radius/tree/macauth-roaming.

We need to recognize users who are moving from one NAS to another one so that they can avoid having to authenticate again.
This feature must be optional and turned off by default, we want to allow turning this on via the org admin settings.

PR #492, opened 2023-10-10
[feature] Added mac address authentication for roaming users #490

pandafy:
Closes https://github.com/openwisp/openwisp-radius/issues/490


nemesifier (changes requested):

openwisp_radius/api/freeradius_views.py
```diff
@@ -112,13 +114,42 @@ def _check_client_ip_and_return(self, request, uuid):
...
+                    username = open_session.username
+                    if hasattr(request.data, '_mutable'):
+                        request.data._mutable = True
+                    request.data['username'] = username
+                    if hasattr(request.data, '_mutable'):
+                        request.data._mutable = False
+                    cache.set(
+                        f'rt-{calling_station_id}',
+                        str(open_session.organization_id),
+                        86400,
```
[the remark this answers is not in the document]
pandafy: In this case, the RADIUS Session would get created with org-A unless authorization header is set.

openwisp_radius/receivers.py
```diff
@@ -105,7 +105,6 @@ def close_previous_radius_accounting_sessions(instance, created, **kwargs):
         unique_id=instance.unique_id
     ).filter(
         stop_time__isnull=True,
-        called_station_id=instance.called_station_id,
         calling_station_id=instance.calling_station_id,
         username=instance.username,
```
nemesifier, line 109: Are we sure the `instance.username` we get here is the real username and not the mac address?
Please add some verbose logging in this part so we can debug this issue.

nemesifier (changes requested):
Great progress @pandafy :+1: 

During my testing the previous session did not get closed though:
Screenshot from 2023-10-23 16-37-26

I tried twice.

openwisp_radius/receivers.py
```diff
@@ -100,22 +100,27 @@ def close_previous_radius_accounting_sessions(instance, created, **kwargs):
...
     )
+    logger.info(f'open_sessions: {open_sessions}')
     for session in open_sessions:
         session.stop_time = session.update_time
```
pandafy, line 114: @nemesifier in our testing we are not sending a RADIUS packet (`Interim-Update`) which sets the `update_time` field. Since, the `update_time` field is also set to `None`, the `stop_time` also stays `None`.

nemesifier: ok so let's check if `update_time` is None and in that case let's set the current time on it, therefore `stop_time` should follow suit.

docs/source/user/settings.rst
```diff
@@ -603,6 +603,31 @@ via the organization radius settings section of the admin interface.
...
+
+.. image:: /images/mac-address-roaming.png
+   :alt: Organization MAC Address Roaming settings
+
+.. note::
+
+    We recommend using the override via the admin interface only when there
+    are special organizations which need a different configuration, otherwise,
+    if all the organization use the same configuration, we recommend
+    changing the global setting.
```
nemesifier, line 628: is this necessary? Wouldn't this apply to all settings? Why this one in particular?

docs/source/user/settings.rst
```diff
@@ -603,6 +603,31 @@ via the organization radius settings section of the admin interface.
...
+
+
```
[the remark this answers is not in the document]
nemesifier: forgot this one?

openwisp_radius/api/freeradius_views.py
```diff
@@ -141,7 +169,24 @@ def authenticate(self, request):
...
+                username, request
+            )
+            if username is None and request is None:
+                # When using MAC roaming, the "username" attribute contains the
+                # MAC address (calling_station_id). If there's no open-session
+                # for the given MAC address, then don't authenticate the user.
+                #
+                # NOTE: We return "None" here instead of raising "AuthenticationFailed"
+                # to prevent unnecessary authentication rejected errors in freeradius
+                # logs.
```
nemesifier, line 187: I'd rewrite it like:

```
# When using MAC auth roaming, the "username" attribute contains the
# MAC address (calling_station_id). When the user connects the first time,
# since it doesn't have any open session, the mac address authorization
# will fail, in order to avoid filling the log with failed mac address
# authorization requests we return "None" here
# (instead of raising "AuthenticationFailed").
# freeradius will take care of rejecting the authorization if no
# explicit "Auth-Type: Accept" is returned
```

openwisp_radius/api/freeradius_views.py
```diff
@@ -216,7 +261,8 @@ def post(self, request, *args, **kwargs):
         serializer.is_valid(raise_exception=True)
         username = serializer.validated_data.get('username')
         password = serializer.validated_data.get('password')
-        user = self.get_user(request, username)
+
```
nemesifier, line 264: why this additional blank line?

nemesifier (commented):
Please see my comments below.

docs/source/user/settings.rst
pandafy: We have this warning on other settings as well. 

https://github.com/search?q=repo%3Aopenwisp%2Fopenwisp-radius+%22We+recommend+using+the+override+via+the+admin+interface%22&type=code

nemesifier (approved):

PR #501, opened 2023-12-18
[feature] Added API endpoint to return user's RADIUS usage #499

### LINKED ISSUE #499: [feature] Add API endpoint to return user's data usage
We need an endpoint that returns data used by the user/total data. The implementation should be flexible to also work with other RADIUS checks (e.g. session-time). 

We need to implement something similar to `openwisp_radius.api.freeradius_views.AuthorizeView.get_replies` where result from each counter is returned. 

https://github.com/openwisp/openwisp-radius/blob/90786b1a9d84d7d9121ca962de0c159e46985551/openwisp_radius/api/freeradius_views.py#L294-L335

We may need to add a verbose name for the Check model. We cannot show the `Check.attribute` in the UI as it won't make any sense to the users.

pandafy:
Closes #499

**TODOS**

- [ ] Improve coverage 
- [ ] Add docs

### DIFF
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -666,7 +666,8 @@ User Radius Usage
 Returns the radius usage of the logged-in user and the organization specified
 in the URL.
 
-It executes the relevant RADIUS counters and returns the exhausted values.
+It executes the relevant RADIUS counters and returns information that
+shows how much time and/or traffic the user has consumed.
 
 .. code-block:: text
 
--- a/docs/source/user/api.rst
+++ b/docs/source/user/api.rst
@@ -656,6 +656,24 @@ in the URL.
 
 Responds only to **GET**.
 
+.. _radius_usage_api_view:
+
+User Radius Usage
+-----------------
+
+**Requires the user auth token (Bearer Token)**.
+
+Returns the radius usage of the logged-in user and the organization specified
+in the URL.
+
+It executes the relevant RADIUS counters and returns the exhausted values.
+
+.. code-block:: text
+
+    /api/v1/radius/organization/<organization-slug>/account/usage/
+
+Responds only to **GET**.
+
 Create SMS token
 ----------------
 
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -26,14 +26,16 @@ from rest_framework.authtoken.serializers import (
     AuthTokenSerializer as BaseAuthTokenSerializer,
 )
 from rest_framework.fields import empty
+from rest_framework.generics import get_object_or_404
 
 from openwisp_radius.api.exceptions import CrossOrgRegistrationException
 from openwisp_users.backends import UsersAuthenticationBackend
 
 from .. import settings as app_settings
 from ..base.forms import PasswordResetForm
+from ..counters.exceptions import MaxQuotaReached, SkipCheck
 from ..registration import REGISTRATION_METHOD_CHOICES
-from ..utils import get_organization_radius_settings, load_model
+from ..utils import get_group_checks, get_organization_radius_settings, load_model
 from .utils import ErrorDictMixin, IDVerificationHelper
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


openwisp_radius/api/serializers.py
```diff
@@ -266,6 +270,41 @@ class Meta:
...
 
+class UserGroupCheckSerializer(serializers.ModelSerializer):
+    result = serializers.SerializerMethodField()
+
+    class Meta:
+        model = RadiusGroupCheck
+        fields = ('attribute', 'op', 'value', 'result')
+
+    def get_result(self, obj):
+        counter = app_settings.CHECK_ATTRIBUTE_COUNTERS_MAP.get(obj.attribute)
```
nemesifier, line 281: I think this could fail, can you double check if other parts of the code are handling this possible failure?

openwisp_radius/api/serializers.py
```diff
@@ -266,6 +270,41 @@ class Meta:
...
 
 
+class UserGroupCheckSerializer(serializers.ModelSerializer):
+    result = serializers.SerializerMethodField()
+
+    class Meta:
+        model = RadiusGroupCheck
+        fields = ('attribute', 'op', 'value', 'result')
+
+    def get_result(self, obj):
```
nemesifier, line 280: I think we'll reuse this logic method in different places, please double check if we can avoid duplication, I think it could be stored in the counter subdir.

openwisp_radius/tests/test_api/test_api.py
```diff
@@ -1008,3 +1008,155 @@ def test_organization_registration_enabled(self):
...
+                    'result': 0,
+                },
+            )
+            self.assertDictEqual(
+                dict(checks[1]),
+                {
+                    'attribute': 'Max-Daily-Session-Traffic',
+                    'op': ':=',
+                    'value': '3000000000',
+                    'result': 0,
```
nemesifier, line 1041: double check if this doesn't mean 0 bytes left

openwisp_radius/api/serializers.py
pandafy: We only move the following code to the counter base class 

```diff
diff --git a/openwisp_radius/counters/base.py b/openwisp_radius/counters/base.py
index 7c08ab5..2f396c6 100644
--- a/openwisp_radius/counters/base.py
+++ b/openwisp_radius/counters/base.py
@@ -126,6 +126,14 @@ class BaseCounter(ABC):
 
         return int(remaining)
 
+    def result(self):
+        try:
+            return self.check()
+        except MaxQuotaReached:
+            return '0'
+        except SkipCheck:
+            return None
+
 
 class BaseDailyCounter(BaseCounter):
     check_name = 'Max-Daily-Session'

```

We will still need to instantiate the Counter class in the serializer. That's how it is done in FreeRADIUS authorize view. 

https://github.com/openwisp/openwisp-radius/blob/90786b1a9d84d7d9121ca962de0c159e46985551/openwisp_radius/api/freeradius_views.py#L294-L335

docs/source/user/api.rst
```diff
@@ -656,6 +656,24 @@ in the URL.
...
+
+User Radius Usage
+-----------------
+
+**Requires the user auth token (Bearer Token)**.
+
+Returns the radius usage of the logged-in user and the organization specified
+in the URL.
+
+It executes the relevant RADIUS counters and returns the exhausted values.
```
nemesifier, line 669: I think "the exhausted values" may not be very clear to every user.

Something like the following may work better, feel free to improve upon my suggestion.

```suggestion
It executes the relevant RADIUS counters and returns information that allows to know how much time and/or traffic the user has left.
```

openwisp_radius/api/views.py
```diff
@@ -444,6 +447,29 @@ def get_queryset(self):
 user_accounting = UserAccountingView.as_view()
 
 
+@method_decorator(
+    name='get',
+    decorator=swagger_auto_schema(
+        operation_description="""
+        **Requires the user auth token (Bearer Token).**
+        Returns the user's RADIUS usage and limit for the organization.
+        It executes the relevant counters and returns exhausted values.
```
nemesifier, line 456: This desc has to be updated too.

openwisp_radius/api/serializers.py
```diff
@@ -266,6 +270,50 @@ class Meta:
...
+        except KeyError:
+            return None
+        else:
+            return counter.get_attribute_type()
+
+
+class UserRadiusUsageSerializer(serializers.Serializer):
+    def to_representation(self, obj):
+        organization = self.context['view'].organization
+        user_group = get_object_or_404(
```
nemesifier, line 307: What if the user has multiple groups?

nemesifier (changes requested):
Almost ready, see my comments below.

openwisp_radius/api/serializers.py
pandafy: A user will have only one group per organization.

nemesifier (approved):

PR #513, opened 2024-02-18
[fix] Fixed FreeRADIUS PostAuthView returning 500 error

### LINKED ISSUE #467: [bug] FreeRADIUS PostAuthView returns 500 error
The FreeRADIUS PostAuthView (`/api/v1/freeradius/postauth/`) returns HTTP 500 error if the length of `called_station_id` field exceeds 50 characters. 

A possible fix could be to update the PostAuthSerializer and set `max_length` attribute on the `called_station_id` field. This way, the view would return HTTP 400 with the error reason.

Shiva953:
Addresses the issue of PostAuthView returning HTTP status code 500, and sets a restriction on the server side so that it gives a client side error(400) instead.

Updates the PostAuthSerializer and sets the `max_length attribute` to be 50 on the `called_station_id` field, such that it returns HTTP Status Code 400 on `called_station_id` exceeding 50.

Fixes #467

### DIFF
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -120,7 +120,7 @@ class RadiusPostAuthSerializer(serializers.ModelSerializer):
         allow_blank=True,
         style={'input_type': 'password'},
     )
-    called_station_id = serializers.CharField(required=False, allow_blank=True)
+    called_station_id = serializers.CharField(required=False, allow_blank=True, max_length=50)
     calling_station_id = serializers.CharField(required=False, allow_blank=True)
 
     def validate(self, data):


nemesifier (changes requested):
A failing test is missing here.

Shiva953:
> A failing test is missing here.

the test shall be added in `test_freeradius_api.py`, as a `TestFreeradiusApi` class method right?

pandafy (commented):
@Shiva953 yes, you are right! I will add the test after this test case 

https://github.com/openwisp/openwisp-radius/blob/67fb11273444fe73018d7da40f139e787b26dee3/openwisp_radius/tests/test_api/test_freeradius_api.py#L586-L591

I would name the test something like `test_postauth_called_station_id_validation`. The test should ensure that upon sending a request payload which has called_station_id with more than 50 characters, a HTTP 400 response should be returned.

Shiva953:
> @Shiva953 yes, you are right! I will add the test after this test case
> 
> https://github.com/openwisp/openwisp-radius/blob/67fb11273444fe73018d7da40f139e787b26dee3/openwisp_radius/tests/test_api/test_freeradius_api.py#L586-L591
> 
> I would name the test something like `test_postauth_called_station_id_validation`. The test should ensure that upon sending a request payload which has called_station_id with more than 50 characters, a HTTP 400 response should be returned.
thanks for the clarification, just added the required failing test.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -590,6 +590,16 @@ def test_postauth_400(self):
...
 
+    def test_postauth_called_station_id_validation(self):
+        payload = {
+            'called_station_id': 'C0-4A-00-EE-D1-0D:' + 'B' * 46
+        }  # taking a >50 char value of called_station_id
+        response = self.client.post(
+            reverse('radius:postauth'), payload, HTTP_AUTHORIZATION=self.auth_header
+        )
+        self.assertEqual(RadiusPostAuth.objects.all().count(), 0)
+        self.assertEqual(response.status_code, 400)
```
nemesifier, line 601: Have you double checked that this test fails if the fix is removed?

Shiva953:
> @Shiva953 don't forget to run `./run-qa-checks` locally before pushing, the build will fail because of QA checks.

@nemesifier yes, I have confirmed the test by running `./run-qa-checks` locally:
Screenshot 2024-02-22 at 5 28 25 PM

Also, I double checked the test and re-ran it after removing the fix, and the test case failed(which is indeed the desired scenario):
Screenshot 2024-02-22 at 5 30 29 PM

nemesifier (changes requested):
The build is failing because of QA checks. You need to run:

```
openwisp-qa-check
```

Then run:

```
./run-qa-checks
```

If the exit code (`echo $?` in a linux shell) of the last command is not zero and there's any error output, don't push, fix it first.

Please read the build output, it also provides a link to our guidelines which explain all these aspects in detail, please read that page carefully.

Shiva953:
> @Shiva953 the build is failing because of QA checks. You need to run:
> 
> ```
> openwisp-qa-check
> ```
> 
> Then run:
> 
> ```
> ./run-qa-checks
> ```
> 
> If the exit code (`echo $?` in a linux shell) of the last command is not zero and there's any error output, don't push, fix it first.
> 
> Please read the build output, it also provides a link to our guidelines which explain all these aspects in detail, please read that page carefully.

Apologies for the oversight. I ran the `openwisp-qa-format` command check now and fixed the linting errors. After that, ran `openwisp-qa-check` and found out the build was failing due to formatting errors and also the commit not following the conventional guidelines, both of which are successful now.

nemesifier (changes requested):
@Shiva953 I ran the tests without the fix and the test passes, the response from the API is:

```
{'username': ['This field is required.'], 'reply': ['This field is required.']}
```

Please make sure the test replicates the bug first, and make sure the assertions check different aspects of the expected results to ensure the result is really what we expect.

Here you're checking the status code as 400, which means error on the client side, but you're not checking for specific aspects of the error response, so any error could make the test pass, which generates a false sense of security.

nemesifier:
Closing for lack of follow up.

PR #527, opened 2024-06-30
Issues/525 fix management command inconsistency

### LINKED ISSUE #525: [change] `delete_old_radiusbatch_users` management command is inconsistent
The `delete_old_radiusbatch_users` management command expects the argument to be present in months while all the other management commands requires the argument in days. 

This inconsistency may lead to misconfiguration. We shall update the `delete_old_radiusbatch_users` command to accept days instead of months.

kaushikaryan04:
As mentioned in the issue I have changed the ```delete_old_radiusbatch_users``` argument to take in days instead of months and have changed the tasks and tests accordingly. 

-----

Fixes #525


nemesifier (changes requested):
Thank you for contributing @kaushikaryan04.

The current changes woulbe backward incompatible and can break existing instances.
We should implement this change in a backward compatible way:

- We should not remove the `--older-than-months` options from the code
- We should give precedence to `--older-than-days` and use the old option only if `--older-than-days` is not present
- We should update the documentation too (do a project wide search for `delete_old_radiusbatch_users` to find it, it's in docs/source/user/management_commands.rst`

So, both `older-than-days` and `older-than-months` should be optional but we must ensure there's at least one supplied or we should return an error.

kaushikaryan04:
Due to some error I was not able to run the tests locally (working on it ) but I think this should be good. All old features are working as they were before and --older-than-days is added as well.

openwisp_radius/management/commands/base/delete_old_radiusbatch_users.py
```diff
@@ -16,15 +16,29 @@ def add_arguments(self, parser):
...
             '--older-than-days',
             action='store',
+            type=int,
+            help='Delete users that are older than days provided',
+        )
+        parser.add_argument(
+            '--older-than-months',
+            action='store',
+            type=int,
             default=BATCH_DELETE_EXPIRED,
```
nemesifier, line 26: Can we update the default so that it's used only on the new argument? The default for this old argument should be `None`, so that the new argument will be used by default.

nemesifier (changes requested):
@kaushikaryan04 looks good, I found something that doesn't convince me, please see my comment below.

openwisp_radius/tasks.py
```diff
@@ -43,9 +43,9 @@ def deactivate_expired_users():
 
 
 @shared_task
-def delete_old_radiusbatch_users(older_than_months=12):
+def delete_old_radiusbatch_users(older_than_days=365):
```
nemesifier, line 46: shouldn't the default here be the same as `BATCH_DELETE_EXPIRED`? This was probably a mistake.

openwisp_radius/tests/test_commands.py
```diff
@@ -174,7 +174,7 @@ def test_delete_old_radiusbatch_users_command(self):
         self.assertEqual(get_user_model().objects.all().count(), 6)
         call_command('delete_old_radiusbatch_users')
         self.assertEqual(get_user_model().objects.all().count(), 3)
-        call_command('delete_old_radiusbatch_users', older_than_months=12)
+        call_command('delete_old_radiusbatch_users', older_than_days=365)
```
nemesifier, line 177: We need a test that ensures the old behavior continues to work.

nemesifier (changes requested):

openwisp_radius/management/commands/base/delete_old_radiusbatch_users.py
kaushikaryan04: Yes this could be done but the action would be same if no argument is provided that is why I did not changed it.I can do this in the next commit.

openwisp_radius/tasks.py
kaushikaryan04: Sure we can change this

openwisp_radius/tests/test_commands.py
kaushikaryan04: Sure I can add some more users to test.

nemesifier (approved):

nemesifier:
Follw up:

https://github.com/openwisp/ansible-openwisp2/issues/500
https://github.com/openwisp/docker-openwisp/issues/392

PR #544, opened 2024-07-27
Issues/523 saml auth create email address objects

### LINKED ISSUE #523: [change] SAML should create EmailAddress objects
The SAML authentication of this module does not create `EmailAddress` objects, which is problematic because we are gradually changing the entire application to send notification emails to verified addresses only.

Therefore we should proceed to make sure SAML auth creates `EmailAddress` objects and always flags the email coming from SAML as verified and primary.

kaushikaryan04:
## Checklist

Closes #523 .

## Description of Changes

- Check if the NameID is an email and use it as the user's email.
- If NameID is not an email, check for the 'email' attribute in the SAML response.
- Create an EmailAddress object using the retrieved email.

### DIFF
--- a/openwisp_radius/saml/views.py
+++ b/openwisp_radius/saml/views.py
@@ -2,6 +2,7 @@ import logging
 from urllib.parse import parse_qs, quote, urlencode, urlparse
 
 import swapper
+from allauth.account.models import EmailAddress
 from allauth.account.utils import send_email_confirmation
 from django import forms
 from django.conf import settings
@@ -73,6 +74,22 @@ class AssertionConsumerServiceView(
         try:
             user.registered_user
         except ObjectDoesNotExist:
+            email = None
+            uid_is_email = 'email' in getattr(
+                settings, 'SAML_ATTRIBUTE_MAPPING', {}
+            ).get('uid', ())
+            if uid_is_email:
+                email = session_info['name_id'].text
+            if email is None:
+                email = session_info['ava'].get('email', [None])[0]
+            if email:
+                user.email = email
+                user.save()
+                email_address = EmailAddress.objects.create(
+                    user=user, email=email, verified=True, primary=True
+                )
+                email_address.save()
+
             registered_user = RegisteredUser(
                 user=user, method='saml', is_verified=app_settings.SAML_IS_VERIFIED
             )
--- a/openwisp_radius/saml/views.py
+++ b/openwisp_radius/saml/views.py
@@ -4,7 +4,6 @@ from urllib.parse import parse_qs, quote, urlencode, urlparse
 import swapper
 from allauth.account.models import EmailAddress
 from allauth.account.utils import send_email_confirmation
-from allauth.utils import valid_email_or_none
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model, logout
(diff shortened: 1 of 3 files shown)


openwisp_radius/saml/views.py
```diff
@@ -66,6 +67,22 @@ def post_login_hook(self, request, user, session_info):
...
+            uid_is_email = 'email' in getattr(
+                settings, 'SAML_ATTRIBUTE_MAPPING', {}
+            ).get('uid', ())
+            if uid_is_email:
+                email = session_info['name_id'].text
+            if email is None:
+                email = session_info['ava'].get('email', [None])[0]
+            if email:
+                user.email = email
+                user.save()
```
nemesifier, line 80: Saving this information without validating it is dangerous.

Here we need to run `full_clean()` and catch any possible exception in case the email we get from the identity provider is invalid for some reason. If the email is invalid wont save/create, log a warning and skip, eg:

```python
try:
    user.full_clean()
except ValidationError:
    logger.exception(f'Failed email validation for {user} during SAML user creation')
else:
    user.save()
    email_address = EmailAddress.objects.create(
       user=user, email=email, verified=True, primary=True
    )
```

Please add a new test for this error case, you can use `mock` or any other way to pass an invalid email and catch the validation error, there's similar tests in this and other modules you can base your test on, look for `ValidationError` in the test suites of the codebase.

openwisp_radius/saml/views.py
```diff
@@ -66,6 +67,22 @@ def post_login_hook(self, request, user, session_info):
...
+                email_address = EmailAddress.objects.create(
+                    user=user, email=email, verified=True, primary=True
+                )
+                email_address.save()
```
nemesifier, line 84: Remove:
```suggestion
```

nemesifier (changes requested):
Thank you very much for your patience @kaushikaryan04.
I've been super busy publishing the new [unified documentation website of OpenWISP](https://openwisp.io/docs/dev/), now I am back reviewing PRs.

This looks almost ready to merge, but there's a bit of validation we have to do before saving data got from external sources to the DB.

Please see my comments below.

openwisp_radius/saml/views.py
kaushikaryan04: No problem.The new unified documentation looks great. I will make the suggested changes.

openwisp_radius/saml/views.py
```diff
@@ -66,6 +68,37 @@ def post_login_hook(self, request, user, session_info):
         try:
             user.registered_user
         except ObjectDoesNotExist:
+            email = None
```
pandafy, line 71: Instead of retrieving the email address from the SAML request, can we use the `user.email` field to get the email address? The field should get automatically populated by djangosaml2 using attribute mapping. Then, we can create the EmailAddress object with that value.

pandafy (changes requested):

nemesifier (dismissed):

nemesifier (approved):

nemesifier:
Follow up:

- https://github.com/openwisp/ansible-openwisp2/issues/500
- https://github.com/openwisp/docker-openwisp/issues/392

Issue #229, opened 2021-01-27
Allow IP range to for allowed freeradius hosts


atb00ker:
Currently, we can only give IP address of freeradius server allowed to interact with openwisp-radius by using the [OPENWISP_RADIUS_FREERADIUS_ALLOWED_HOSTS](https://openwisp-radius.readthedocs.io/en/latest/user/settings.html#openwisp-radius-freeradius-allowed-hosts) option but this is not sufficient if the IP is not fixed, example in docker-openwisp, the IP address can be in range "172.18.0.0/16".

Hence, we need to allow user to enter IP range as well.

Ki1er:
working on it

Ki1er:
can you assign it to me @atb00ker

UltraBot05:
Hello @atb00ker, @nemesifier. I'd like to take on this issue.

---

___Proposed Fix:___
I will implement CIDR range support for `FREERADIUS_ALLOWED_HOSTS` by utilizing Python's native `ipaddress` module, which allows this project to support _Individual IPs_ and _Ranges (something like 172.18.0.0/16)_ for authorized `FreeRADIUS hosts`.

PR #556, opened 2024-10-19
Fix: Enable IP range configuration for FreeRADIUS API access

Sachin-NK:
Related to issue #229  Allow IP range to for allowed freeradius hosts 

### Description:

This pull request updates the system to support IP address ranges (using CIDR notation) for the OPENWISP_RADIUS_FREERADIUS_ALLOWED_HOSTS setting and the organization's RADIUS settings. Previously, it only allowed individual IP addresses, which was a limitation in environments where the FreeRADIUS server’s IP address could change.

Now, administrators can specify a range of allowed IP addresses instead of just single IPs, making the configuration more flexible and reliable. The update ensures that the system correctly handles both individual IPs and ranges during authentication and adds better error handling for any incorrect IP range settings.

### DIFF
--- a/openwisp_radius/api/freeradius_views.py
+++ b/openwisp_radius/api/freeradius_views.py
@@ -97,16 +97,20 @@ class FreeradiusApiAuthentication(BaseAuthentication):
         ip_list = self._get_ip_list(uuid)
 
         for ip in ip_list:
-            try:
-                if ipaddress.ip_address(client_ip) in ipaddress.ip_network(ip):
-                    return (AnonymousUser(), uuid)
-            except ValueError:
-                invalid_addr_message = _(
-                    'Request rejected: ({ip}) in organization settings or '
-                    'settings.py is not a valid IP address. '
-                    'Please contact administrator.'
-                ).format(ip=ip)
-                raise AuthenticationFailed(invalid_addr_message)
+            if '/' in ip:  # Check for CIDR notation (IP range)
+                try:
+                    if ipaddress.ip_address(client_ip) in ipaddress.ip_network(ip):
+                        return (AnonymousUser(), uuid)
+                except ValueError as e:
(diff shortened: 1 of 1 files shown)


openwisp_radius/api/freeradius_views.py
```diff
@@ -97,16 +97,20 @@ def _check_client_ip_and_return(self, request, uuid):
...
-                if ipaddress.ip_address(client_ip) in ipaddress.ip_network(ip):
-                    return (AnonymousUser(), uuid)
-            except ValueError:
-                invalid_addr_message = _(
-                    'Request rejected: ({ip}) in organization settings or '
-                    'settings.py is not a valid IP address. '
-                    'Please contact administrator.'
-                ).format(ip=ip)
-                raise AuthenticationFailed(invalid_addr_message)
+            if '/' in ip:  # Check for CIDR notation (IP range)
```
nemesifier, line 100: I am not a fan of this approach.

See the following:

```
In [2]: ipaddress.ip_network('10.0.0.1')
Out[2]: IPv4Network('10.0.0.1/32')

In [3]: ipaddress.ip_network('fe80::714:f09:20f3:d39d')
Out[3]: IPv6Network('fe80::714:f09:20f3:d39d/128')
```

As you can see, initialitizing ip_network() with a single IP is possible and yields a /32 on ipv4 or a /128 on ipv6.

So:

```
In [4]: ipaddress.ip_address('10.0.0.1') in ipaddress.ip_network('10.0.0.1')
Out[4]: True
```

Would work. So I am not sure what problem your this is really solving. To me it looks like this code isn't implementing anything which the current code doesn't already do.

nemesifier (changes requested):
Thanks for contributing @Ki1er. 

I don't think this solution can be accepted as of now, see below for more information.

PR #557, opened 2024-10-25
[feature] Added view to collect email in SAML registration

pandafy:
## Checklist

## Description of Changes

If the IdP does not include email address in SAML response, then the user will be asked for email during sign-up.

### DIFF
--- a/openwisp_radius/locale/de/LC_MESSAGES/django.po
+++ b/openwisp_radius/locale/de/LC_MESSAGES/django.po
@@ -158,10 +158,15 @@ msgstr "Fügen Sie Ihre E-Mail hinzu, um die Einrichtung Ihres Kontos abzuschlie
 
 #: templates/djangosaml2/login_additional_info.html:6
 msgid ""
-"It looks like we didn't receive your email from your identity provider.</"
-"br>Please enter your email address below to complete setting up your account."
+"It looks like we didn't receive your email from your identity provider."
 msgstr ""
-"Es sieht so aus, als hätten wir Ihre E-Mail von Ihrem Identitätsanbieter nicht erhalten.<br>Bitte geben Sie Ihre E-Mail-Adresse unten ein, um die Einrichtung Ihres Kontos abzuschließen."
+"Es scheint, dass wir Ihre E-Mail-Adresse von Ihrem Identitätsanbieter nicht erhalten haben."
+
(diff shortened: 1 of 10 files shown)


openwisp_radius/saml/urls.py
```diff
@@ -26,6 +26,11 @@ def get_saml_urls(saml_views=None):
             saml_views.AssertionConsumerServiceView.as_view(),
             name='saml2_acs',
         ),
+        path(
+            'additional-info/',
+            saml_views.LoginAdditionalInfoView.as_view(),
```
nemesifier, line 31: I don't love the naming here, if we are asking only for the email here, why are we calling everything "additional info"?
I think we should call this view `LoginCollectEmailView`, the URL `collect-email/` and the template `collect-email.html`.

nemesifier (commented):
All looks good except the naming, see my comment below and let me know what

openwisp_radius/saml/urls.py
pandafy: I used a generic name in-case we used this view to ask for additional information. E.g., we could ask the user to verify the name received from the IdP.

nemesifier (changes requested):
Is the `fur` translation missing?

openwisp_radius/locale/de/LC_MESSAGES/django.po
```diff
@@ -151,3 +151,22 @@ msgstr "Session verwalten"
...
 msgstr "Neue Session WiFi gestartet"
+
+#: templates/djangosaml2/login_additional_info.html:5
+msgid "Add Your Email to Complete Your Account Setup"
+msgstr "Fügen Sie Ihre E-Mail hinzu, um die Einrichtung Ihres Kontos abzuschließen"
+
+#: templates/djangosaml2/login_additional_info.html:6
+msgid ""
+"It looks like we didn't receive your email from your identity provider.</"
+"br>Please enter your email address below to complete setting up your account."
```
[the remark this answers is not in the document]
pandafy: I can try splitting the string in two different paragraphs. 

This [old StackOverflow post](https://stackoverflow.com/questions/3397218/translating-text-blocks-with-django-what-to-do-with-the-html) says that it's okay to have HTML tag in the string.

nemesifier (approved):

PR #602, opened 2025-06-09
[change] Added CSV batch download endpoint visibility on Swagger

### LINKED ISSUE #373: [change] Download batch CSV URL is inconsistent and not showing up in API docs
The URL generated with `radius:serve_private_file` is not consistent with the rest of the API URLs:

```
./manage.py show_urls | grep batch | grep api
/api/v1/radius/batch/	openwisp_radius.api.views.BatchView	radius:batch
/api/v1/radius/organization/<slug:slug>/batch/<uuid:pk>/pdf/	openwisp_radius.api.views.DownloadRadiusBatchPdfView	radius:download_rad_batch_pdf
/api/v1/radiusbatch/csv/<path:csvfile>	openwisp_radius.private_storage.views.RadiusBatchCsvDownloadView	radius:serve_private_file
```

Shouldn't this CSV url be as follows?

`/api/v1/radius/organization/<slug:slug>/batch/<uuid:pk>/csv/`.

We have an inconsistency here, I will create a new issue for this, another problem is that this URL does not show up in the swagger API docs (`/api/v1/docs/`), let's find way to show it there too.

This is a follow up of https://github.com/openwisp/openwisp-radius/pull/366#pullrequestreview-868237372.

**Further follow up**: https://github.com/openwisp/openwisp-radius/commit/0e764626e80e19c0b9682ed07d61cfcccc368835 fixes the inconsistencies, but the URL is not showing up in the API docs, it would be great it we can add it in some way, for consistency, so I will leave this open.

stktyagi:
## Checklist

Closes #373.

## Description of Changes

Integrated PrivateStorageDetailView within a DRF APIView and applied @extend_schema to enable proper Swagger documentation. This update ensures the CSV download endpoint is both functional and visible in the Swagger UI. Verified the endpoint works correctly and appears as expected in the API docs.

### DIFF
--- a/openwisp_radius/private_storage/urls.py
+++ b/openwisp_radius/private_storage/urls.py
@@ -16,5 +16,13 @@ def get_private_store_urls():
             urljoin(app_settings.CSV_URL_PATH, "<path:path>"),
             views.rad_batch_csv_download_view,
             name="serve_private_file",
-        )
+        ),
+        path(
+            urljoin(
+                app_settings.CSV_URL_PATH,
+                "<slug:slug>/batch/<uuid:pk>/csv/<str:filename>/",
+            ),
+            views.RadiusBatchCsvDownloadAPIView.as_view(),
+            name="radius_organization_batch_csv_read",
+        ),
     ]
--- a/openwisp_radius/private_storage/views.py
+++ b/openwisp_radius/private_storage/views.py
@@ -12,7 +12,7 @@ class RadiusBatchCsvDownloadView(PrivateStorageDetailView):
     model = load_model("RadiusBatch")
     model_file_field = "csvfile"
     slug_field = "csvfile"
-    slug_url_kwarg = "filename"
+    slug_url_kwarg = "path"
 
     def can_access_file(self, private_file):
         user = private_file.request.user
--- a/openwisp_radius/private_storage/views.py
+++ b/openwisp_radius/private_storage/views.py
@@ -1,4 +1,7 @@
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import OpenApiParameter, OpenApiResponse, extend_schema
 from private_storage.views import PrivateStorageDetailView
+from rest_framework.views import APIView
 
 from ..settings import PRIVATE_STORAGE_INSTANCE
 from ..utils import load_model
@@ -9,7 +12,7 @@ class RadiusBatchCsvDownloadView(PrivateStorageDetailView):
     model = load_model("RadiusBatch")
     model_file_field = "csvfile"
     slug_field = "csvfile"
-    slug_url_kwarg = "path"
+    slug_url_kwarg = "filename"
 
     def can_access_file(self, private_file):
         user = private_file.request.user
@@ -19,3 +22,38 @@ class RadiusBatchCsvDownloadView(PrivateStorageDetailView):
 
 
 rad_batch_csv_download_view = RadiusBatchCsvDownloadView.as_view()
+
+
+# this is a drf wrapper view
+@extend_schema(
+    operation_id="radius_organization_batch_csv_read",
+    description="Download the CSV export file for a specific RADIUS batch.",
+    parameters=[
+        OpenApiParameter(
+            name="slug",
+            type=str,
+            location=OpenApiParameter.PATH,
+            description="Organization slug",
+        ),
+        OpenApiParameter(
+            name="pk",
+            type=str,
+            location=OpenApiParameter.PATH,
+            description="Batch UUID",
+        ),
+        OpenApiParameter(
+            name="filename",
+            type=str,
+            location=OpenApiParameter.PATH,
+            description="CSV filename",
+        ),
+    ],
+    responses={
+        200: OpenApiResponse(description="CSV file", response=OpenApiTypes.BINARY)
+    },
+    tags=["radius"],
+)
+class RadiusBatchCsvDownloadAPIView(APIView):
+    def get(self, request, pk, filename):
+        view = RadiusBatchCsvDownloadView.as_view()
+        return view(request, pk=pk, filename=filename)


openwisp_radius/private_storage/views.py
```diff
@@ -19,3 +22,38 @@ def can_access_file(self, private_file):
 
 
 rad_batch_csv_download_view = RadiusBatchCsvDownloadView.as_view()
+
+
+# this is a drf wrapper view
+@extend_schema(
```
nemesifier: It should not be needed to add a new dependency to do this, please refer to https://github.com/openwisp/openwisp-controller/pull/1086/files#diff-3e530f9831121ff0f6b2c2d5a9dcb9ffe54d1bba1a1e3629095b50bce9521f88R79-R91.

nemesifier (changes requested):

openwisp_radius/private_storage/views.py
stktyagi: Switched the wrapper view to drf-yasg and removed the drf-spectacular dependency, aligning with the implementation in openwisp-controller. The CSV download endpoint is now properly documented in Swagger, no extra dependencies required. Tested locally and fixed URL ordering to prevent the private storage view from being shadowed.

The CI failure is due to:-
```ERROR: ReStructuredText check failed!
File '/home/runner/work/openwisp-radius/openwisp-radius/docs/user/rest-api.rst' could be reformatted.
```
which I've addressed in #666

openwisp_radius/private_storage/urls.py
```diff
@@ -8,6 +8,14 @@
 
 def get_private_store_urls():
     return [
+        path(
+            urljoin(
+                app_settings.CSV_URL_PATH,
+                "<slug:slug>/batch/<uuid:pk>/csv/<str:filename>/",
+            ),
+            views.RadiusBatchCsvDownloadAPIView.as_view(),
```
nemesifier: For consistency with the rest of the code can you move this to the view file?

openwisp_radius/private_storage/views.py
```diff
@@ -19,3 +23,28 @@ def can_access_file(self, private_file):
...
+        ],
+        responses={
+            200: openapi.Response(
+                description=_("CSV file"), schema=openapi.Schema(type=openapi.TYPE_FILE)
+            ),
+        },
+        tags=["radius"],
+    )
+    def get(self, request, slug, pk, filename, *args, **kwargs):
+        return rad_batch_csv_download_view(request, pk=pk, filename=filename, **kwargs)
```
nemesifier:
```suggestion
        return rad_batch_csv_download_view(request, pk=pk, **kwargs)


rad_batch_csv_download_api_view = RadiusBatchCsvDownloadAPIView.as_view()
```

docs/user/rest-api.rst
```diff
@@ -861,16 +861,25 @@ Batch CSV Download
 
 .. code-block:: text
 
-    /api/v1/radius/organization/<organization-slug>/batch/<id>/csv/<filename>
+    /api/v1/radius/organization/<organization-slug>/batch/<id>/csv/<filename>/
 
-Responds only to **GET**.
+Responds only to **GET**. Allows downloading the CSV export file for a
```
nemesifier: Allows downloading the CSV file used to import users for a specific batch user creation operation.

docs/user/rest-api.rst
```diff
@@ -861,16 +861,25 @@ Batch CSV Download
 
 .. code-block:: text
 
-    /api/v1/radius/organization/<organization-slug>/batch/<id>/csv/<filename>
+    /api/v1/radius/organization/<organization-slug>/batch/<id>/csv/<filename>/
```
nemesifier: Actually I reviewed the original issue I had creted, and the last `filename/` part is not there. I think we can simplify this to avoid requiring the filename (it's not needed).

```suggestion
    /api/v1/radius/organization/<organization-slug>/batch/<id>/csv/
```

Unfortunately it will be a breaking change but I don' think anyone is using this API endpoint anyway because in its current form it's pretty unusable as I don't think it's easiy to know the CSV filename so the right URL to be queried cannot be constructed easily.

docs/user/rest-api.rst
```diff
@@ -861,16 +861,25 @@ Batch CSV Download
...
+id       string (batch UUID)
+filename string (name of the CSV file)
+======== =============================
+
+Example:
+
+.. code-block:: shell
+
+    curl -X GET \
+        'http://127.0.0.1:8000/api/v1/radius/organization/default/batch/f4943c8a-462e-40ba-89b6-91a2541c9cf4/csv/test.csv/' \
```
nemesifier:
```suggestion
        'http://127.0.0.1:8000/api/v1/radius/organization/default/batch/f4943c8a-462e-40ba-89b6-91a2541c9cf4/csv/' \
```

openwisp_radius/private_storage/urls.py
```diff
@@ -8,6 +8,14 @@
 
 def get_private_store_urls():
     return [
+        path(
+            urljoin(
+                app_settings.CSV_URL_PATH,
+                "<slug:slug>/batch/<uuid:pk>/csv/<str:filename>/",
```
nemesifier:
```suggestion
                "<slug:slug>/batch/<uuid:pk>/csv/",
```

openwisp_radius/private_storage/views.py
```diff
@@ -19,3 +23,28 @@ def can_access_file(self, private_file):
 
 
 rad_batch_csv_download_view = RadiusBatchCsvDownloadView.as_view()
+
+
+class RadiusBatchCsvDownloadAPIView(APIView):
+    @swagger_auto_schema(
+        operation_id="radius_organization_batch_csv_read",
+        operation_description=_(
+            "Download the CSV export file for a specific RADIUS batch."
```
nemesifier: Allows downloading the CSV file used to import users for a specific batch user creation operation.

docs/user/rest-api.rst
```diff
@@ -861,16 +861,25 @@ Batch CSV Download
...
 Param    Description
-======== ===========
-slug     string
-id       string
-filename string
-======== ===========
+======== =============================
+slug     string (organization slug)
+id       string (batch UUID)
+filename string (name of the CSV file)
```
nemesifier: Being this a GET API URL, I think we can remove this table altogether and leave the curl example below which is more useful.

nemesifier (changes requested):
Thanks for this @stktyagi, it looks good but needs improvement, please see my comments below.

nemesifier (approved):
LGTM, I will do another round of testing once the CI build passes and if I don't see anything wrong I will merge, thanks 🙏

nemesifier (approved):

PR #603, opened 2025-06-09
[fix] Making batch imports atomic #551

### LINKED ISSUE #551: [change] Make batch import atomic
I was reviewing the batch CSV import code due to #550 and it seems to me the create operation is not atomic.

What if the import operation fails in the middle?

I believe we would end up with half users created.

I think it's better to make everything run in a transaction and if anything fails rollback.

stktyagi:
Used transaction.atomic() to ensure all operations inside succeed together. Incase of failure rollback will occure.

Fixes #551

## Checklist

Closes #551.

## Description of Changes

Added usage of transaction.atomic() functions to introduce atomicity and rollback in case of transaction operation failure.

### DIFF
--- a/openwisp_radius/tests/test_batch_add_users.py
+++ b/openwisp_radius/tests/test_batch_add_users.py
@@ -182,6 +182,10 @@ class TestBatchAtomicity(FileMixin, BaseTransactionTestCase):
         with self.assertRaises(IntegrityError):
             batch.add(data)
         self.assertFalse(User.objects.filter(username="user_one").exists())
+        self.assertTrue(
+            RadiusBatch.objects.filter(name="atomic-integrity-test").exists()
+        )
+        self.assertEqual(batch.users.count(), 0)
 
 
 del BaseTestCase


nemesifier (commented):
In #551 I wrote:

What if the import operation fails in the middle?
I believe we would end up with half users created.

Have you tried making the import fail on purpose in the middle to see what happens?
Can we recreate this in the tests?

Example from chatgpt to raise an error the second time the save method is called:

```python
from unittest.mock import patch
from django.db import DatabaseError
from myapp.models import MyModel

def fail_on_second_call(real_method):
    calls = [True, False]  # First call real, second call raises

    def wrapper(self, *args, **kwargs):
        if calls.pop(0):
            return real_method(self, *args, **kwargs)
        else:
            raise DatabaseError("Simulated DB failure on second save")

    return wrapper

with patch.object(MyModel, 'save', autospec=True) as mocked_save:
    # Wrap original method
    mocked_save.side_effect = fail_on_second_call(MyModel.save)

    obj = MyModel(field="test")
    obj.save()  # ✅ Real save

    obj.name = "test2"
    try:
        obj.save()  # ❌ Fails with simulated DatabaseError
    except DatabaseError as e:
        print(f"Caught simulated error: {e}")
```

We could do something like this to cause a failure the second time the `save_user()` method is called. Without the patch, I expect that the save operations wouldn't be rolledback, and the first user created during the import would remain in the DB.

With the fix, that shouldn't happen and any imported data should be automatically rolledback.

Just adding the `transaction.atomic()` block could solve it, but it doesn't give us any confidence that it's really working, let's validate our assumptions.

nemesifier:
Closing due to lack of follow up, feel free to follow up and reopen this.

stktyagi:
> In #551 I wrote:
> 
> What if the import operation fails in the middle? I believe we would end up with half users created.
> 
> Have you tried making the import fail on purpose in the middle to see what happens? Can we recreate this in the tests?
> 

I've implemented 2 tests that check if both the add() and csvfile_upload() functions are truly atomic and would fail without `transaction.atomic()`.

> Example from chatgpt to raise an error the second time the save method is called:
> 
> ```python
> from unittest.mock import patch
> from django.db import DatabaseError
> from myapp.models import MyModel
> 
> def fail_on_second_call(real_method):
>     calls = [True, False]  # First call real, second call raises
> 
>     def wrapper(self, *args, **kwargs):
>         if calls.pop(0):
>             return real_method(self, *args, **kwargs)
>         else:
>             raise DatabaseError("Simulated DB failure on second save")
> 
>     return wrapper
> 
> with patch.object(MyModel, 'save', autospec=True) as mocked_save:
>     # Wrap original method
>     mocked_save.side_effect = fail_on_second_call(MyModel.save)
> 
>     obj = MyModel(field="test")
>     obj.save()  # ✅ Real save
> 
>     obj.name = "test2"
>     try:
>         obj.save()  # ❌ Fails with simulated DatabaseError
>     except DatabaseError as e:
>         print(f"Caught simulated error: {e}")
> ```
> 
> We could do something like this to cause a failure the second time the `save_user()` method is called. Without the patch, I expect that the save operations wouldn't be rolledback, and the first user created during the import would remain in the DB.
> 
> With the fix, that shouldn't happen and any imported data should be automatically rolledback.
> 
> Just adding the `transaction.atomic()` block could solve it, but it doesn't give us any confidence that it's really working, let's validate our assumptions.

Athough making a test for failure for this was complicated as standard test tools hide the very database behavior needed to observe, in the test I am triggering a database IntegrityError via duplicate email constraints. This verifies that a failure midway through the import loop correctly triggers a ROLLBACK of both the partially created users and the RadiusBatch metadata record.

openwisp_radius/tests/test_batch_add_users.py
```diff
@@ -145,5 +147,43 @@ def test_verified_batch_user_creation(self):
...
+        ]
+        batch = RadiusBatch(
+            name="total-rollback",
+            strategy="csv",
+            organization=org,
+            csvfile=self._get_csvfile(data),
+        )
+        with self.assertRaises(IntegrityError):
+            batch.csvfile_upload()
+
```
nemesifier:
```suggestion
```

nemesifier (approved):

PR #619, opened 2025-08-10
[change:radius] Replace third-party JSONField with Django built-in JSONField

### LINKED ISSUE #600: [change:radius] Replace thirdparty JSONField with Django built in JSONField
The JSONField we currently use is just a text field with JSON validation and formatting, which has served us pretty well but hasn't received updates in 5 years: https://github.com/rpkilby/jsonfield/.

The Django [JSONField](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.JSONField) has been around for some time and seems supported on all the databases we use, hence can switch to this field.

We'll need to pay particular attention to ensuring data is automatically migrated without loss.

Eeshu-Yadav:
## Checklist

Closes #600.

## Description of Changes

This PR replaces the third-party `jsonfield` library with Django's built-in JSONField. The third-party JSONField library hasn't received updates in 5 years, while Django's built-in JSONField is now mature and actively maintained.

### Changes Made:
- Replaced `from jsonfield import JSONField` with `from django.db.models import JSONField` in models
- Updated `RadiusBatch.user_credentials` and `OrganizationRadiusSettings.sms_meta_data` fields
- Removed `"jsonfield~=3.1.0"` dependency from setup.py
- Added migration `0041_replace_jsonfield_with_django_builtin.py` to handle field type transition
- Updated test migrations to use Django's JSONField

### Testing:
- All existing tests pass with the new implementation
- Verified RadiusBatch and SMS metadata functionality
- Migration tested without data loss


openwisp_radius/migrations/0043_alter_organizationradiussettings_sms_meta_data_and_more.py
```diff
@@ -0,0 +1,52 @@
...
+                null=True,
+                verbose_name="SMS meta data",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="radiusbatch",
+            name="user_credentials",
+            field=models.JSONField(blank=True, null=True, verbose_name="PDF"),
+        ),
+        migrations.RunPython(
```
nemesifier: Data migrations need to be executed in a dedicated migration file, please refer to https://stackoverflow.com/a/66799446 for more information.

openwisp_radius/migrations/0043_alter_organizationradiussettings_sms_meta_data_and_more.py
```diff
@@ -0,0 +1,52 @@
...
+    to proper JSON objects for Django's built-in JSONField.
+    """
+    RadiusBatch = apps.get_model("openwisp_radius", "RadiusBatch")
+    for batch in RadiusBatch.objects.exclude(user_credentials__isnull=True):
+        if isinstance(batch.user_credentials, str):
+            try:
+                batch.user_credentials = json.loads(batch.user_credentials)
+                batch.save(update_fields=["user_credentials"])
+            except (json.JSONDecodeError, TypeError):
+                # If it's already decoded or invalid JSON, skip it
```
nemesifier: It shouldn't fail. If it fails we must let it fail loudly, unless there can be failure which would be common and in that case we should catch the exception and resolve this automatically.

Why do you think it would fail here?

openwisp_radius/migrations/0043_alter_organizationradiussettings_sms_meta_data_and_more.py
```diff
@@ -0,0 +1,52 @@
...
+from django.db import migrations, models
+
+
+def convert_user_credentials_data(apps, schema_editor):
+    """
+    Convert existing double-encoded JSON strings in user_credentials field
+    to proper JSON objects for Django's built-in JSONField.
+    """
+    RadiusBatch = apps.get_model("openwisp_radius", "RadiusBatch")
+    for batch in RadiusBatch.objects.exclude(user_credentials__isnull=True):
```
nemesifier:
```suggestion
    for batch in RadiusBatch.objects.exclude(user_credentials__isnull=True).iterator():
```

Not using `iterator()` may crash instances with large amount of rows and must be avoided.

nemesifier (changes requested):
Thanks @Eeshu-Yadav, see my comments below.

Eeshu-Yadav:
@nemesifier kindly review

openwisp_radius/migrations/0044_convert_user_credentials_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+def convert_user_credentials_data(apps, schema_editor):
+    """
+    Convert existing double-encoded JSON strings in user_credentials field
+    to proper JSON objects for Django's built-in JSONField.
+    """
+    RadiusBatch = apps.get_model("openwisp_radius", "RadiusBatch")
+    for batch in RadiusBatch.objects.exclude(user_credentials__isnull=True).iterator():
+        if isinstance(batch.user_credentials, str):
+            batch.user_credentials = json.loads(batch.user_credentials)
```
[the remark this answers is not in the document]
Eeshu-Yadav: @nemesifier  previously requested that this migration "fail loudly" on invalid data rather than silently skip errors, to ensure data integrity issues are caught and investigated rather than hidden. If empty strings or invalid JSON turn out to be common during testing, we can address them explicitly with proper handling.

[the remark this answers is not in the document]
nemesifier: Ok but this is not the right way of doing this, we should:

try/except and if the exception is reached: call `log.exception()` and print an error message with the exception message, eg:

`print("Encountered error while processing {batch}: {e}")`

openwisp_radius/base/models.py
```diff
@@ -18,12 +17,11 @@
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.mail import send_mail
 from django.db import models, transaction
-from django.db.models import ProtectedError, Q
+from django.db.models import JSONField, ProtectedError, Q
```
nemesifier, line 21: Please add the encoder parameter as done here:
https://github.com/openwisp/openwisp-notifications/commit/bad5232c983b4c2b66c11a9a30dbf175568b4c90#diff-ab7047dc14d6cae284453bf3f59fa7be6c346cb3a4c510176645da8d0ebf8241R94

nemesifier (changes requested):

openwisp_radius/base/models.py
Eeshu-Yadav: okk , updated that

nemesifier (changes requested):
@Eeshu-Yadav is there a reason for which the jsonfield python dependency is not being removed?

Eeshu-Yadav:
> @Eeshu-Yadav is there a reason for which the jsonfield python dependency is not being removed?

updated , kindly review

nemesifier (approved):

PR #621, opened 2025-08-10
[feature] Handle stale sessions using Accounting-On packets

### LINKED ISSUE #617: [feature] Handle stale sessions using Accounting-On packets (Simultaneous-Use issue)
**Description:**

I encountered a situation where I couldn't log in due to the Simultaneous-Use check, even though the session was from the previous day. This likely happened because the device rebooted, which caused the session to remain marked as active.

While this isn't expected to happen frequently, it can be problematic in real-world scenarios where devices may experience unexpected power outages. In such cases, stale sessions may remain open, blocking legitimate new logins due to the Simultaneous-Use restriction.

**Proposed solution:**

We should listen for Accounting-On requests and, when they are received, check if there are any open sessions associated with the corresponding NAS. If any are found, they should be flagged as closed.

This would help prevent stale sessions from lingering after a reboot or power outage and improve the reliability of the Simultaneous-Use feature.

Eeshu-Yadav:
## Checklist

Closes #617.

#### Core Changes
1. **Enable Accounting-On Support**: Removed "Accounting-On" from `UNSUPPORTED_STATUS_TYPES` and updated status type choices
2. **Session Cleanup Logic**: Added `_handle_accounting_on()` method that:
   - Extracts NAS IP address from the packet
   - Finds all open sessions for that NAS within the same organization
   - Closes sessions with proper metadata (`terminate_cause="NAS-Reboot"`)
   - Logs the number of sessions closed for visibility

#### Security & Isolation
- **Organization Boundaries**: Only affects sessions within the same organization as the requesting NAS
- **Selective Closure**: Only closes sessions specifically for the NAS that sent the Accounting-On packet
- **Proper Metadata**: Sessions are closed with appropriate timestamps and termination cause

#### Configuration
- New setting: `ACCOUNTING_ON_STALE_SESSION_TIMEOUT` (default: 24 hours) - prepared for future sophisticated timeout handling
- Fully backwards compatible - no breaking changes

### Testing Coverage
- **Updated Tests**: Modified existing Accounting-On tests to expect new behavior
- **Edge Cases**: Added comprehensive tests covering:
  - Missing `nas_ip_address` in packets
  - No open sessions scenario
  - Multi-organization isolation verification
  - Basic functionality with multiple sessions
- **Test Results**: All 102 freeradius API tests pass, including 39 accounting-specific tests

### Files Modified
- `openwisp_radius/api/freeradius_views.py` - Core implementation
- `openwisp_radius/api/serializers.py` - Status type updates  
- `openwisp_radius/settings.py` - New configuration setting
- `openwisp_radius/tests/test_api/test_freeradius_api.py` - Test updates and additions

### Usage Example
When a NAS device at IP `192.168.1.1` restarts and sends an Accounting-On packet:
1. OpenWISP receives the packet and detects `status_type="Accounting-On"`
2. All open sessions for NAS `192.168.1.1` in the requesting organization are found
3. Sessions are closed with `terminate_cause="NAS-Reboot"` and current timestamp
4. Log message reports: `"Accounting-On: Closed 3 stale sessions for NAS 192.168.1.1 in organization abc123"`
5. Users can now authenticate normally without "session already active" errors

### DIFF
--- a/openwisp_radius/api/freeradius_views.py
+++ b/openwisp_radius/api/freeradius_views.py
@@ -506,18 +506,7 @@ class AccountingView(ListCreateAPIView):
         if status_type in UNSUPPORTED_STATUS_TYPES:
             return Response(None)
         if status_type == "Accounting-On":
-            called_station_id = data.get("called_station_id")
-            closed_count = RadiusAccounting.close_stale_sessions_by_called_station_id(
-                called_station_id=called_station_id,
-                organization_id=self.request.auth,
-            )
-            if closed_count:
-                logger.info(
-                    f"Closed {closed_count} stale session(s) for device with "
-                    f"Called-Station-Id {called_station_id} in organization "
-                    f"{self.request.auth} due to receiving an "
-                    "Accounting-On packet."
-                )
+            self._handle_accounting_on(data)
             return Response(None)
         # Create or Update
         try:
@@ -547,6 +536,27 @@ class AccountingView(ListCreateAPIView):
             serializer.update(instance, acct_data)
             self.send_radius_accounting_signal(serializer.validated_data)
             return Response(None)
+
+    def _handle_accounting_on(self, data):
+        """
+        When NAS devices are cold rebooted,
+        they may not close active sessions,
+        which can cause issues with Simultaneous-Use.
+        For this reason, OpenWISP closes any open session from
+        the same NAS when it receives an Accounting-On request.
+        """
+        called_station_id = data.get("called_station_id")
+        closed_count = RadiusAccounting._close_stale_sessions_on_nas_boot(
+            called_station_id=called_station_id,
+            organization_id=self.request.auth,
+        )
+        if closed_count:
+            logger.info(
+                f"Closed {closed_count} stale session(s) for device with "
+                f"Called-Station-Id {called_station_id} in organization "
+                f"{self.request.auth} due to receiving an "
+                "Accounting-On packet."
+            )
 
     def _is_interim_update_corner_case(self, error, data):
         """
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -558,9 +558,10 @@ class AbstractRadiusAccounting(OrgMixin, models.Model):
             session.save()
 
     @classmethod
-    def close_stale_sessions_by_called_station_id(
-        cls, called_station_id, organization_id
-    ):
+    def _close_stale_sessions_on_nas_boot(cls, called_station_id, organization_id):
+        """
+        Called during RADIUS Accounting-On.
+        """
         if not called_station_id or not organization_id:
             return 0
         stale_sessions = cls.objects.filter(
--- a/openwisp_radius/tests/test_api/test_freeradius_api.py
+++ b/openwisp_radius/tests/test_api/test_freeradius_api.py
@@ -1294,7 +1294,6 @@ class TestFreeradiusApi(AcctMixin, ApiTokenMixin, BaseTestCase):
         )
         stale_session = self._create_radius_accounting(**stale_session_data)
         self.assertIsNone(stale_session.stop_time)
-
         other_user = self._create_user(username="other_user", email="other@user.com")
         self._create_org_user(user=other_user)
         other_session_data = self.acct_post_data.copy()
@@ -1308,9 +1307,7 @@ class TestFreeradiusApi(AcctMixin, ApiTokenMixin, BaseTestCase):
         )
         other_session = self._create_radius_accounting(**other_session_data)
         self.assertIsNone(other_session.stop_time)
-
         self.assertEqual(RadiusAccounting.objects.count(), 2)
-
         # Simulate Accounting-On packet from the first NAS
         accounting_on_data = {
             "status_type": "Accounting-On",
@@ -1318,16 +1315,13 @@ class TestFreeradiusApi(AcctMixin, ApiTokenMixin, BaseTestCase):
             "unique_id": "accounting-on-packet-id",
         }
         response = self.post_json(accounting_on_data)
-
         self.assertEqual(response.status_code, 200)
         self.assertIsNone(response.data)
-
         self.assertEqual(RadiusAccounting.objects.count(), 2)
-
         stale_session.refresh_from_db()
         self.assertIsNotNone(stale_session.stop_time)
         self.assertEqual(stale_session.terminate_cause, "NAS-Reboot")
-
+        # Session from other NAS unaffected
         other_session.refresh_from_db()
         self.assertIsNone(other_session.stop_time)
 


openwisp_radius/api/freeradius_views.py
```diff
@@ -454,13 +455,20 @@ def post(self, request, *args, **kwargs):
...
         processing the response without generating warnings
         """
         if request.user.is_anonymous and request.auth is None:
             return Response(status=status.HTTP_200_OK)
         data = request.data.copy()
-        if data.get("status_type", None) in UNSUPPORTED_STATUS_TYPES:
+        status_type = data.get("status_type")
+        
+        if status_type == "Accounting-On":
+            self._handle_accounting_on(data)
```
nemesifier, line 509: I think the logic could be moved to a model class method.

openwisp_radius/api/freeradius_views.py
```diff
@@ -525,6 +533,49 @@ def send_radius_accounting_signal(self, accounting_data):
...
+        closed_sessions_count = 0
+        for session in open_sessions:
+            now_time = timezone.now()
+            session.stop_time = now_time
+            session.update_time = now_time
+            if session.start_time:
+                session.session_time = int((now_time - session.start_time).total_seconds())
+            session.terminate_cause = "NAS-Reboot"
+            session.save()
+            closed_sessions_count += 1
```
nemesifier: This could be slow if there's many sessions.

We could do it in an update query but we would not send the `post_save` signal, which some implementations may need to get notified when sessions are closed.

Therefore the safest approach is to excute this logic from a background celery task (the celery task calls the model class method).

openwisp_radius/api/freeradius_views.py
```diff
@@ -525,6 +533,49 @@ def send_radius_accounting_signal(self, accounting_data):
...
+            logger.warning("Accounting-On packet received without nas_ip_address")
+            return
+            
+        organization_id = self.request.auth
+        if not organization_id:
+            logger.warning("Accounting-On packet received without valid organization")
+            return
+            
+        from .. import settings as app_settings
+        timeout_hours = app_settings.ACCOUNTING_ON_STALE_SESSION_TIMEOUT
```
nemesifier: Where is this setting coming from? I don't think it makes sense.
If the NAS was rebooted, sessions will be rebooted too so it doesn't make sense to wait.

openwisp_radius/api/freeradius_views.py
Eeshu-Yadav: I've moved the session closing logic to a model class method `AbstractRadiusAccounting.close_nas_sessions()`

openwisp_radius/api/freeradius_views.py
Eeshu-Yadav: implemented this as told

openwisp_radius/api/freeradius_views.py
Eeshu-Yadav: removed the `ACCOUNTING_ON_STALE_SESSION_TIMEOUT` setting completely

openwisp_radius/api/freeradius_views.py
```diff
@@ -525,6 +533,39 @@ def send_radius_accounting_signal(self, accounting_data):
...
+        # Import here to avoid circular imports
+        from ..tasks import handle_accounting_on
+
+        # Execute the task in the background to avoid blocking the API response
+        # and to prevent performance issues when there are many sessions to close
+        handle_accounting_on.delay(
+            organization_id=organization_id, nas_ip_address=nas_ip_address
+        )
+
+        logger.info(
```
nemesifier: Let's use `logger.debug`, this would be the normal behavior, we shouldn't emit output for standard behavior.

openwisp_radius/api/freeradius_views.py
```diff
@@ -525,6 +533,39 @@ def send_radius_accounting_signal(self, accounting_data):
...
+        All active sessions on that NAS should be considered stale and closed.
+        """
+        nas_ip_address = data.get("nas_ip_address")
+        if not nas_ip_address:
+            logger.warning("Accounting-On packet received without nas_ip_address")
+            return
+
+        organization_id = self.request.auth
+        if not organization_id:
+            logger.warning("Accounting-On packet received without valid organization")
```
nemesifier: Well, I am not sure about this. I believe in this case we can just ignore the organization: if a NAS has been cold rebooted, we can just flag any accounting session related to it to terminated, regardless of the org, therefore we could avoid checking and requiring the org. I don't see downsides in this. The pro is that it simplifies the implementation.

openwisp_radius/api/freeradius_views.py
```diff
@@ -525,6 +533,39 @@ def send_radius_accounting_signal(self, accounting_data):
...
+        Handle Accounting-On packets by triggering a background task
+        to close stale sessions for the NAS.
+
+        When a NAS (Network Access Server) sends an Accounting-On packet,
+        it indicates that the NAS has restarted or experienced a power failure.
+        All active sessions on that NAS should be considered stale and closed.
+        """
+        nas_ip_address = data.get("nas_ip_address")
+        if not nas_ip_address:
+            logger.warning("Accounting-On packet received without nas_ip_address")
```
nemesifier: why do we care of the NAS ip address? We care about its mac address more.

openwisp_radius/base/models.py
```diff
@@ -557,6 +557,40 @@ def close_stale_sessions(cls, days=None, hours=None):
             session.terminate_cause = "Session Timeout"
             session.save()
 
+    @classmethod
+    def close_nas_sessions(cls, organization_id, nas_ip_address):
```
nemesifier: `def close_nas_sessions(cls, mac_address):`

openwisp_radius/tasks.py
```diff
@@ -231,3 +231,32 @@ def get_radius_attributes(user):
                 f' RadiusAccounting object of "{user}" user'
             )
     RadiusAccounting.objects.bulk_update(updated_sessions, fields=["groupname"])
+
+
+@shared_task(base=OpenwispCeleryTask)
+def handle_accounting_on(organization_id, nas_ip_address):
```
nemesifier: Same here, I believe we should only pass the mac address.

openwisp_radius/tasks.py
```diff
@@ -231,3 +231,32 @@ def get_radius_attributes(user):
...
+    Returns:
+        Number of sessions closed
+    """
+    RadiusAccounting = load_model("RadiusAccounting")
+
+    closed_sessions_count = RadiusAccounting.close_nas_sessions(
+        organization_id=organization_id, nas_ip_address=nas_ip_address
+    )
+
+    logger.info(
```
nemesifier: Move this log line to the class method please.

openwisp_radius/tasks.py
```diff
@@ -231,3 +231,32 @@ def get_radius_attributes(user):
...
+    This task is executed asynchronously to avoid blocking the API response
+    and to prevent performance issues when there are many sessions to close.
+
+    Args:
+        organization_id: UUID of the organization
+        nas_ip_address: IP address of the NAS
+
+    Returns:
+        Number of sessions closed
+    """
```
nemesifier: This task is pretty small, so this verbose docstring is definitely overkill. I'd move any useful info to the docstring of the model class method and remove the docstring entirely here.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1132,14 +1132,65 @@ def test_accounting_filter_is_open(self):
 
     @freeze_time(START_DATE)
     def test_coova_accounting_on_200(self):
-        self.assertEqual(RadiusAccounting.objects.count(), 0)
+        nas_ip = "192.168.182.1"
```
nemesifier: I recommend avoiding to change existing tests. This test verifies that a specific case works in a certain way, I'd leave it alone and add a new one if needed.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1195,14 +1268,137 @@ def test_coova_accounting_off_200(self):
         self.assertIsNone(response.data)
         self.assertEqual(RadiusAccounting.objects.count(), 0)
 
+    @freeze_time(START_DATE)
+    def test_accounting_on_no_nas_ip(self):
```
nemesifier: `test_accounting_on_no_nas_mac_address`

nemesifier (changes requested):
Great progress @Eeshu-Yadav! :+1: 

Let's use tha nas mac address, also known as `called_station_id`, instead of the NAS IP address, which in some deployments can be the same for all network access servers because they can each operate on a separate private IPv4 subnet and have an IP like `192.168.10.1`.

In order to deal with the Called Station ID effectively, **please make sure to read** #624, it's useful to know those nuances to implement this.

Eeshu-Yadav:
@nemesifier working on that , will raise pr soon

nemesifier:
> @nemesifier working on that , will raise pr soon

@Eeshu-Yadav update this PR by updating this branch, please do not create a new PR.

openwisp_radius/base/models.py
```diff
@@ -557,6 +557,20 @@ def close_stale_sessions(cls, days=None, hours=None):
             session.terminate_cause = "Session Timeout"
             session.save()
 
+    @classmethod
+    def close_stale_sessions_for_nas(cls, nas_ip, organization_id):
+        if not nas_ip or not organization_id:
+            return 0
+        stale_sessions = cls.objects.filter(
+            nas_ip_address=nas_ip,
```
nemesifier: We can't use the IP for this, as multiple NAS deployed on different geografic may use the same private IP (eg: 192.168.0.1) and this code would close the wrong sessions, causing issues.

We need to use the `called_station_id` field, which contains the mac address of the NAS, which by definition is always unique.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1278,6 +1279,82 @@ def test_user_accounting_list_empty_diff_organization(self):
...
+
+        stale_session.refresh_from_db()
+        self.assertIsNotNone(stale_session.stop_time)
+        self.assertEqual(stale_session.terminate_cause, "Stale-Session")
+
+        other_session.refresh_from_db()
+        self.assertIsNone(other_session.stop_time)
+
+    @freeze_time(START_DATE)
+    def test_accounting_on_no_nas_ip(self):
```
nemesifier: This test is not useful.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1221,6 +1221,7 @@ def test_accounting_when_nas_using_pfsense_started(self):
         response = self.client.post(
             self._acct_url,
             data=json.dumps(data),
+            HTTP_AUTHORIZATION=self.auth_header,
```
nemesifier, line 1224: Why this change here? Seems out of place. If the test passes without this, please remove it.

nemesifier (changes requested):

openwisp_radius/tests/test_api/test_freeradius_api.py
Dhanus3133: No, it didn't pass. I got this without adding the change.

```
AssertionError: 403 != 200
```

openwisp_radius/base/models.py
```diff
@@ -557,6 +557,20 @@ def close_stale_sessions(cls, days=None, hours=None):
...
+            organization_id=organization_id,
+            stop_time__isnull=True,
+        )
+        closed_count = stale_sessions.update(
+            stop_time=now(), terminate_cause="Stale-Session"
```
nemesifier: I think we can use:

```suggestion
            stop_time=now(), terminate_cause="NAS Reboot"
```

For more info refer to the definition of `Acct-Terminate-Cause` in the RADIUS accounting RFC.

nemesifier (changes requested):

openwisp_radius/base/models.py
Dhanus3133: Thanks for pointing that out! I missed your comment in the initial review.

nemesifier (approved):

Eeshu-Yadav:
@nemesifier extremely sorry for not updating from a while , i was quite busy in the past two weeks , for clearing the selection criteritia in the lfx

PR #627, opened 2025-08-13
[feature] Added support for simultaneous-use #615

### LINKED ISSUE #615: [feature] Add support for simultaneous-use
Right now it's possible to use the freeradius sql module to support the Simultaneous-Use check, however, this module doesn't know about multi-tenancy, so if a user is member of multiple organizations which have different simultaneous use policies, the freeradius sql module may pick the wrong settings.

To fix this, we need to implement the simultaneous-use feature in the freeradius API, in the authorize view.

We could run this after the counters logic.

I think it's conservative to allow disabling this feature with a global setting, in case users may prefer using the freeradius sql module or any other freeradius module, eg: `OPENWISP_RADIUS_SIMULTANEOUS_USE_ENABLED`.

pandafy:
## Checklist

Closes  #615


openwisp_radius/settings.py
```diff
@@ -113,6 +113,12 @@ def get_default_password_reset_url(urls):
         "location": "disabled",
     },
 )
+ADDITIONAL_RADIUS_CHECKS = []
```
nemesifier: Did you mean to allow overriding this? It looks unoverridable, I think we can remove this and anything related to it to reduce the new lines of code added.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1437,6 +1439,185 @@ def test_authorize_counters_exception_handling(self):
...
+        )
+        self._create_radius_accounting(**acct_data)
+
+        # Authorization should succeed
+        response = self._authorize_user(auth_header=self.auth_header)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data["control:Auth-Type"], "Accept")
+
+    @mock.patch.object(app_settings, "ADDITIONAL_RADIUS_CHECKS", ["Simultaneous-Use"])
+    def test_authorize_simultaneous_use_exceeds_limit(self):
```
nemesifier: what do you think of squashing this test with the previous test and have the assertions run in a `subTest`? That we we can reduce the amount of lines of code needed for this.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -1437,6 +1439,185 @@ def test_authorize_counters_exception_handling(self):
...
+        # Create multiple open sessions
+        for i in range(3):
+            acct_data["unique_id"] = f"test_session_{i}"
+            self._create_radius_accounting(**acct_data)
+
+        # Authorization should still succeed
+        response = self._authorize_user(auth_header=self.auth_header)
+        self.assertEqual(response.status_code, 200)
+        print(response.data)
+        self.assertEqual(response.data["control:Auth-Type"], "Accept")
```
nemesifier: This probably can also be squashed as you simply need to remove the group check.

openwisp_radius/api/freeradius_views.py
```diff
@@ -300,43 +300,99 @@ def get_replies(self, user, organization_id):
...
-                if reply_name not in data or remaining < self._get_reply_value(
-                    data, counter
-                ):
-                    data[reply_name] = remaining
+            group_checks = get_group_checks(
+                user_group.group, include_additional_checks=True
+            )
+
+            # Validate additional radius checks
+            additional_checks_result = self._validate_additional_radius_checks(
```
nemesifier: Why don't we just execute the simultaneous check directly here? Without additional logic? I still don't see the benefits of this "additional radius checks" thing. It reminded me of my early errors in the infant openwisp2 of trying to do too many things too early :joy: :stuck_out_tongue:, I am still traumatized.

When in the future we'll need to support more checks, we can just update the code.

openwisp_radius/api/freeradius_views.py
```diff
@@ -300,43 +300,99 @@ def get_replies(self, user, organization_id):
...
+        """
+        Check if user has exceeded simultaneous use limit
+        Returns True if limit is exceeded, False otherwise
+        """
+        max_simultaneous = int(group_check.value)
+        if max_simultaneous <= 0:
+            return False
+        # Count open sessions for this user
+        open_sessions = RadiusAccounting.objects.filter(
+            username=user.username, stop_time__isnull=True
```
nemesifier: What if the same user has 2 radius accountings in two different organizations with 2 different devices?

Quite unlikely, but the goal is to support multitenancy, right? So I believe we should also add the org here.

docs/user/simultaneous_use.rst
```diff
@@ -0,0 +1,65 @@
+Limiting concurrent sessions (Simultaneous-Use)
```
nemesifier: The docs look a lot better, thanks, I'll push a few minor adjustments.

openwisp_radius/api/freeradius_views.py
```diff
@@ -300,43 +300,99 @@ def get_replies(self, user, organization_id):
...
+                if self._check_simultaneous_use(user, group_check):
+                    data = self.reject_attributes.copy()
+                    if "Reply-Message" not in data:
+                        data["Reply-Message"] = (
+                            "You are already logged in - access denied"
+                        )
+                    return data, self.reject_status
+        return None
+
+    def _check_simultaneous_use(self, user, group_check):
```
nemesifier, line 344:
```suggestion
    def _check_simultaneous_use(self, user, group_check):
```

openwisp_radius/api/freeradius_views.py
```diff
@@ -300,43 +300,99 @@ def get_replies(self, user, organization_id):
...
+        """
+        Check if user has exceeded simultaneous use limit
```
nemesifier, line 323:
```suggestion
        Check if the user has exceeded the Simultaneous-Use limit
```

openwisp_radius/api/freeradius_views.py
```diff
@@ -300,43 +300,99 @@ def get_replies(self, user, organization_id):
...
+        max_simultaneous = int(group_check.value)
+        if max_simultaneous <= 0:
+            return False
+        # Count open sessions for this user
+        open_sessions = RadiusAccounting.objects.filter(
+            username=user.username, stop_time__isnull=True
+        ).count()
+        return open_sessions >= max_simultaneous
+
+    def _execute_counter_checks(self, data, user, group, group_checks):
```
nemesifier:
```suggestion
    def _check_counters(self, data, user, group, group_checks):
```

openwisp_radius/utils.py
```diff
@@ -250,12 +250,13 @@ def get_user_group(user, organization_id):
...
 
 
-def get_group_checks(group):
+def get_group_checks(group, include_additional_checks=False):
     """
     Retrieves a dictionary of checks for the given group.
 
     Parameters:
         group (Group): The group object for which to retrieve the checks.
+        include_additional_checks (bool): Whether to include additional radius checks.
```
nemesifier: Would it work also without these changes? I'd try to simplify this.

nemesifier (changes requested):
It's looking good and it's working! Thanks @pandafy :pray: 

However, I am still convinced the code and tests can be simplified so we can reduce maintenance overhead and test run time. Please see my comments below.

openwisp_radius/utils.py
pandafy: I think, this is necessary here. 

We don't want to include the Simultaneous-Use check in the RADIUS Usage API response, right? 

The other option would be to make a query in the AuthorizeView to get the `Simultaneous-Use` check object if setting is enabled which has a performance penalty.

nemesifier: It seems this function it's used in two places:

1. freeradius authorize view: we need all the checks
2. UserRadiusUsageSerializer: we need only the counters

Point 1 has priority right?

Therefore I'd do the opposite: add a parameter named `counters_only`, which if `True` only returns the `CHECK_ATTRIBUTE_COUNTERS_MAP`.

openwisp_radius/api/freeradius_views.py
```diff
@@ -302,41 +302,83 @@ def get_replies(self, user, organization_id):
...
+            # in the RadiusGroup or if it permits unlimited concurrent sessions.
+            return None
+        open_sessions = RadiusAccounting.objects.filter(
+            username=user.username,
+            organization_id=organization_id,
+            stop_time__isnull=True,
+        ).count()
+        if open_sessions >= max_simultaneous:
+            data = self.reject_attributes.copy()
+            if "Reply-Message" not in data:
```
nemesifier, line 340: I don't understand this if statement: the code of this method doesn't seem to honor the radius replies that are defined in the radius group, so this if statement doesn't seem to provide any value, does it?

Is there a reason for which you decided to ignore the radius-replies defined in the radius groups? I am referring to the `data` variable which contains a dictionary of the radius replies fetched from the DB.

I am not sure we can foresee all the use cases that can be implemented, so I believe it would be useful to honor the radius replies defined in the DB, that would allow users to send any reply message or reply attribute they need without modifying the source code.

openwisp_radius/api/freeradius_views.py
```diff
@@ -302,41 +302,83 @@ def get_replies(self, user, organization_id):
...
+                data["Reply-Message"] = "You are already logged in - access denied"
+            return data, self.reject_status
+
+    def _check_counters(self, data, user, group, group_checks):
```
nemesifier, line 344: You see here you're passing `data` around right? This is what I am referring to.

nemesifier (commented):
Almost there, I have one last doubt remaining, see my 2 comments below.

PR #632, opened 2025-08-24
[models] Sanitize MAC addresses in called_station_id field

### LINKED ISSUE #624: [bug] Mac address of called-station-id needs to be sanitized
The MAC address in the called-station-id field of radius accounting needs to be sanitized because it can come in different formats, eg:

- 00:1A:2B:3C:4D:5E
- 00-1A-2B-3C-4D-5E
- 001A.2B3C.4D5E

Right now, using different formats for the same mac address displays a different mac address in the radius accounting data and it impacts the statistics generated by the monitoring integration.

We are already using [netaddr.EUI in the codebase](https://github.com/search?q=repo%3Aopenwisp%2Fopenwisp-radius%20EUI&type=code), I believe it would be better if we make a function based on it which can sanitize the mac address and return it in the semicolon format.

We could sanitize it at model level (in `RadiusAccounting`), hopefully that would work for most use cases and would also propagate to the monitoring integration.

When doing sanitization at model level, I think we shouldn't fail if the called-station-id is not recognized as a mac address, because the RADIUS RFC does not define called-station-id solely as a mac address, even though it's the most common use case in OpenWISP.

Eeshu-Yadav:
## Checklist

Closes #624 

## Description of Changes
This change implements MAC address sanitization in the called_station_id
field of RadiusAccounting to ensure consistent formatting across different
input formats. MAC addresses are now automatically converted to lowercase
colon-separated format (aa:bb:cc:dd:ee:ff) while preserving non-MAC values
unchanged for RFC compliance.

- Added sanitize_mac_address function in base/models.py using netaddr.EUI
- Integrated automatic MAC sanitization in AbstractRadiusAccounting.save()
- Updated test imports to use sanitize_mac_address from base.models
- Supports multiple MAC formats: colon, dash, dot, and no separators


Eeshu-Yadav:
@nemesifier kindly review

Eeshu-Yadav:
@nemesifier kindly review this

setup.py
```diff
@@ -8,7 +8,7 @@
 
 if sys.argv[-1] == "publish":
     # delete any *.pyc, *.pyo and __pycache__
-    os.system('find . | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf')
+    os.system(r'find . | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf')
```
nemesifier, line 11: why this change? Seems unrelated AI slop

openwisp_radius/management/commands/base/convert_called_station_id.py
```diff
@@ -128,40 +123,101 @@ def handle(self, *args, **options):
...
-                    logger.warning(
+                    BaseConvertCalledStationIdCommand.logger.warning(
                         f'Failed to find a MAC address in "{common_name}". '
                         f"Skipping {radius_session.session_id}!"
                     )
-                else:
-                    radius_session.save()
+                    continue
+
+                from openwisp_radius.base.models import sanitize_mac_address
```
nemesifier: why the import here?

openwisp_radius/base/models.py
```diff
@@ -178,6 +180,49 @@
...
+    Handles various MAC address formats:
+    - 00:1A:2B:3C:4D:5E -> 00:1a:2b:3c:4d:5e
+    - 00-1A-2B-3C-4D-5E -> 00:1a:2b:3c:4d:5e
+    - 001A.2B3C.4D5E -> 00:1a:2b:3c:4d:5e
+    - 001A2B3C4D5E -> 00:1a:2b:3c:4d:5e
+    """
+    # Return empty string or non-string input as is
+    if not mac or not isinstance(mac, str):
+        return mac
+
```
nemesifier: too many blank lines being added here and in other files, please use blank lines sparingly because they bloat the code unnecessarily.

nemesifier (changes requested):
@Eeshu-Yadav when you do UI changes, please always add a screenshot and/or a short screencast demonstrating the result really fixes the issue.

openwisp_radius/base/models.py
Eeshu-Yadav: will update that and keep in mind for next time

setup.py
Eeshu-Yadav: will remove that

openwisp_radius/management/commands/base/convert_called_station_id.py
Eeshu-Yadav: will make it to the top

setup.py
Eeshu-Yadav: actually during qa check this is showing , not major thing to raise a separate so just updated here , 
Screenshot from 2026-01-30 21-25-06
suggest me that should i keep this or remove

openwisp_radius/base/models.py
```diff
@@ -528,6 +565,8 @@ class AbstractRadiusAccounting(OrgMixin, models.Model):
     )
 
     def save(self, *args, **kwargs):
+        if self.called_station_id:
+            self.called_station_id = sanitize_mac_address(self.called_station_id)
```
nemesifier, line 568: This looks like it should go into field level validation or a validator rule.

nemesifier (changes requested):

openwisp_radius/base/models.py
Eeshu-Yadav: fixed

PR #635, opened 2025-08-26
[fix] Added error handling in task.perform_change_of_authorization

pandafy:
## Checklist

--- 

How a user can encounter this edge case:

1. A user initially belongs to a RADIUS Group that provides unrestricted access (`power-users`).
2. The user consumes more than 1 GB of traffic.
3. The user’s RADIUS group is then downgraded to a RADIUS Group with a daily limit of 100 MB (Max-Daily-Session-Traffic).
4. When executing DailyTrafficCounter.check(),  the counter will raise MaxQuotaReached because the user has consumed more than 100 MB on the same day.

### DIFF
--- a/openwisp_radius/tasks.py
+++ b/openwisp_radius/tasks.py
@@ -153,13 +153,12 @@ def perform_change_of_authorization(user_id, old_group_id, new_group_id):
             counter = Counter(user=user, group=check.group, group_check=check)
             value = counter.check()
             return counter.reply_name, value
+        except MaxQuotaReached:
+            return check.attribute, 0
         except Exception as e:
-            # Ignore MaxQuotaReached and KeyError exceptions:
-            # - MaxQuotaReached: raised by Counter.check() when user has
-            #       reached their quota
-            # - KeyError: raised when the check attribute is not in
-            #       CHECK_ATTRIBUTE_COUNTERS_MAP
-            if not isinstance(e, (MaxQuotaReached, KeyError)):
+            # KeyError is raised when the check attribute is not in
+            # CHECK_ATTRIBUTE_COUNTERS_MAP
+            if not isinstance(e, KeyError):
                 logger.exception(
                     f"Got {e} while executing counter for {check.attribute} check."
                 )
--- a/openwisp_radius/tasks.py
+++ b/openwisp_radius/tasks.py
@@ -16,6 +16,7 @@ from openwisp_utils.admin_theme.email import send_email
 from openwisp_utils.tasks import OpenwispCeleryTask
 
 from . import settings as app_settings
+from .counters.exceptions import MaxQuotaReached
 from .radclient.client import RadClient
 from .utils import get_one_time_login_url, load_model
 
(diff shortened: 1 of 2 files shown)


openwisp_radius/tasks.py
```diff
@@ -147,15 +148,22 @@ def get_radsecret_from_radacct(rad_acct):
...
+            Counter = app_settings.CHECK_ATTRIBUTE_COUNTERS_MAP[check.attribute]
+            counter = Counter(user=user, group=check.group, group_check=check)
             value = counter.check()
             return counter.reply_name, value
-        except KeyError:
-            return check.attribute, check.value
         except Exception as e:
-            logger.exception(f"Got {e} while CoA for counter {Counter}")
+            # Ignore MaxQuotaReached and KeyError exceptions:
+            # - MaxQuotaReached: raised by Counter.check() when user has
```
nemesifier: If `MaxQuotaReached` is raised, wouldn't be more correct to `return  counter.reply_name, 0`?

Or am I missing something?

openwisp_radius/tasks.py
```diff
@@ -147,15 +148,21 @@ def get_radsecret_from_radacct(rad_acct):
...
-        counter = Counter(user=user, group=check.group, group_check=check)
         try:
+            Counter = app_settings.CHECK_ATTRIBUTE_COUNTERS_MAP[check.attribute]
+            counter = Counter(user=user, group=check.group, group_check=check)
             value = counter.check()
             return counter.reply_name, value
-        except KeyError:
-            return check.attribute, check.value
+        except MaxQuotaReached:
+            return check.attribute, 0
```
nemesifier, line 157: I gave it a deeper though and I am not sure this makese sense. What's the use of this?
It looks like we're just trying to handle the error without performing any useful action.

I think we need to discuss this further.

Can you please write in the PR description what's the possibles scenarios in which this happens?

pandafy: I have update d the PR description.

nemesifier: Thanks @pandafy, so if that's the case, we should log out the user, shouldn't we?

pandafy: In my testing, if we send a CoA packet with limits less than the amount user has consumed (e.g. limit is 100 MB and user has consumed 1 GB), the captive portal automatically logs out the user. 

RFC defines a special kind of CoA packet called ["Disconnect Messages"](https://wiki.freeradius.org/protocol/Disconnect-Messages). This will require adding a new method in the RadClient class. 

https://github.com/openwisp/openwisp-radius/blob/3d6264b31b10e3316de529aecb5f6e876e77e9c3/openwisp_radius/radclient/client.py#L29-L73

nemesifier: What if the max quota is due to time usage? Setting Session-Timeout to 0 AFAIK means disabling it completely. 

For that reason I think sending an explicit disconnect may be more reliable and probably more readable.

pandafy:
Superseded by https://github.com/openwisp/openwisp-radius/pull/645

PR #637, opened 2025-08-30
[fix] Error when creating RadiusGroup with inlines (#604)

### LINKED ISSUE #604: [bug] Cannot create RadiusGroup and RadiusGroupCheck in same request
**Describe the bug**
It is currently not possible to create both a RadiusGroup and RadiusGroupCheck in a single request from the Django admin interface. This limitation requires users to first create the RadiusGroup, save it, and then add the checks, resulting in additional steps and a less efficient workflow.

**Steps To Reproduce**
1. Go to http://127.0.0.1:8000/admin/openwisp_radius/radiusgroup/add/.
2. Attempt to create a new RadiusGroup and add RadiusGroupCheck object from the admin inline before saving.
3. Save the form.

**Expected behavior**
It should be possible to create a RadiusGroup and its related RadiusGroupCheck objects in a single save operation from the admin interface.

**Screenshots**

Image

Dhanus3133:
Closes #604 

The crash happened because the `clean()` method tried to validate using the parent group before that group was actually saved, which caused the ValidationError.

Fixed it by:

* Skip `clean()` validation when the parent group hasn’t been saved yet (fixes error when creating a new group).
* Add a `save()` method to ensure `groupname` is populated from the parent’s name before saving (prevents empty strings).

### DIFF
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -219,7 +219,7 @@ class AutoGroupnameMixin(object):
 
     def save(self, *args, **kwargs):
         self._set_groupname()
-        super().save(*args, **kwargs)
+        return super().save(*args, **kwargs)
 
 
 class AttributeValidationMixin(object):
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -200,6 +200,10 @@ class AutoUsernameMixin(object):
 
 
 class AutoGroupnameMixin(object):
+    def _set_groupname(self):
+        if self.group:
+            self.groupname = self.group.name
+
     def clean(self):
         """
         automatically sets groupname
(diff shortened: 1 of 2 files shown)


openwisp_radius/base/models.py
```diff
@@ -212,6 +214,14 @@ def clean(self):
                 {"groupname": _NOT_BLANK_MESSAGE, "group": _NOT_BLANK_MESSAGE}
             )
 
+    def save(self, *args, **kwargs):
+        """
+        Ensure groupname is populated from the group relation before saving.
```
nemesifier: Remove this docstring as it refers to a specific line in this method, while the goal of the docstring is to document an entire method.

openwisp_radius/base/models.py
```diff
@@ -212,6 +214,14 @@ def clean(self):
...
+        """
+        if self.group:
+            self.groupname = self.group.name
```
nemesifier: I am puzzled, this is already performed by the `clean` method, why do we need to do it again?

If we really need to do this, at least we should move the duplicated code to a method and call the method instead of duplicating the code.

openwisp_radius/base/models.py
```diff
@@ -212,6 +214,14 @@ def clean(self):
...
+        super().save(*args, **kwargs)
```
nemesifier: In case we keep this overridden save method, let's return:
```suggestion
        return super().save(*args, **kwargs)
```

nemesifier (changes requested):

openwisp_radius/base/models.py
Dhanus3133: Removed it.

openwisp_radius/base/models.py
Dhanus3133: When we create a `RadiusGroup` and its inlines at the same time, the inline’s `clean()` has to skip validation because the parent group doesn’t exist yet. That also means the `groupname` doesn’t get set. The `save()` method is there to fix that, it makes sure the `groupname` is filled from the parent just before saving, so it's not empty.

I have updated it to be reused.

nemesifier (approved):

PR #645, opened 2025-09-05
[fix] Improved logic of CoA #643

### LINKED ISSUE #643: [feature] Improve logic of CoA operation for RADIUS Replies, counter handling, and code duplication

There are several shortcomings in the current implementation of the Change of Authorization (CoA) operation:

1. The code does not include the RADIUS Replies of the new group in the CoA packet, potentially resulting in incomplete or incorrect authorization updates.
2. When processing counters for the remaining data, there is no error handling if any `Counter.check()` call raises `MaxQuotaReached`. This can cause uncaught exceptions and unpredictable behavior.
3. If any counter raises `MaxQuotaReached`, the correct behavior should be to send a [Disconnect Message](https://wiki.freeradius.org/protocol/Disconnect-Messages) instead of setting RADIUS attributes for the new group.
4. There is duplicate logic between `AuthorizeView._check_counters` and the CoA logic for retrieving RADIUS attributes for the new RADIUS group. This violates DRY principles and makes maintenance harder.

- Include RADIUS Replies of the new group in the CoA packet so that attributes are correctly updated.
- Add error handling in the CoA logic: if any `Counter.check()` raises `MaxQuotaReached`, send a Disconnect Message per FreeRADIUS protocol, rather than proceeding with setting group attributes.
- Refactor the code to remove duplication between `AuthorizeView._check_counters` and the CoA logic for retrieving RADIUS attributes, making the code DRY and easier to maintain.

**Additional context**
- See FreeRADIUS documentation on [Disconnect Messages](https://wiki.freeradius.org/protocol/Disconnect-Messages) for reference.
- https://github.com/openwisp/openwisp-radius/pull/635

pandafy:
## Checklist

Closes #643

## Description of Changes

- Include RADIUS replies of the new group in CoA packet
- Issue Discconect request is check raises MaxQuotaReached
- Use Counter.reply_name instead of Counter attribute in the CoA attributes


openwisp_radius/radclient/client.py
```diff
@@ -25,6 +21,19 @@ def _EncodeKeyValues(self, key, values):
             return (key, values)
         return super()._EncodeKeyValues(key, values)
 
+    def AddAttribute(self, key, value):
+        try:
+            return super().AddAttribute(key, value)
+        except KeyError:
+            # Skip attributes not found in the dictionary
+            logger.warning(f"RADIUS attribute '{key}' not found in dictionary")
```
nemesifier, line 29: Are you sure this shouldn't be logged as error?

Do we need to be notified if this happens?

openwisp_radius/utils.py
```diff
@@ -283,6 +285,104 @@ def get_group_checks(group, counters_only=False):
...
+            remaining = counter.check()
+        except SkipCheck:
+            continue
+        except Exception as e:
+            if not isinstance(e, MaxQuotaReached):
+                logger.exception(f'Got exception "{e}" while executing {counter}')
+                continue
+            if raise_quota_exceeded:
+                raise
+
```
nemesifier: Can we reduce the amount of blank lines in this PR please? Let's keep the code compact please.

nemesifier (commented):
Looks good. I have two observations, see below.

**PS: are you sure we don't need to update anything in the docs?**

openwisp_radius/radclient/client.py
pandafy: This ignores checks like `Simultaneous-Use` which are not supposed to be sent to the NAS.

nemesifier (approved):

PR #646, opened 2025-09-08
[feature] Run batch user creation asynchronously #608

### LINKED ISSUE #608: [feature/change] Allow running batch creation operations asynchronously
When creating many users with the batch user creation feature, the web server may return a time out before the application responds.

To make the solution more robust, I think we should do something along these lines (this is not an exhaustive plan, further research may be needed):

- Move the logic which generates the users to a background celery task.
- Add some kind of status field, this could be a simple boolean, eg: `completed`, defaulting to `False`.
- The UI must be updated to clearly indicate the operation is being processed and not complete yet, we shall hide/disable some aspects if needed to avoid users tampering with the data. We can use a warning header (like when using the django message framework, but [overriding the template instead of relying on the message framework](https://github.com/openwisp/openwisp-controller/blob/3a45fd9ecc9bc3eaf7e5c8be4f82f573829cdf9d/openwisp_controller/config/templates/admin/config/device/change_form.html#L7) which is unpractical here due to the need of passing around the HTTP request object). 
- Once the import/generation operation is completed, we flag the new field as `True`.
- In the change_form template, we add websocket JS logic to reload the page when the operation has completed, however, if we detect that nobody is browsing the page anymore (eg: we can program the JS to send an acknowledgment to the server if needed), we can send a notification to all the org admins and superusers with openwisp-notifications to let them know the operation has completed. This point happens only when running in async mode.
- We must update the REST API to take this flow into account, eg: add the new field, do not allow any delete operation while the operation is being completed.
- REST API: let's make sure the batch object cannot be edited once created (in the same as the admin works)
- Admin: do not allow to delete until the operation is completed
- We must ensure the new flow works well also for small batches that would have worked well with the current synchronous code, maybe we can use the async flow only for batches that are larger than a certain amount  of users, like 15 and make this configurable, eg: `OPENWISP_RADIUS_BATCH_ASYNC_THRESHOLD`.

Dhanus3133:
Closes #608


openwisp_radius/api/views.py
```diff
@@ -108,17 +109,42 @@ def post(self, request, *args, **kwargs):
...
-            valid_data["organization"] = valid_data.pop("organization_slug", None)
-            batch = serializer.create(valid_data)
+            num_of_users = valid_data.get("number_of_users", 0)
             strategy = valid_data.get("strategy")
-            if strategy == "csv":
-                batch.csvfile_upload()
-                response = RadiusBatchSerializer(batch, context={"request": request})
+            organization = valid_data.pop("organization_slug", None)
+
+            valid_data.pop("number_of_users", None)
```
nemesifier, line 113: All of this can be done in the model and we avoid duplicating the logic in admin and REST API.

openwisp_radius/api/views.py
```diff
@@ -108,17 +109,42 @@ def post(self, request, *args, **kwargs):
...
+                and num_of_users >= app_settings.BATCH_ASYNC_THRESHOLD
+            ) or (strategy == "csv")
+
+            if is_async:
+                process_radius_batch.delay(batch.pk, number_of_users=num_of_users)
+                response_serializer = RadiusBatchSerializer(
+                    batch, context={"request": request}
+                )
+                return Response(
+                    response_serializer.data, status=status.HTTP_202_ACCEPTED
```
nemesifier: The different status code depending on `app_settings.BATCH_ASYNC_THRESHOLD` is probably the only thing that is worth keeping here.

openwisp_radius/tasks.py
```diff
@@ -231,3 +233,37 @@ def get_radius_attributes(user):
...
+
+
+@shared_task
+def process_radius_batch(batch_id, number_of_users=None):
+    RadiusBatch = load_model("RadiusBatch")
+    channel_layer = get_channel_layer()
+    group_name = f"radius_batch_{batch_id}"
+    batch = None
+
+    try:
```
nemesifier, line 134: This try block looks too big, you should only wrap lines which can fail.

openwisp_radius/tasks.py
```diff
@@ -231,3 +233,37 @@ def get_radius_attributes(user):
...
+        batch = RadiusBatch.objects.get(pk=batch_id)
+        batch.status = "processing"
+        batch.save(update_fields=["status"])
```
nemesifier: Let's move this logic as a class method and keep this file lean.

openwisp_radius/tasks.py
```diff
@@ -231,3 +233,37 @@ def get_radius_attributes(user):
...
+
+@shared_task
+def process_radius_batch(batch_id, number_of_users=None):
+    RadiusBatch = load_model("RadiusBatch")
+    channel_layer = get_channel_layer()
+    group_name = f"radius_batch_{batch_id}"
+    batch = None
+
+    try:
+        batch = RadiusBatch.objects.get(pk=batch_id)
```
nemesifier, line 135: Please Use the same pattern we use in other modules, eg:

https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/tasks.py#L23-L29
https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/tasks.py#L43-L49

openwisp_radius/tasks.py
```diff
@@ -231,3 +233,37 @@ def get_radius_attributes(user):
                 f' RadiusAccounting object of "{user}" user'
             )
     RadiusAccounting.objects.bulk_update(updated_sessions, fields=["groupname"])
+
+
+@shared_task
```
nemesifier: Let's set a soft limit if possible:
https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/tasks.py#L79
https://github.com/openwisp/openwisp-controller/blob/master/openwisp_controller/config/tasks.py#L69

tests/openwisp2/settings.py
```diff
@@ -311,6 +316,27 @@
 
 OPENWISP_USERS_AUTH_API = True
 
+ASGI_APPLICATION = "openwisp2.routing.application"
+
+if TESTING:
+    CHANNEL_LAYERS = {"default": {"BACKEND": "channels.layers.InMemoryChannelLayer"}}
```
nemesifier: Are we sure this wouldn't cause issues with selenium tests?

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_change_form.html
```diff
@@ -1,10 +1,62 @@
...
+    <script>
+        (function($) {
+            const batchStatus = '{{ original.status }}';
+            if (batchStatus === 'processing') {
+                const batchId = '{{ original.pk }}';
+                const protocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
+                const wsPath = protocol + '//' + window.location.host + '/ws/radius/batch/' + batchId + '/';
+                const socket = new WebSocket(wsPath);
+
+                socket.onmessage = function(e) {
```
nemesifier, line 54: Let's make sure we have at least 1 browser test for this.

nemesifier (commented):
We strive to follow the pattern of fat models, lean views.
Admin, REST API, Celery tasks are views of the business logic.
We prefer to group key logic as this in either models or separate classes.

openwisp_radius/base/models.py
```diff
@@ -1064,6 +1086,55 @@ def _remove_files(self):
         if self.csvfile:
             self.csvfile.storage.delete(self.csvfile.name)
 
+    def schedule_processing(self, number_of_users=0):
+        is_async = (
+            self.strategy == "prefix"
+            and number_of_users >= app_settings.BATCH_ASYNC_THRESHOLD
+        ) or (self.strategy == "csv" and self.csvfile)
```
nemesifier: I think for consistency we should trigger async for CSV only if the amount of rows is higher or equal than `app_settings.BATCH_ASYNC_THRESHOLD`.

openwisp_radius/api/serializers.py
```diff
@@ -435,7 +441,7 @@ def validate(self, data):
     class Meta:
         model = RadiusBatch
         fields = "__all__"
-        read_only_fields = ("created", "modified", "user_credentials")
+        read_only_fields = ("created", "modified", "user_credentials", "status")
```
nemesifier: Can we reorder thes fields?

```suggestion
        read_only_fields = ("status", "user_credentials", "created", "modified")
```

openwisp_radius/base/models.py
```diff
@@ -913,6 +929,12 @@ class AbstractRadiusBatch(OrgMixin, TimeStampedEditableModel):
         blank=True,
         help_text=_("If left blank users will never expire"),
     )
+    status = models.CharField(
```
nemesifier: Please move this field so it's placed before strategy.

openwisp_radius/base/models.py
```diff
@@ -1064,6 +1086,55 @@ def _remove_files(self):
...
+            self.status = self.COMPLETED
+            self.save(update_fields=["status"])
+
+            if is_async:
+                notify.send(
+                    sender=self.organization,
+                    type="batch_processing_completed",
+                    target=self.organization,
+                )
+
```
nemesifier: Let's not add too many blank lines please.
```suggestion
```

A few blank lines here and there can help readability but many blank lines just make the the code less compact which IMHO is not great for readability.

openwisp_radius/base/models.py
```diff
@@ -1064,6 +1086,55 @@ def _remove_files(self):
...
+            )
+            self.status = self.FAILED
+            self.save(update_fields=["status"])
+        finally:
+            if channel_layer:
+                async_to_sync(channel_layer.group_send)(
+                    group_name, {"type": "batch_status_update", "status": self.status}
+                )
+            else:
+                logger.warning("Skipping WebSocket notification.")
```
nemesifier: I don't understand what this log line is warning about? Can we add a bit more context for the admins who'll debug this?

openwisp_radius/migrations/0041_radiusbatch_status.py
```diff
@@ -0,0 +1,34 @@
...
+                    ("processing", "Processing"),
+                    ("completed", "Completed"),
+                    ("failed", "Failed"),
+                ],
+                db_index=True,
+                default="pending",
+                max_length=16,
+            ),
+        ),
+        migrations.RunPython(set_existing_batches_completed, migrations.RunPython.noop),
```
nemesifier: This will fail in prod on postgres. Please move it to a separata data migration.

See https://docs.djangoproject.com/en/5.2/ref/migration-operations/#runpython for more info.

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_change_form.html
```diff
@@ -1,10 +1,62 @@
...
+            <a href="{{ download_rad_batch_pdf_url }}" id="downloadpdflink">{% trans "Download User Credentials" %}</a>
+        </li>
+    {% endif %}
+    {{ block.super }}
+{% endblock %}
+
+{% block content %}
+    {% if original.status == 'processing' %}
+        <div class="ow-alert-info">
+            <p><b>{% trans 'Processing:' %}</b> {% trans 'This batch is currently being processed. This page will update automatically when the operation is complete.' %}</p>
```
nemesifier: let's break long lines please

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_change_form.html
```diff
@@ -1,10 +1,62 @@
...
+        </div>
+    {% elif original.status == 'failed' %}
+        <div class="ow-alert-error">
+            <p><b>{% trans 'Failed:' %}</b> {% trans 'This batch operation failed. Please check the system logs for details or use the "Re-process" action.' %}</p>
+        </div>
+    {% elif original.status == 'completed' %}
+        <div class="ow-alert-success">
```
nemesifier: We assume this will be the standard case so I don't think showing the message is needed at all.

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_change_form.html
```diff
@@ -1,10 +1,62 @@
...
+        <li>
+            <a href="{{ download_rad_batch_pdf_url }}" id="downloadpdflink">{% trans "Download User Credentials" %}</a>
+        </li>
+    {% endif %}
+    {{ block.super }}
+{% endblock %}
+
+{% block content %}
+    {% if original.status == 'processing' %}
+        <div class="ow-alert-info">
```
nemesifier: Django already has styles for warning and error messages, we are leveraging this in several places, see:
https://github.com/search?q=org%3Aopenwisp+messagelist&type=code

Overriding the `messages` template block is a robust solution, eg:

```django
{% block messages %}
{{ block.super }}
  <ul class="messagelist">
    <li class="[warning|error]">
      ...
    </li>
  </ul>
{% endblock %}
```

openwisp_radius/templates/openwisp-radius/admin/rad_batch_users_change_form.html
```diff
@@ -1,10 +1,62 @@
...
+    {% if download_rad_batch_pdf_url %}
+        <li>
+            <a href="{{ download_rad_batch_pdf_url }}" id="downloadpdflink">{% trans "Download User Credentials" %}</a>
+        </li>
+    {% endif %}
+    {{ block.super }}
+{% endblock %}
+
+{% block content %}
+    {% if original.status == 'processing' %}
```
nemesifier: Let's use double quotes for consistency please. It seems like black is not reformatting this file automatically. Applies to all the other occurrences.

openwisp_radius/tests/test_selenium.py
```diff
@@ -199,3 +204,50 @@ def test_csv_user_generation(self):
...
+            By.CSS_SELECTOR, ".ow-alert-info b", 10
+        )
+        self.assertIn("Processing:", processing_message_element.text)
+
+        tasks.process_radius_batch(batch.pk, number_of_users=0)
+
+        WebDriverWait(self.web_driver, 10).until(
+            expected_conditions.staleness_of(processing_message_element)
+        )
+
```
nemesifier: can we get rid of all the blank lines please?

openwisp_radius/tests/test_selenium.py
```diff
@@ -199,3 +204,50 @@ def test_csv_user_generation(self):
...
+        completed_message_element = self.wait_for_visibility(
+            By.CSS_SELECTOR, ".ow-alert-success b", 10
+        )
+        self.assertIn("Completed:", completed_message_element.text)
```
nemesifier: As we're changing this, you can look for the status line.

openwisp_radius/admin.py
```diff
@@ -399,19 +402,15 @@ def get_fields(self, request, obj=None):
...
-            elif strategy == "prefix":
-                prefix = data.get("prefix")
-                n = data.get("number_of_users")
-                obj.prefix_add(prefix, n)
-        else:
-            obj.save()
+        if change:
+            super().save_model(request, obj, form, change)
+            return
+
```
nemesifier: Too many blank lines.

openwisp_radius/admin.py
```diff
@@ -466,16 +465,42 @@ def get_readonly_fields(self, request, obj=None):
...
+                "and is now being processed in the background."
+            ) % {"name": str(obj)}
+            self.message_user(request, msg, messages.SUCCESS)
+            if post_url_continue is None:
+                post_url_continue = reverse(
+                    f"admin:{self.opts.app_label}_{self.opts.model_name}_change",
+                    args=(obj.pk,),
+                    current_app=self.admin_site.name,
+                )
+            return redirect(post_url_continue)
```
nemesifier: This redirect must be the reason for which if you hit "Save" instead of "Save and continue" you're redirected to the batch details instead of the batch list page. I think this changes the behavior of the admin in a way that is not consistent with the rest of the application so it's not great, what's the reason for this redirect? It is really needed? If not, we could remove it.

openwisp_radius/apps.py
```diff
@@ -66,6 +68,28 @@ def ready(self, *args, **kwargs):
                 strong_identity=True,
             )
 
+    def register_notification_types(self):
+        RadiusBatch = load_model("RadiusBatch")
+        register_notification_type(
```
nemesifier: Registering a new type creates a new notification setting, which brings a few disadvantages:

1. More DB stuff, more load, more complexity, more code
2. It can be disabled, so if users decide they don't want to receive any notifications they won't receive this

We recently introduced a new notification called ["generic_message"](https://openwisp.io/docs/dev/notifications/developer/notification-types.html#generic-message), eg:

Image

This core notification type is suited for cases like this one and has the following advantages:

1. It doesn't require registering, as it's registered by default in the notifications module
2. it can be used anywhere in openwisp without creating more settings DB records
3. generic messages do not send emails and cannot be turned off
4. by design the aim of this notification type is to notify users about stuff that happened in the background without having to look at application logs, making debugging easier
5. it allows adding long text which is shown in an overlay. So if there's any error that was logged somewhere, we can also add it in the description of the notification and it will be shown after clicking on it

Please [refer to the "generic_message" docs](https://openwisp.io/docs/dev/notifications/developer/notification-types.html#generic-message), it should be well explained (if you think something is not clear please let us know).

openwisp_radius/consumers.py
```diff
@@ -0,0 +1,47 @@
...
+    def _user_can_access_batch(self, user, batch_id):
+        RadiusBatch = load_model("RadiusBatch")
+        # Superusers have access to everything,
+        if user.is_superuser:
+            return RadiusBatch.objects.filter(pk=batch_id).exists()
+        # For non-superusers, check their managed organizations
+        try:
+            RadiusBatch.objects.get(
+                pk=batch_id, organization__in=user.organizations_managed
+            )
```
nemesifier: Why also not use exists here?

Eg:

```python
filter_kwargs = {"pk": "batch_id"}
if not user.is_superuser:
    filter_kwargs["organization__in"] = user.organizations_managed
return RadiusBatch.objects.filter(**filter_kwargs).exists()
```

openwisp_radius/tasks.py
```diff
@@ -231,3 +232,22 @@ def get_radius_attributes(user):
...
+
+@shared_task(soft_time_limit=7200)
+def process_radius_batch(batch_id, number_of_users=None):
+    RadiusBatch = load_model("RadiusBatch")
+    try:
+        batch = RadiusBatch.objects.get(pk=batch_id)
+    except ObjectDoesNotExist as e:
+        logger.warning(f'process_radius_batch("{batch_id}") failed: {e}')
+        return
+
```
nemesifier:
```suggestion
```

openwisp_radius/urls.py
```diff
@@ -9,7 +9,7 @@
 def get_urls(api_views=None, social_views=None, saml_views=None):
     """
     Returns a list of urlpatterns
-    Arguements:
+    Arguments:
```
nemesifier, line 12: Thanks for fixing this.

tests/openwisp2/settings.py
```diff
@@ -311,6 +316,37 @@
 
 OPENWISP_USERS_AUTH_API = True
 
+ASGI_APPLICATION = "openwisp2.routing.application"
+
+if TESTING:
+    DATABASES["default"]["TEST"] = {
+        "NAME": os.path.join(BASE_DIR, "test_openwisp_radius.db"),
+    }
```
nemesifier: Can we adopt the same strategy as this (for consistency)?
https://github.com/openwisp/openwisp-network-topology/commit/f8bb68c60c16dab1ba0144e8a5d170a7f555b9e1#diff-26cea5c755fb048b49985aa06de44b90bfdfe6e9165f73f08bac2197df7e61e8R19-R22

runtests
```diff
@@ -1,8 +1,14 @@
...
 set -e
 
-coverage run runtests.py --parallel > /dev/null 2>&1 || ./runtests.py
-SAMPLE_APP=1 coverage run ./runtests.py --parallel > /dev/null 2>&1 || SAMPLE_APP=1 ./runtests.py
-MONITORING_INTEGRATION=1 coverage run runtests.py
+coverage run runtests.py --parallel \
+  \ --exclude-tag=no_parallel >/dev/null 2>&1 || ./runtests.py --exclude-tag=no_parallel
+SAMPLE_APP=1 coverage run ./runtests.py --parallel \
+  \ --exclude-tag=no_parallel --exclude-tag=selenium_tests >/dev/null 2>&1 || SAMPLE_APP=1 ./runtests.py \
+  \ --exclude-tag=no_parallel --exclude-tag=selenium_tests
```
nemesifier: isn't the initial slash here and above redundant?

nemesifier (changes requested):
So far it seems to almost meet all requirements, I found one blunder regarding the notification not targeting the right object and a few details to improve. See below.

## Notification points to org instead of batch object

I created a batch and called it "AsyncTest".

The notification looks like:

```
The batch user creation "[testing](http://localhost:8000/admin/openwisp_users/organization/51689c38-7a02-4609-9326-a79e1e7f8b1d/change/)" was successfully processed.
```

And points to the organization.

It should instead point to the Batch User Creation object.

## New batch: hide status

The status line doesn't provide any value here and it would be better to hide it either with python code or CSS.

Image

## Existing batch

Current batch details page:

Image

I think we should show the status field somewhere on top.

I am also not sure about the line "Completed: This batch operation has completed successfully." provides any real value.

I will leave more comments below about both.

nemesifier (commented):
There's a conflict with `master`.

openwisp_radius/base/models.py
Dhanus3133: Removed it

openwisp_radius/apps.py
Dhanus3133: Thanks, Updated it.

openwisp_radius/consumers.py
Dhanus3133: Nice catch, I missed it earlier. Fixed it now.

openwisp_radius/api/views.py
```diff
@@ -108,17 +108,26 @@ def post(self, request, *args, **kwargs):
...
+            organization = valid_data.pop("organization_slug", None)
+
+            valid_data.pop("number_of_users", None)
+            batch = serializer.save(organization=organization)
+
+            is_async = batch.schedule_processing(number_of_users=num_of_users)
+
+            batch.refresh_from_db()
+            response_serializer = self.get_serializer(batch)
+
```
nemesifier: Please remove these blank lines as they don't provide any value.

openwisp_radius/api/views.py
```diff
@@ -108,17 +108,26 @@ def post(self, request, *args, **kwargs):
...
+                return Response(
+                    response_serializer.data, status=status.HTTP_202_ACCEPTED
+                )
             else:
-                batch.prefix_add(valid_data.get("prefix"), num_of_users)
-                response = RadiusBatchSerializer(batch, context={"request": request})
-            return Response(response.data, status=status.HTTP_201_CREATED)
+                return Response(
+                    response_serializer.data, status=status.HTTP_201_CREATED
+                )
```
nemesifier: Let's reduce repetition:

```python
status_code = status.HTTP_202_ACCEPTED if is_async else status.HTTP_201_CREATED
return Response(
    response_serializer.data, status=status_code
)
```

openwisp_radius/base/models.py
```diff
@@ -1064,6 +1086,73 @@ def _remove_files(self):
...
+                csv_data = self.csvfile.read()
+                decoded_data = decode_byte_data(csv_data)
+                items_to_process = sum(1 for row in csv.reader(StringIO(decoded_data)))
+                self.csvfile.seek(0)
+            except Exception as e:
+                logger.error(f"Could not count rows in CSV for batch {self.pk}: {e}")
+                items_to_process = app_settings.BATCH_ASYNC_THRESHOLD
+
+        is_async = items_to_process >= app_settings.BATCH_ASYNC_THRESHOLD
+
```
nemesifier: Too many blank lines.
```suggestion
```

openwisp_radius/base/models.py
```diff
@@ -1064,6 +1086,73 @@ def _remove_files(self):
...
+        except Exception as e:
+            logger.error(
+                "RadiusBatch %s failed during processing:", self.pk, exc_info=True
+            )
+            self.status = self.FAILED
+            self.save(update_fields=["status"])
+            notify.send(
+                type="generic_message",
+                level="error",
+                message=_(f"The batch creation operation for '{self.name}' failed."),
```
nemesifier:
```suggestion
                message=_(f"The batch creation operation for \"{self.name}\" failed."),
```

openwisp_radius/base/models.py
```diff
@@ -1064,6 +1086,73 @@ def _remove_files(self):
...
+            elif self.strategy == "csv":
+                self.csvfile_upload()
+            self.status = self.COMPLETED
+            self.save(update_fields=["status"])
+            if is_async:
+                notify.send(
+                    type="generic_message",
+                    level="success",
+                    message=_(
+                        f"The batch creation operation for '{self.name}' "
```
nemesifier:
```suggestion
                        f"The batch creation operation for \"{self.name}\" "
```

nemesifier (changes requested):
The status field in the admin change page was still being shown as last, so I fixed this in 6f5a12a (and pushed some other minor improvements).

During testing I found another minor issue I didn't see before, but it should be quick to fix:
when using the API to create a batch operation, the `pdf_link` in the 202 response refers to an empty PDF (since the batch hasn't been processed yet).

We need to change the `get_pdf_link` method to return the link only when the status is completed and add a test for this, I think we can add a `subTest` in the existing `test_api_batch_pdf_link`.

See my suggestions below to improve the readability and maintenability of the code.

openwisp_radius/settings.py
```diff
@@ -244,3 +244,4 @@ def get_default_password_reset_url(urls):
     ow_users_settings.EXPORT_USERS_COMMAND_CONFIG["select_related"].extend(
         ["registered_user"]
     )
+BATCH_ASYNC_THRESHOLD = get_settings_value("BATCH_ASYNC_THRESHOLD", 15)
```
nemesifier, line 247: Please add this setting to the settings.rst docs page, I hadn't noticed this in my previous reviews as I focused mostly on testing the functionality and reviewing the code, now I was about to approve but I didn't see any documentation update.

We can add this above `OPENWISP_RADIUS_BATCH_DEFAULT_PASSWORD_LENGTH`.

openwisp_radius/settings.py
```diff
@@ -244,4 +244,6 @@ def get_default_password_reset_url(urls):
     ow_users_settings.EXPORT_USERS_COMMAND_CONFIG["select_related"].extend(
         ["registered_user"]
     )
-BATCH_ASYNC_THRESHOLD = get_settings_value("BATCH_ASYNC_THRESHOLD", 15)
+OPENWISP_RADIUS_BATCH_ASYNC_THRESHOLD = get_settings_value(
```
nemesifier: Why did you change this? It was fine before: it was consistent with what we already have.

Dhanus3133: Oops, reverted back.

docs/user/settings.rst
```diff
@@ -111,8 +111,8 @@ The default encryption format for storing radius check values.
 A list of disabled encryption formats, by default all formats are enabled
 in order to keep backward compatibility with legacy systems.
 
-``OPENWISP_RADIUS_BATCH_ASYNC_THRESHOLD``
```
nemesifier: This was right.

PR #647, opened 2025-09-09
[change] Allow counters to return multiple replies #634

### LINKED ISSUE #634: [change] Allow counters to return multiple replies
We need to allow counters to return multiple replies.

In these cases think we can allow `reply_name` to be a list. The `check()` method can return the values for the replies listed in `reply_name` following the same order.

We then need to update the freeradius authorize view to take this case into account to be able to add multiple replies when `reply_name` is a list and the value returned by `check()` is a list.

Maybe to keep the implementation simple we could just make these always a list, even if there's only one value to be returned we can convert these to be a list with a single element, I am not sure about this, whatever ends up simpler to read and maintain is good to me.

Docs may need to be updated.

pandafy:
## Checklist

Closes #634

**Blockers**

- [ ] #645

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,6 @@ on:
   pull_request:
     branches:
       - master
-      # TODO: Remove before merging
-      - issues/643-coa
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches:
       - master
+      # TODO: Remove before merging
+      - issues/643-coa
 
 jobs:
   build:
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -33,7 +33,7 @@ from openwisp_users.backends import UsersAuthenticationBackend
 
 from .. import settings as app_settings
 from ..base.forms import PasswordResetForm
-from ..counters.exceptions import MaxQuotaReached, SkipCheck
+from ..counters.exceptions import SkipCheck
 from ..registration import REGISTRATION_METHOD_CHOICES
 from ..utils import (
     get_group_checks,
@@ -304,9 +304,10 @@ class UserGroupCheckSerializer(serializers.ModelSerializer):
                 group=self.context["group"],
                 group_check=obj,
             )
-            return counter.consumed()
-        except MaxQuotaReached:
-            return int(obj.value)
+            consumed = counter.consumed()
+            if consumed > int(obj.value):
+                consumed = int(obj.value)
+            return consumed
         except (SkipCheck, ValueError, KeyError):
             return None
 
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -304,10 +304,7 @@ class UserGroupCheckSerializer(serializers.ModelSerializer):
                 group=self.context["group"],
                 group_check=obj,
             )
-            # Python can handle 64 bit numbers and
-            # hence we don't need to display Gigawords
-            remaining = counter.check(gigawords=False)
-            return int(obj.value) - remaining
+            return counter.consumed()
         except MaxQuotaReached:
             return int(obj.value)
         except (SkipCheck, ValueError, KeyError):
--- a/openwisp_radius/api/serializers.py
+++ b/openwisp_radius/api/serializers.py
@@ -305,8 +305,9 @@ class UserGroupCheckSerializer(serializers.ModelSerializer):
                 group_check=obj,
             )
             consumed = counter.consumed()
-            if consumed > int(obj.value):
-                consumed = int(obj.value)
+            value = int(obj.value)
+            if consumed > value:
+                consumed = value
             return consumed
         except (SkipCheck, ValueError, KeyError):
             return None
--- a/openwisp_radius/counters/base.py
+++ b/openwisp_radius/counters/base.py
@@ -21,9 +21,17 @@ class BaseCounter(ABC):
         pass
 
     @property
-    @abstractmethod
-    def reply_name(self):  # pragma: no cover
-        pass
+    def reply_names(self):
+        # BACKWARD COMPATIBILITY: In previous versions of openwisp-radius,
+        # the Counter.reply_name was a string instead of a tuple. Thus,
+        # we need to convert it to a tuple if it's a string.
+        if not hasattr(self, "reply_name"):
+            raise NotImplementedError(
+                "Counter classes must define 'reply_names' property."
+            )
+        if isinstance(self.reply_name, str):
+            return (self.reply_name,)
+        return self.reply_name
 
     @property
     @abstractmethod
@@ -43,7 +51,6 @@ class BaseCounter(ABC):
     # sqlcounter module, now we can translate it with gettext
     # or customize it (in new counter classes) if needed
     reply_message = _("Your maximum daily usage time has been reached")
-    gigawords = False
 
     def __init__(self, user, group, group_check):
         self.user = user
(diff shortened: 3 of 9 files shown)


openwisp_radius/counters/base.py
```diff
@@ -72,7 +79,7 @@ def get_attribute_type(self):
 
     def get_reset_timestamps(self):
         try:
-            return resets[self.reset](self.user)
+            return resets[self.reset](self.user, self)
```
pandafy: Passing the counter object to the reset allows flexibility to perform operation on user, group and group_check objects.

pandafy:
@nemesifier can you suggest me how can I incorporate this implementation detail in https://openwisp.io/docs/dev/radius/user/enforcing_limits.html

openwisp_radius/counters/base.py
```diff
@@ -21,9 +21,17 @@ def check_name(self):  # pragma: no cover
...
 
     @property
-    @abstractmethod
-    def reply_name(self):  # pragma: no cover
-        pass
+    def reply_names(self):
+        # BACKWARD COMPATIBILITY: In previous versions of openwisp-radius,
+        # the Counter.reply_name was a string instead of a tuple. Thus,
+        # we need to convert it to a tuple if it's a string.
+        if not hasattr(self, "reply_name"):
```
nemesifier: AFAIK, `getattr` and `hasattr` have the same computing cost, so it's better to use `getattr` if we need to use the result later.
```suggestion
        reply_name = getattr(self, "reply_name", None)
        if not reply_name:
```

openwisp_radius/counters/base.py
```diff
@@ -21,9 +21,17 @@ def check_name(self):  # pragma: no cover
...
+            raise NotImplementedError(
+                "Counter classes must define 'reply_names' property."
+            )
+        if isinstance(self.reply_name, str):
```
nemesifier:
```suggestion
        if isinstance(reply_name, str):
```

openwisp_radius/counters/base.py
```diff
@@ -21,9 +21,17 @@ def check_name(self):  # pragma: no cover
...
+            return (self.reply_name,)
```
nemesifier:
```suggestion
            return (reply_name,)
```

openwisp_radius/counters/base.py
```diff
@@ -21,9 +21,17 @@ def check_name(self):  # pragma: no cover
...
+        return self.reply_name
```
nemesifier:
```suggestion
        return reply_name
```

openwisp_radius/counters/base.py
```diff
@@ -21,9 +21,17 @@ def check_name(self):  # pragma: no cover
         pass
 
     @property
-    @abstractmethod
-    def reply_name(self):  # pragma: no cover
-        pass
+    def reply_names(self):
```
nemesifier, line 24: I don't understand very well if this method is really needed here? Where is the logic below used?

openwisp_radius/utils.py
```diff
@@ -344,41 +344,50 @@ def execute_counter_checks(
...
-            counter_data, counter
-        ):
-            counter_data[reply_name] = remaining
+
+        # BACKWARD COMPATIBILITY: The previous implementation of counters
+        # returned a single value instead of a tuple/list when there was
+        # only one reply name defined.
+        # We need to handle this case to avoid breaking existing counters.
+        if not isinstance(results, (list, tuple)):
+            results = (results,)
```
nemesifier, line 352: In the code of the counters, you check if it's a string and only then convert it to a tuple. Here you're checking if it's not a tuple or a list and convert it to a tuple. I think this inconsistent. Is it on purpose? If so, what's the reason?

openwisp_radius/counters/base.py
pandafy: The logic is present for backward compatibility, i.e. if there are custom Counters which define `reply_name` instead of `reply_names`, then this method will ensure that those Counters do not break.

openwisp_radius/utils.py
pandafy: > In the code of the counters, you check if it's a string and only then convert it to a tuple

That code provides backward compatibility to the existing custom counters. 

>  Here you're checking if it's not a tuple or a list and convert it to a tuple. 

The new logic expects both `reply_names` and output of Counter.check()` to be a tuple. Thus, this code is for backward compatibility if any existing custom counter returns a single value.

.github/workflows/ci.yml
```diff
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches:
       - master
+      # TODO: Remove before merging
+      - issues/643-coa
```
nemesifier:
```suggestion
```

openwisp_radius/api/serializers.py
```diff
@@ -304,12 +304,10 @@ def get_result(self, obj):
...
             )
-            # Python can handle 64 bit numbers and
-            # hence we don't need to display Gigawords
-            remaining = counter.check(gigawords=False)
-            return int(obj.value) - remaining
-        except MaxQuotaReached:
-            return int(obj.value)
+            consumed = counter.consumed()
+            if consumed > int(obj.value):
+                consumed = int(obj.value)
```
nemesifier:
```suggestion
            value = int(obj.value)
            if consumed > value:
                consumed = value
```

openwisp_radius/counters/base.py
```diff
@@ -93,7 +101,7 @@ def get_counter(self):
         # or if nothing is returned (no sessions present), return zero
         return row[0] or 0
 
-    def check(self, gigawords=gigawords):
+    def check(self, gigawords=True):
```
nemesifier:
```suggestion
    def check(self):
```

openwisp_radius/counters/resets.py
```diff
@@ -11,22 +11,22 @@ def _timestamp(start, end):
     return int(start.timestamp()), int(end.timestamp())
 
 
-def _daily(user=None):
+def _daily(user=None, counter=None):
```
nemesifier:
```suggestion
def _daily(user=None, **kwargs):
```

openwisp_radius/counters/resets.py
```diff
@@ -11,22 +11,22 @@ def _timestamp(start, end):
...
     dt = _today()
     start = datetime(dt.year, dt.month, dt.day)
     end = datetime(dt.year, dt.month, dt.day) + timedelta(days=1)
     return _timestamp(start, end)
 
 
-def _weekly(user=None):
+def _weekly(user=None, counter=None):
```
nemesifier:
```suggestion
def _weekly(user=None, **kwargs):
```

openwisp_radius/counters/resets.py
```diff
@@ -11,22 +11,22 @@ def _timestamp(start, end):
...
     dt = _today()
     start = dt - timedelta(days=dt.weekday())
     start = datetime(start.year, start.month, start.day)
     end = start + timedelta(days=7)
     return _timestamp(start, end)
 
 
-def _monthly(user=None):
+def _monthly(user=None, counter=None):
```
nemesifier:
```suggestion
def _monthly(user=None, **kwargs):
```

openwisp_radius/counters/resets.py
nemesifier:
```suggestion
def _monthly_subscription(user, **kwargs):
```

openwisp_radius/counters/resets.py
```diff
@@ -45,7 +45,7 @@ def _monthly_subscription(user):
     return _timestamp(start, end)
 
 
-def _never(user=None):
+def _never(user=None, counter=None):
```
nemesifier:
```suggestion
def _never(user=None, **kwargs):
```

nemesifier (approved):

PR #665, opened 2025-12-26
[monitoring/js] Consistent datetime format in RADIUS sessions tab

Eeshu-Yadav:
## Checklist

Use Django locale for date formatting in device page RADIUS sessions tab JS, matching Django admin and localization settings.

Fixes #662

### DIFF
--- a/openwisp_radius/api/urls.py
+++ b/openwisp_radius/api/urls.py
@@ -8,7 +8,7 @@ def get_api_urls(api_views=None):
     if not api_views:
         api_views = views
     if app_settings.RADIUS_API:
-        return [
+        api_urls = [
             path("freeradius/authorize/", api_views.authorize, name="authorize"),
             path("freeradius/postauth/", api_views.postauth, name="postauth"),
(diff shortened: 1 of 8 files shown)


pandafy (changes requested):
I don't think it is working as expected. 

The created/modified timestamp of the Device is in different format from the "Start Time" of the RADIUS session 


Check the formatting of "Created" and "Modified" datetime of the Device object:
Image

Check the formatting of "Start time" field in the "RADIUS Sessions" tab: 

Image

Check if something like this will solve the problem:

https://github.com/openwisp/openwisp-controller/blob/115cd9ac1b418204bdb704c02af308569e19e3fd/openwisp_controller/connection/static/connection/js/commands.js#L631-L638

Eeshu-Yadav:
@stktyagi i am working on that as suggested by @pandafy ,

nemesifier (commented):
@Eeshu-Yadav once you're done, please post a screenshot of the UI result in high res so we can verify the visual result matches our expectations (as @pandafy explained). Thanks 👍

PR #678, opened 2026-02-05
[tests] Increase test coverage #677

shivsubh:
Implement tests for RadiusBatchConsumer to verify authentication, authorization, and group messaging logic.
Increases coverage for consumers.py to 84%.

## Checklist

Closes #677


openwisp_radius/consumers.py
```diff
@@ -12,13 +12,9 @@ def _user_can_access_batch(self, user, batch_id):
         if user.is_superuser:
             return RadiusBatch.objects.filter(pk=batch_id).exists()
         # For non-superusers, check their managed organizations
-        try:
-            RadiusBatch.objects.filter(
-                pk=batch_id, organization__in=user.organizations_managed
-            ).exists()
-            return True
-        except ObjectDoesNotExist:
-            return False
```
shivsubh, line 21: This seems like a bug, because it will never raise `ObjectDoesNotExist`

.github/workflows/ci.yml
```diff
@@ -99,6 +103,12 @@ jobs:
         if: ${{ failure() }}
         run: cat geckodriver.log
 
+      - name: Comment coverage
+        if: ${{ success() }}
+        uses: MishaKav/pytest-coverage-comment@v1.2.0
+        with:
+          pytest-coverage-path: coverage.xml
```
shivsubh: I added this so that we know about coverage in the PR. Let me know if i need to remove this.

nemesifier: Coveralls already takes care of this, so I think this is not needed.

openwisp_radius/tests/test_consumers.py
```diff
@@ -0,0 +1,174 @@
...
+
+        batch = self._create_radius_batch(
+            name="test-batch",
+            strategy="prefix",
+            prefix="test-",
+            organization=org,
+        )
+        return org, user, batch
+
+    def _create_staff_user(self, org):
```
nemesifier: Let's avoid  duplicating test logic as much as possible, try reusing:
https://github.com/openwisp/openwisp-users/blob/master/openwisp_users/tests/utils.py#L32-L170

openwisp_radius/tests/test_consumers.py
```diff
@@ -0,0 +1,174 @@
...
+        user = User.objects.create_user(
+            username="staff_user",
+            email="staff@test.com",
+            password=self.TEST_PASSWORD,
+            is_staff=True,
+        )
+        org.add_user(user, is_admin=True)
+        return user
+
+    def _create_regular_user(self):
```
nemesifier: same here

openwisp_radius/consumers.py
nemesifier: Good catch, I think you're right.

nemesifier (changes requested):
Thank you for this!

Covearlls has reported an increase to 88.45% (+0.3%) from 88.152%, pretty negligible. Was this your aim?

.github/workflows/ci.yml
shivsubh: Yes, have removed this

shivsubh:
> Thank you for this!
> 
> Covearlls has reported an increase to 88.45% (+0.3%) from 88.152%, pretty negligible. Was this your aim?

No, in my local it showed 95%, I am working on adding more tests.

.github/workflows/ci.yml
```diff
@@ -79,15 +79,15 @@ jobs:
...
-          ./run-qa-checks
+      # - name: QA checks
+      #   run: |
+      #     ./run-qa-checks
 
       - name: Tests
         if: ${{ !cancelled() && steps.deps.conclusion == 'success' }}
         run: |
-          coverage run runtests.py --parallel > /dev/null 2>&1 || ./runtests.py
-          SAMPLE_APP=1 coverage run ./runtests.py --parallel > /dev/null 2>&1 || SAMPLE_APP=1 ./runtests.py
```
shivsubh, line 90: Looks like `--parallel` was the problem all along. I feel, it is failing silently and generating out bad coverage data, but once I removed it, we hit 98% coverage.

Should I just open a clean PR for this change first? It’d be good to get the reporting fixed before I start adding the new tests.

nemesifier (commented):
Ok to open a new PR to investigate the parallel issue

PR #691, opened 2026-03-07
[docs] Document Django Channels WebSocket API endpoints

iamavichal-geek:
Closes #682

Changes:
- `docs/user/websocket-api.rst` — new page documenting the RadiusBatch
  WebSocket endpoint (modelled after openwisp-firmware-upgrader)
- `docs/index.rst` — registered in toctree
- `docs/user/generating_users.rst` — added WebSocket cross-reference
- `docs/user/importing_users.rst` — added WebSocket cross-reference
- `docs/user/settings.rst` — added label + note to
  OPENWISP_RADIUS_BATCH_ASYNC_THRESHOLD


nemesifier (changes requested):
Hi @iamavichal-geek, thanks for taking care of this! The WebSocket API was indeed undocumented and this is a very welcome addition.

I checked the docs against the implementation in `routing.py` and `consumers.py` and the endpoint URL, message format, and auth requirements look correct.

A couple of things before we merge:

1. Please make sure the docs build cleanly with Sphinx so we do not introduce any RST warnings. Double-check that the `:ref:` targets (like `radius_websocket_api` and `openwisp_radius_batch_async_threshold`) all resolve correctly.
2. The authorization docs describe org-scoped access control accurately, but the current `_user_can_access_batch` implementation in `consumers.py` has a pre-existing bug: the result of `.exists()` is discarded and the method always returns `True` for non-superusers, meaning any staff user can currently access any batch regardless of organization. Since you are already touching the feature, it would be great if you could fix this here. It is essentially a one-line change: return the result of `.exists()` directly and remove the unnecessary `return True` that follows. The `ObjectDoesNotExist` guard around it is also not needed because `.exists()` never raises it.

If you prefer to keep this PR strictly docs-only, no problem, but please open a separate issue for the consumers bug so we do not lose track of it.

With the docs build verified and the auth bug addressed (or ticketed), this is good to go.

nemesifier (approved):
Approved. The WebSocket API documentation is now aligned with the established OpenWISP WebSocket docs style: client messages are explicitly documented as ignored, server updates use the Real-time Updates heading, auth wording matches the staff requirement, and the deployment setup section was removed because the installer handles it.

One code issue remains outside this docs-only review scope: the consumer authorization check still needs the .exists() return-value fix.

PR #698, opened 2026-03-30
[feature] Made RegisteredUser model support multi-tenancy #692

pandafy:
## Checklist

Closes #692

## Description of Changes

- Updated the RegisteredUser model to support organization-specific records.
- Changed the primary key to UUID and added organization as a nullable ForeignKey.
- Modified related code across the application to handle multiple registered users per organization.
- Updated tests to reflect changes in the RegisteredUser model and ensure proper functionality.
- Added migration scripts to handle the transition from the old model to the new schema.

### Breaking Changes

- Changed ``registered_users`` CSV export column format from two flat
  columns (``registered_user.method``, ``registered_user.is_verified``)
  to a single nested column (``registered_users``) containing
  ``((organization_id,method,is_verified))`` tuples. External integrations
  consuming this CSV output will need to be updated.

## Blockers

- [x] https://github.com/openwisp/openwisp-users/pull/498


openwisp_radius/migrations/0043_registereduser_add_uuid.py
pandafy, line 1: ⚠️ **WARNING:** This is AI generated migration file which uses raw SQL. I'll update this migration to follow the Django migration framework.

pandafy: SQLite cannot modify primary keys in place, so Django works around this by rebuilding the entire table when a PK or certain fields are altered. During your migration, the table rebuild temporarily ends up with both the old implicit PK and the new id column, which leads to the duplicate column name: id error. This typically happens when adding a new PK column and removing the old PK within the same schema transition. Splitting the operations into separate migrations avoids the conflict because each table rebuild sees a consistent schema.

.github/workflows/ci.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           pip install -U pip wheel setuptools
           pip install -U -r requirements-test.txt
           pip install -e .[saml,openvpn_status]
+          pip install --upgrade --no-deps --no-cache-dir --force-reinstall "https://github.com/openwisp/openwisp-users/tarball/issues/497-export-users"
```
pandafy, line 77: TODO: Remove this before merging!

.github/workflows/ci.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           pip install -U pip wheel setuptools
           pip install -U -r requirements-test.txt
           pip install -e .[saml,openvpn_status]
+          pip install --upgrade --no-deps --no-cache-dir "https://github.com/openwisp/openwisp-users/tarball/issues/497-export-users"
```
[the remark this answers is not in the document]
pandafy: @coderabbitai we will remove this from the CI file once the related PR in openwisp-users is merged. You can ignore this.

openwisp_radius/tests/test_admin.py
```diff
@@ -1407,6 +1407,20 @@ def test_inline_registered_user(self):
...
+        RegisteredUser.objects.create(
+            user=user, organization=self.default_org, is_verified=True
+        )
+        RegisteredUser.objects.create(user=user, organization=org2, is_verified=False)
+        RegisteredUser.objects.create(user=user, organization=None, is_verified=True)
+        user_url = reverse(f"admin:{User._meta.app_label}_user_change", args=[user.pk])
+        response = self.client.get(user_url)
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "id_registered_users-TOTAL_FORMS")
+        self.assertIn('value="3"', response.rendered_content)
```
pandafy, line 1422: Let's check for this instead 

```
<input type="hidden" name="registered_users-INITIAL_FORMS" value="1" id="id_registered_users-INITIAL_FORMS">
```

openwisp_radius/tests/test_api/test_rest_token.py
```diff
@@ -402,3 +407,122 @@ def test_validate_auth_token_password_expired(self):
         User.objects.update(password_updated=now() - timedelta(days=60))
         response = self._test_validate_auth_token_helper(user)
         self.assertEqual(response.data["password_expired"], True)
+
+    @capture_any_output()
+    @mock.patch("openwisp_radius.utils.SmsMessage.send")
+    def test_multi_org_phone_verification_flow(self, *args):
```
pandafy, line 416: Check if we can shorten the code required for this test case. Utlise any pre-existing helper method.

openwisp_radius/tests/test_api/test_rest_token.py
```diff
@@ -241,11 +247,10 @@ def test_unverified_registered_user_different_organization(self):
             self.assertIn("key", response.data)
 
         OrganizationUser.objects.filter(organization=org2, user=user).delete()
+        RegisteredUser.objects.filter(user=user, organization=org2).delete()
         with self.subTest(
             "Test unverified user organization does not need identity verification"
         ):
-            registered_user.is_verified = False
-            registered_user.save()
```
pandafy, line 246: Let's reverse this change. Set all `is_verified=False` to all `RegisteredUser`.

openwisp_radius/tests/test_api/test_phone_verification.py
```diff
@@ -223,9 +224,23 @@ def test_create_phone_token_400_user_already_verified(self):
...
+
+        with self.subTest("global verified record also blocks phone token creation"):
+            # Replace org-specific record with a global verified record
+            reg_user.delete()
+            RegisteredUser.objects.create(
+                user=token.user, organization=None, is_verified=True
+            )
+            r = self.client.post(url, HTTP_AUTHORIZATION=f"Bearer {token.key}")
+            self.assertEqual(r.status_code, 400)
+            self.assertEqual(r.json(), {"user": "This user has been already verified."})
```
pandafy, line 243: I don't think it is correct. Only the org-specific RegisteredUser object should block verification of phone number.

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -258,7 +340,7 @@ def test_postauth_radius_token_accept_201(self):
     def test_postauth_accept_201_querystring(self):
         self.assertEqual(RadiusPostAuth.objects.all().count(), 0)
         params = self._get_postauth_params()
-        post_url = f'{reverse("radius:postauth")}{self.token_querystring}'
+        post_url = f"{reverse('radius:postauth')}{self.token_querystring}"
```
pandafy, line 343: unnecessary change

openwisp_radius/tests/test_api/test_freeradius_api.py
```diff
@@ -2442,7 +2524,7 @@ def test_cache(self):
         )
         self._get_org_user()
         token_querystring = f"?token={rad.token}&uuid={str(self.org.pk)}"
-        post_url = f'{reverse("radius:authorize")}{token_querystring}'
+        post_url = f"{reverse('radius:authorize')}{token_querystring}"
```
pandafy, line 2527: unnecessary change

openwisp_radius/api/views.py
```diff
@@ -772,6 +776,7 @@ def post(self, request, *args, **kwargs):
                 },
             )
             reg_user.is_verified = True
+            reg_user.method = "mobile_phone"
```
pandafy, line 777: Is this really required?

openwisp_radius/base/models.py
```diff
@@ -1058,10 +1058,20 @@ def save_user(self, user):
...
+                "is_verified": self.organization.radius_settings.needs_identity_verification,
+            },
+        )
+        if (
+            not created
+            and self.organization.radius_settings.needs_identity_verification
+        ):
             registered_user.is_verified = True
-        registered_user.save()
+            registered_user.save()
```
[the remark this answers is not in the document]
pandafy: @coderabbitai can you give me a test scenario?

openwisp_radius/api/views.py
```diff
@@ -747,23 +754,32 @@ def post(self, request, *args, **kwargs):
...
+            reg_user.is_verified = True
             # Update username if phone_number is used as username
             if user.username == user.phone_number:
                 user.username = phone_token.phone_number
             # now that the phone number is verified
             # we can write it to the user field
             user.phone_number = phone_token.phone_number
             user.save()
-            user.registered_user.save()
+            reg_user.save()
```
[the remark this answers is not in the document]
pandafy: @coderabbitai can you give me a test scenario for these bugs?

openwisp_radius/base/models.py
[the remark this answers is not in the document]
pandafy: @coderabbitai in your test case, an existing user was adding to creation of new RadiuBatch. Is this possible in real wold scenario?

openwisp_radius/api/views.py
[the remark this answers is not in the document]
pandafy: > Bug 2 — method is never updated on the existing record, leaving is_identity_verified_strong broken

@coderabbitai when the code reaches at this point, then the user has already verified the that they have entered the correct PhoneToken. They will only get the phone token if they have registered using their mobile number, which means the RegisteredUser.method already has "mobile_phone" in the datbase.

.github/workflows/ci.yml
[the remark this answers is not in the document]
pandafy: @coderabbitai this is temporary, ignore!

nemesifier:
@coderabbitai full review

openwisp_radius/tests/test_api/test_phone_verification.py
```diff
@@ -918,6 +946,54 @@ def test_phone_number_change_update_username(self):
...
+            response = self.client.post(
+                url,
+                json.dumps({"phone_number": "+595972157444"}),
+                content_type="application/json",
+                HTTP_AUTHORIZATION=f"Bearer {user_token.key}",
+            )
+        self.assertEqual(response.status_code, 200)
+
+        reg_org1.refresh_from_db()
+        self.assertEqual(reg_org1.is_verified, False)
```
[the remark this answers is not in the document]
pandafy: @nemesifier when I thought more about `test_phone_change_unverifies_only_specific_org`, the test looked wrong. Since, the phone number is saved in the User model changing the phone number will change it for all organizations. So, technically, we should unverify the user from all Organizations. 

But, doing so will make the user verify to every organization they had previously registered to. 

Do you think that the result is acceptable?

pandafy:
@coderabbitai review! Summarize if any previously flagged comment has been unresolved. Focus on real problems

nemesifier (changes requested):
Below are the items that I think still need to be addressed before merge.

## Bugs / security

### Cross-org verification leak in the FreeRADIUS authorize query

`openwisp_radius/api/freeradius_views.py:405-429` builds the verified-user
filter as:

```python
org_or_global = Q(registered_users__organization_id=organization_id) | Q(
    registered_users__organization__isnull=True
)
is_verified = Q(registered_users__is_verified=True) & org_or_global
```

Because this is a JOIN-based filter, a user with **any** registered_users row
satisfying the conditions passes. In particular, a user that has a global
record (`organization=NULL`) marked `is_verified=True` plus an org-specific
record for the org being authenticated marked `is_verified=False` is wrongly
authorized — exactly the bypass scenario the issue is trying to prevent.

`api/utils.py:IDVerificationHelper.is_identity_verified_strong` already has
the right behavior (prefer org-specific, fall back to global). The FreeRADIUS
side should match it: do an explicit lookup that uses the org-specific row
when present and only falls back to the global row when no org-specific row
exists. The same applies to the `authorize_unverified` branch a few lines
below.

`test_global_fallback_for_orgs_without_specific_records` and
`test_multi_org_user_different_verification_states` cover the "only one
record" cases but not the conflicting "global verified + org-specific
unverified" case.

### `ValidatePhoneTokenView` leaves the method as `pending_verification`

`openwisp_radius/api/views.py:766-782`:

```python
reg_user, __ = RegisteredUser.get_or_create_for_user_and_org(
    user=user,
    organization=self.organization,
    defaults={"is_verified": True, "method": "mobile_phone"},
)
reg_user.is_verified = True
...
reg_user.save()
```

`defaults` is only applied when the row is created. In the cross-org path the
row already exists with `method="pending_verification"` (created by
`ObtainAuthTokenView.validate_membership`), so after a successful SMS
verification the row ends up `is_verified=True` but `method=
"pending_verification"`. Because `pending_verification` is in
`_weak_verification_methods`, `is_identity_verified_strong` keeps returning
False and the user remains locked out. The expected flow currently relies on
the client calling `update_registered_user_registration_method` first; if
they don't, validation appears to succeed but login keeps returning 401 with
no clear remediation.

Either flip the method to `mobile_phone` here when the existing one is
`pending_verification`, or reject the validation request until the method has
been chosen. There is no test covering the "skipped UpdateRegisteredUserMethod
step" path.

### `post_save_radiusaccounting` orders NULLs the wrong way (Postgres)

`openwisp_radius/integrations/monitoring/tasks.py:189-198`:

```python
RegisteredUser.objects
  .only("method")
  .filter(user__username=username)
  .filter(Q(organization_id=organization_id) | Q(organization__isnull=True))
  .order_by("-organization_id")
  .first()
```

The intent (per surrounding code) is to prefer the org-specific row over the
global one. But on PostgreSQL the default for `ORDER BY x DESC` is `NULLS
FIRST`, so when both rows are present `.first()` returns the global one.
SQLite/MySQL actually behave the opposite way, so the result also depends on
the backend, which is worse. Use an explicit ordering — e.g.
`F("organization_id").desc(nulls_last=True)` — or two queries (org-specific
first, fall back to global).

### Resulting metric tag can be `None`

`clean_registration_method` was changed to return `None` for
`pending_verification` (`integrations/monitoring/utils.py`). The two
`_write_user_signup_metric_*` helpers correctly skip the `None` case, but
`post_save_radiusaccounting` does not — it assigns
`extra_tags["method"] = registration_method` regardless. A user whose
registration is still pending will produce metric points with a literal
`method=None` tag. Either skip writing the metric, or fall back to
`"unspecified"`.

### `social/views.py` saves twice after `get_or_create`

`openwisp_radius/social/views.py:51-58`:

```python
registered_user, created = RegisteredUser.objects.get_or_create(
    user=user, organization=org,
    defaults={"method": "social_login", "is_verified": False},
)
if created:
    registered_user.full_clean()
    registered_user.save()
```

`get_or_create` already persists the new row, so the `full_clean()`/`save()`
runs against an already-saved object. If `full_clean()` were to raise, the
record would already exist in the DB — the validation has no chance to block
insertion. The matching block in `saml/views.py` was already cleaned up; this
one should follow the same pattern (drop the redundant `full_clean()` and
`save()`, or do `validate_constraints`/`full_clean()` *before* the
`get_or_create`).

## Code quality / dead code

### Dead `except` in `UpdateRegisteredUserMethodView.post()`

`openwisp_radius/api/views.py:855-866`:

```python
try:
    reg_user = get_object_or_404(
        RegisteredUser, user_id=user.pk, organization=self.organization,
    )
except RegisteredUser.DoesNotExist:
    raise NotFound(...)
```

`get_object_or_404` raises `Http404`, not `DoesNotExist`, so the `except`
branch is unreachable. Either drop the try/except entirely (Http404 already
yields a 404 response with the swagger-documented message) or replace it with
`RegisteredUser.objects.get(...)` and catch `DoesNotExist` to produce the
custom message.

### Dead `except ObjectDoesNotExist` in `OpenwispRadiusSaml2Backend`

`openwisp_radius/saml/backends.py:14-33`. The old code accessed
`user.registered_user.method` (the OneToOne could raise
`ObjectDoesNotExist`); the new code uses
`user.registered_users.exclude(method="saml").exists()`, which can never
raise that exception. The `try/except ObjectDoesNotExist: pass` wrapper is
dead.

### Behavior change in `OpenwispRadiusSaml2Backend._update_user`

While in there: the new check is "skip the SAML username update if **any**
record on this user is non-SAML." With multi-tenancy this is more
conservative than the old "skip if *this* user's method is non-SAML". For a
user that signed up via mobile_phone in org A and via SAML in org B, the
SAML login on org B will no longer update their username, even though for
that org they did sign up via SAML. If that is intentional it deserves a
comment; otherwise the check should target the org being authenticated (the
SAML login already knows the org from the relay state).

### `AbstractRegisteredUser.clean()` mostly duplicates the unique constraint

`openwisp_radius/base/models.py:1693-1702` re-implements the per-org
uniqueness check in Python. The two `UniqueConstraint`s declared on the
model are validated by Django's `validate_constraints()` during
`full_clean()` (including the partial one with the `Q(organization__isnull=True)`
condition), so the dedicated `clean()` runs an extra query without
catching anything new. The one thing it does add is a custom error
message; if that's the goal, consider customizing the constraint
violation message via the `violation_error_message` argument on the
`UniqueConstraint` instead and dropping the extra query. Not blocking.

## Documentation / compatibility

### `export_users` CSV output format changed

`openwisp_radius/settings.py:232-241` now contributes a single
`registered_users` column whose value is rendered as
`((organization_id,method,is_verified))(...)` instead of the previous two
flat columns `registered_user.method` / `registered_user.is_verified`. The
test in `tests/test_users_integration.py` was rewritten to match. Anything
external that consumes that CSV will break, so this should be called out
in the release notes / upgrade guide.

### REST API doc has an awkward sentence

`docs/user/rest-api.rst` (Update Registered User Method section) says:
"The user must complete verification for that organization before they can
create account with the new organization." Reads oddly — should be e.g.
"before they can finish registering with the new organization" or similar.

## Tests / coverage gaps

- No test for the cross-org leak above (global verified + org-specific
  unverified) on the FreeRADIUS authorize path.
- No test for `ValidatePhoneTokenView` being called without first calling
  `update_registered_user_registration_method` (the bug above).
- `post_save_radiusaccounting` is not tested with both an org-specific and a
  global `RegisteredUser` for the same user.
- The end-to-end multi-org flow test
  (`test_multi_org_phone_verification_flow`) only verifies the happy path
  where the client follows the prescribed order (login → update method →
  verify); error/edge orderings are not covered.

openwisp_radius/tests/test_api/test_phone_verification.py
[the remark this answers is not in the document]
pandafy: If a user changes their phone number during cross-organization login, then we will flag the user as unverified in all Organizations where the user has registered using `mobile_phone` method. As soon as they verify the phone number in one organization, we will mark them as verified.

nemesifier (changes requested):
I tried saving 2 RegisteredUser instances with  the same org:

Image

The error message does not explain the problem very well.

I tried the same elsewhere:

Image

This validation error message is a lot better, can you make the two consistent?

openwisp_radius/migrations/__init__.py
```diff
@@ -1,19 +1,278 @@
...
+BATCH_SIZE = 1000
+
 
 def get_swapped_model(apps, app_name, model_name):
     model_path = swapper.get_model_name(app_name, model_name)
     app, model = swapper.split(model_path)
     return apps.get_model(app, model)
 
 
+def _batched_iterator(iterator, batch_size=BATCH_SIZE):
```
nemesifier, line 20: This is a lot of new logic in this file, is it used only in the last migrations or is it meant to be reusable?
It may make sense to move this to a dedicated file.

pandafy:
@coderabbitai full review!

openwisp_radius/api/freeradius_views.py
```diff
@@ -405,23 +406,45 @@ def _check_counters(self, data, user, group, group_checks):
...
         if not needs_verification:
             return is_active
-        # if identity verification is enabled
-        is_verified = Q(registered_user__is_verified=True)
+        organization_id = request._auth
         AUTHORIZE_UNVERIFIED = registration.AUTHORIZE_UNVERIFIED
-        # and no method should authorize unverified users
-        # ensure user is active AND verified
+        # Use subqueries to ensure org-specific records take precedence over
+        # global (organization=NULL) records.
```
nemesifier: is this still a thing?

openwisp_radius/base/models.py
```diff
@@ -1602,12 +1622,18 @@ def __check(self, token):
...
-    user = models.OneToOneField(
+class AbstractRegisteredUser(UUIDModel):
+    user = models.ForeignKey(
         settings.AUTH_USER_MODEL,
         on_delete=models.CASCADE,
-        related_name="registered_user",
-        primary_key=True,
+        related_name="registered_users",
+    )
+    organization = models.ForeignKey(
```
nemesifier: Why not just inherit `OrgMixin` as we have in pretty much all the django models of OpenWISP?

openwisp_radius/api/freeradius_views.py
```diff
@@ -405,23 +406,45 @@ def _check_counters(self, data, user, group, group_checks):
...
+        # A JOIN-based filter would allow a user to pass if ANY registered_users
+        # row matched, causing a bypass when a global verified record coexisted
+        # with an org-specific unverified record.
+        #
+        # Strategy: check if org-specific record exists and satisfies criteria;
+        # if not, fall back to checking the global record. This matches the
```
nemesifier: All this complexity for what exactly?

openwisp_radius/api/freeradius_views.py
```diff
@@ -405,23 +406,45 @@ def _check_counters(self, data, user, group, group_checks):
...
+            user=OuterRef("pk"),
+            organization_id=organization_id,
+        )
+        global_only = RegisteredUser.objects.filter(
+            user=OuterRef("pk"),
+            organization_id__isnull=True,
+        )
+
+        # is_verified: user passes if org-specific record is verified, or if
+        # no org-specific record exists and the global record is verified.
```
nemesifier: Also this code I think needs to be simplified as we only care for the org specific record and nothing else

docs/user/rest-api.rst
```diff
@@ -803,6 +803,48 @@ Param        Description
 phone_number string
 ============ ===========
 
+Update Registered User Method
```
nemesifier:
```suggestion
Update user registration method
```

docs/user/rest-api.rst
```diff
@@ -803,6 +803,48 @@ Param        Description
...
++++++++++++++++++++++++++++++
```
nemesifier:
```suggestion
+++++++++++++++++++++++++++++++
```

openwisp_radius/saml/backends.py
```diff
@@ -12,20 +11,27 @@ def _update_user(self, user, attributes, attribute_mapping, force_save=False):
...
-                        for attr in value:
-                            if attr != "username":
-                                attribute_mapping[key].append(attr)
-
-            except ObjectDoesNotExist:
-                pass
+            attribute_mapping = attribute_mapping.copy()
+            # Check if any of the user's registered_users records
+            # were NOT created via SAML.
+            # NOTE: This uses a global check (any org) rather than org-specific.
```
nemesifier, line 17: Also here the code still mentions the global settings as a real thing.

nemesifier (changes requested):
I spotted a few things that need to be improved.

openwisp_radius/api/freeradius_views.py
```diff
@@ -405,23 +406,45 @@ def _check_counters(self, data, user, group, group_checks):
...
+        # with an org-specific unverified record.
+        #
+        # Strategy: check if org-specific record exists and satisfies criteria;
+        # if not, fall back to checking the global record. This matches the
+        # behavior in api/utils.py:IDVerificationHelper.is_identity_verified_strong.
+        org_specific = RegisteredUser.objects.filter(
+            user=OuterRef("pk"),
+            organization_id=organization_id,
+        )
+        global_only = RegisteredUser.objects.filter(
```
nemesifier: This global fallback should be removed now that `organization=None` RegisteredUser rows are out of scope. The authorize endpoint can require the row for `request.auth` directly. That also lets us drop the correlated `Exists` subqueries from this hot path and use a much simpler organization scoped `Q` filter.

openwisp_radius/api/views.py
```diff
@@ -747,23 +755,32 @@ def post(self, request, *args, **kwargs):
                 _("No verification code found in the system for this user.")
             )
         try:
-            is_valid = phone_token.is_valid(serializer.data["code"])
+            is_valid = phone_token.is_valid(
```
nemesifier, line 765: Passing the organization into validation is not enough because the token selected above is still the latest token for the user, regardless of organization. A token created for org A can be submitted to org B and verify org B's RegisteredUser row. PhoneToken needs to store the requested organization, and token creation, active status, sending, and validation should all filter/use that organization.

openwisp_radius/integrations/monitoring/tasks.py
```diff
@@ -115,6 +117,8 @@ def _write_user_signup_metrics_for_orgs(metric_key):
     # The query returns a tuple of organization_id, registration_method and
     # count of users who registered with that organization and method.
     registered_users_query = RegisteredUser.objects.exclude(
```
nemesifier, line 120: The org metrics query below still groups through `user__openwisp_users_organizationuser__organization_id`. With per organization RegisteredUser rows, it should group by `RegisteredUser.organization_id` instead. For a user with two memberships and two RegisteredUser rows, the current join can produce every organization/method combination and count methods under the wrong organization. The `users_without_registereduser` query has the related problem: it only finds users with no RegisteredUser anywhere, not users missing a row for the specific organization.

openwisp_radius/saml/views.py
```diff
@@ -67,20 +68,21 @@ def post_login_hook(self, request, user, session_info):
...
-            user.registered_user
-        except ObjectDoesNotExist:
-            registered_user = RegisteredUser(
-                user=user, method="saml", is_verified=app_settings.SAML_IS_VERIFIED
+        with transaction.atomic():
+            if not is_member:
+                orgUser = OrganizationUser(organization=org, user=user)
+                orgUser.full_clean()
+                orgUser.save()
+            registered_user, created = RegisteredUser.get_or_create_for_user_and_org(
```
nemesifier, line 76: This only applies the SAML method and `SAML_IS_VERIFIED` when the org specific row is newly created. If the row already exists as `pending_verification`, for example after password login added the user to the org, a successful SAML login leaves it pending and possibly unverified. Existing pending rows should be updated to reflect the successful SAML login, while preserving stronger existing states.

openwisp_radius/social/views.py
```diff
@@ -42,18 +43,19 @@ def authorize(self, request, org, *args, **kwargs):
...
-            user.registered_user
-        except ObjectDoesNotExist:
-            registered_user = RegisteredUser(
-                user=user, method="social_login", is_verified=False
+        with transaction.atomic():
+            if not is_member:
+                orgUser = OrganizationUser(organization=org, user=user)
+                orgUser.full_clean()
+                orgUser.save()
+            registered_user, created = RegisteredUser.get_or_create_for_user_and_org(
```
nemesifier, line 51: Same issue as the SAML path: defaults only apply on create. If this org row already exists as `pending_verification`, successful social login leaves it pending instead of recording `social_login`. Please update pending rows explicitly or document why they should remain pending.

tests/openwisp2/sample_radius/migrations/0033_registered_user_multitenant.py
```diff
@@ -0,0 +1,211 @@
...
+                        (
+                            "organization",
+                            models.ForeignKey(
+                                blank=True,
+                                help_text=(
+                                    "The organization this registration info belongs"
+                                    " to."
+                                ),
+                                null=True,
+                                on_delete=django.db.models.deletion.CASCADE,
```
nemesifier, line 193: The sample swapped app migration creates this database column as nullable, but the state later declares it non nullable and there is no equivalent of core migration 0045 to alter the database column. Please add a database `AlterField` or follow up migration so swapped models cannot keep creating `organization=NULL` rows at the DB level.

nemesifier (commented):
Tests that would catch the issues mentioend in my other inline comments below:

- Add a multi organization SMS token test: create a token through organization A, then try to validate it through organization B. Organization B must not be verified by organization A's token.
- Add monitoring metric coverage with one user, two organization memberships, and two org specific RegisteredUser rows with different methods. Each organization should be counted only with its own method.
- Add SAML and social tests where an existing `pending_verification` row is updated after a successful org scoped login.
- Add a sample swapped app migration/schema test that confirms `RegisteredUser.organization` is non nullable at the database level.
- After optimizing the admin verified column, add a changelist query count test to prevent one `registered_users` query per row.

openwisp_radius/migrations/__init__.py
```diff
@@ -4,16 +4,298 @@
...
+    return Case(
+        When(method="pending_verification", then=Value(-1)),
+        When(method="", then=Value(0)),
+        When(method="email", then=Value(1)),
+        default=Value(2),
+        output_field=IntegerField(),
+    )
+
+
+def copy_registered_users_ctcr_forward(
```
nemesifier, line 55: These RegisteredUser migration helpers are large, migration specific, and imported by historical migrations. Keeping them in `migrations/__init__.py` makes old migrations depend on a mutable shared module. Please move them to a dedicated versioned helper module, treated as immutable once shipped, or inline them in the migrations that use them.

openwisp_radius/admin.py
```diff
@@ -549,12 +570,18 @@ def has_delete_permission(self, request, obj=None):
...
     PhoneTokenInline,
 ]
-UserAdmin.list_filter += (RegisteredUserFilter, "registered_user__method")
+UserAdmin.list_filter += (RegisteredUserFilter, "registered_users__method")
 
 
 def get_is_verified(self, obj):
     try:
-        value = "yes" if obj.registered_user.is_verified else "no"
+        is_verifieds = obj.registered_users.values_list("is_verified", flat=True)
```
nemesifier: This introduces one `registered_users` query per user row in the admin changelist. With the new one to many relation, please prefetch or annotate the verification state in the user admin queryset and read that cached value here.

openwisp_radius/social/views.py
```diff
@@ -42,18 +43,19 @@ def authorize(self, request, org, *args, **kwargs):
...
+                user=user,
+                organization=org,
+                defaults={"method": "social_login", "is_verified": False},
             )
-            registered_user.full_clean()
-            registered_user.save()
+            if not created:
```
nemesifier, line 56: This branch validates and saves an unchanged RegisteredUser row, so it is a no-op database write. It can be removed, or replaced with explicit pending row update logic if that is the intended behavior.

openwisp_radius/saml/views.py
```diff
@@ -67,20 +68,21 @@ def post_login_hook(self, request, user, session_info):
...
+                defaults={
+                    "method": "saml",
+                    "is_verified": app_settings.SAML_IS_VERIFIED,
+                },
             )
-            registered_user.full_clean()
-            registered_user.save()
-            # The user is just created, it will not have an email address
-            if user.email:
+            if created and user.email:
```
nemesifier: `created` here refers to the RegisteredUser row, not the user. On cross organization SAML login, the user may already have this email address, so creating a new EmailAddress can raise a validation error and log a misleading exception even though the login continues. Please update the comment and use `get_or_create()` or check for an existing EmailAddress first.

docs/user/rest-api.rst
```diff
@@ -803,6 +803,48 @@ Param        Description
...
+**Requires the user auth token (Bearer Token)**.
+
+Allows users to update their registered user method for an organization.
+The method can only be updated when it is currently set to
+``pending_verification``. Once updated, it cannot be changed again via
+this endpoint.
+
+This endpoint is used during cross-organization login when a user
+authenticates to a new organization. The user must complete verification
+for that organization before they can create account with the new
```
nemesifier, line 818: This describes the flow inaccurately. The implementation already creates the OrganizationUser and a pending RegisteredUser row during cross organization login when registration is enabled; access remains denied until org specific verification is completed. Please reword this to avoid saying the account is created only after verification.

openwisp_radius/saml/backends.py
```diff
@@ -12,20 +11,27 @@ def _update_user(self, user, attributes, attribute_mapping, force_save=False):
...
+            # This is intentionally conservative: if a user has ever signed up
+            # via a non-SAML method in any org, their username won't be updated
+            # during SAML login in any org. This prevents the SAML identity
+            # provider from overwriting a username set or preferred by the user
+            # elsewhere. Since the User model is shared across organizations,
+            # updating the username based solely on one org's SAML flow could
+            # unexpectedly change the user's identity in other orgs.
+            has_non_saml = user.registered_users.exclude(method="saml").exists()
```
nemesifier, line 25: This may be intentional because `User.username` is global, but it deserves an explicit decision with the new per organization registration model. A user who registered with a non SAML method in org A will not have their username updated when logging in with SAML to org B. If this global policy is intended, please make the comment clear that this is about the shared User model, not leftover global RegisteredUser fallback behavior.

openwisp_radius/migrations/__init__.py
pandafy: We re-use this in the sample app.

openwisp_radius/saml/backends.py
pandafy: Here, global does not mean global RegisteredUser. As you can see that the queryset does not check (organization=None), but it checks if the user was not registered with SAML.

openwisp_radius/api/serializers.py
```diff
@@ -753,20 +755,64 @@ def save(self):
...
+
+    class Meta:
+        model = RegisteredUser
+        fields = ["method"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["method"].choices = get_registration_choices()
+
+    def validate_method(self, value):
```
nemesifier, line 788: I do not think we can blindly trust all registered method values here. Some values are not normal user choices: they represent server side flows which should set the method only after that flow actually happened. Practical examples are `saml` and `social_login`.

Right now, could a user submit `{"method": "saml"}` or `{"method": "social_login"}` here? If yes, the DB would say the user registered with SAML or social login even though neither flow happened. Maybe this does not lead to an immediate bypass today, but it is still unsafe to let users write values which other parts of the system may treat as trusted registration state.

Please let me know what you think. If you think it cannot be abused, explain why. Otherwise let's discuss the right way to restrict this endpoint before changing it, because a quick allowlist/denylist could also be wrong if we do not define the intended flow clearly.

openwisp_radius/integrations/monitoring/tasks.py
```diff
@@ -115,6 +117,8 @@ def _write_user_signup_metrics_for_orgs(metric_key):
...
+        method="pending_verification"
+    ).exclude(
         user__openwisp_users_organizationuser__created__gt=end_time,
```
nemesifier, line 124: The grouping was fixed to use `RegisteredUser.organization_id`, good. But the time filter still goes through `user__openwisp_users_organizationuser__created`, which can match any membership of the user, not necessarily the membership for the same organization as this `RegisteredUser` row.

Example: a user joined org A months ago and joins org B now. The current query can count the org A `RegisteredUser` again in `user_signups`, because the user has a recent membership in org B.

The `users_without_registereduser_query` below has the same org scoping problem: it only counts users with no `RegisteredUser` anywhere, so a user with a row in org A but no row in org B will not be counted as `unspecified` for org B.

openwisp_radius/base/admin_filters.py
```diff
@@ -14,8 +15,15 @@ def lookups(self, request, model_admin):
...
+        if self.value() is None:
+            return queryset
+        where = Q()
+        if not request.user.is_superuser:
+            where = Q(
+                registered_users__organization__in=request.user.organizations_managed
+            )
         if self.value() == "unknown":
-            return queryset.filter(registered_user__isnull=True)
+            where &= Q(registered_users__isnull=True)
```
nemesifier, line 43: This does not look right for organization managers. For non superusers, `where` already requires `registered_users__organization__in=request.user.organizations_managed`; adding `registered_users__isnull=True` means both conditions must be true on the same join, which cannot happen.

This also does not match the admin display logic, where a user is shown as unknown when there is no `RegisteredUser` in the manager's organizations. The test added around `openwisp_radius/tests/test_admin.py:1687-1690` seems to encode the wrong expectation: `org2_only` is displayed as unknown for the manager, so the `unknown` filter should include it.

I think `unknown` means no `RegisteredUser` exists for this user in the managed organizations, using `~Exists(...)` or an equivalent filtered relation. Is my understanding incorrect? If so, please let me know.

openwisp_radius/saml/views.py
```diff
@@ -67,30 +68,61 @@ def post_login_hook(self, request, user, session_info):
...
+                registered_user.save()
             if user.email:
                 try:
-                    email_address = EmailAddress(
-                        user=user, email=user.email, primary=True, verified=True
+                    user_has_primary_email = EmailAddress.objects.filter(
+                        user=user, primary=True
+                    )
+                    email_address, email_created = EmailAddress.objects.get_or_create(
+                        user=user,
```
nemesifier: `EmailAddress.objects.get_or_create()` saves the new row before `email_address.full_clean()` is called below. Is this on purpose and if so, why? It looks wrong to me because an invalid `user.email` would already have been written by the time validation runs.

nemesifier (changes requested):
I did some testing with the related WLP branch and I haven't found bugs up to now but I still have some concerns which I explained inline.

One general note: I had asked to move the new migration helpers out of `migrations/__init__.py`, but I realize it's quite futile as there is already a lot there. So there's nothing to do there now, but please keep in mind that this file is becoming huge and we shouldn't keep adding more as it makes maintenance harder. I wrote about squashing old migrations here: https://github.com/openwisp/openwisp-radius/issues/705#issuecomment-4468323568

openwisp_radius/api/serializers.py
pandafy: @nemesifier the [RegisterSerializer](https://github.com/openwisp/openwisp-radius/blob/bf61cdb117757c91369e5085003862bab9130149/openwisp_radius/api/serializers.py#L550) already accepts a method field 

https://github.com/openwisp/openwisp-radius/blob/bf61cdb117757c91369e5085003862bab9130149/openwisp_radius/api/serializers.py#L568-L575

So the tightening of the allowed registration choices should also apply there.

nemesifier: Makes sense. We can add another setting which specifies with registration methods are self initiated.

openwisp_radius/api/serializers.py
```diff
@@ -571,9 +570,13 @@ class RegisterSerializer(
             'verification in its "Organization RADIUS Settings."'
         ),
         default="",
-        choices=REGISTRATION_METHOD_CHOICES,
+        choices=(),
     )
 
+    def __init__(self, instance=None, data=..., **kwargs):
```
nemesifier: This `data=...` default is not DRF's `empty` sentinel. Instantiating `RegisterSerializer()` without data passes `Ellipsis` as data, so the serializer becomes bound with invalid `initial_data`. This can break schema generation, browsable API behavior, or any code expecting an unbound serializer. I would use `*args, **kwargs` or the DRF `empty` sentinel already imported in this file.

openwisp_radius/api/serializers.py
```diff
@@ -753,20 +758,64 @@ def save(self):
...
+            "The registration method to set for the user. "
+            "Cannot be 'pending_verification'."
+        ),
+    )
+
+    class Meta:
+        model = RegisteredUser
+        fields = ["method"]
+
+    def __init__(self, instance=None, data=..., **kwargs):
```
nemesifier: Same issue here: `data=...` makes `UpdateRegisteredUserMethodSerializer()` bound even when no data was provided. I verified `initial_data == Ellipsis` and `is_valid() == False` on an empty serializer instance. Please use `*args, **kwargs` or DRF's `empty` sentinel.

openwisp_radius/registration.py
```diff
@@ -58,3 +60,36 @@ def unregister_registration_method(name, fail_loud=True):
...
+    if invalid_methods:
+        raise ImproperlyConfigured(
+            "OPENWISP_RADIUS_USER_SETTABLE_REGISTRATION_METHODS contains unknown "
+            f"values: {', '.join(repr(method) for method in invalid_methods)}"
+        )
+
+    return [(method, available_choices[method]) for method in methods]
+
+
+validate_user_settable_registration_methods(
```
nemesifier: This validation runs at module import time, before external apps can register custom methods with `register_registration_method()`. A project setting `OPENWISP_RADIUS_USER_SETTABLE_REGISTRATION_METHODS = ["custom_identity"]` and registering that method from an app will fail startup because the method is still unknown here. The test covers manual validation after registering the custom method, but the real startup flow is different. I think this should be delayed until registrations are complete, or moved to a system check.

nemesifier (changes requested):
I found two outstanding issues that should be addressed before merging.

nemesifier (approved):
Thanks @pandafy 👏

PR #699, opened 2026-03-30
Fix: handle missing weasyprint dependency in generate_pdf.

manshisingh31:
Problem:
generate_pdf crashes if weasyprint is not installed because HTML is None.

Solution:
- Added safe import using try-except
- Added check inside generate_pdf to raise a clear Import Error.


nemesifier:
Unsolicited, not a real issue.

PR #719, opened 2026-05-13
[change!] Aligned user expiration handling with openwisp-users  #706

### LINKED ISSUE #706: [change!] Align user expiration handling with openwisp-users
User expiration logic (based on an `expiration_date` field) will be implemented in [openwisp-users#499](https://github.com/openwisp/openwisp-users/issues/499). Currently, openwisp-radius contains its own logic to deactivate expired users created via RadiusBatch, which leads to duplication.

- Remove all logic in openwisp-radius that deactivates expired users:
  - Management command: `deactivate_expired_users`
  - Celery task: `deactivate_expired_users` in `tasks.py`
  - The `expire()` method in the RadiusBatch model
- Retain the `expiration_date` field in the RADIUS batch user model:
  - This field will only be used to assign the expiration date to generated user accounts
  - It should not be used for enforcing expiration within openwisp-radius
- Ensure that user objects created via RADIUS batch correctly receive the `expiration_date` from the batch when they are saved.
- When migrating, set the User.expiration_date field for the users
created as part of RadiusBatch. This will ensure that these users
will be expired when the celery beat executes the new task from openwisp_users.

**Additional context**
- The `delete_old_radiusbatch_users` command should remain as it serves a different purpose (deleting users after 18 months, not just deactivating).
- Users of openwisp-radius will need to remove the `deactivate_expired_users` and `delete_old_radiusbatch_users` entries from their `CELERY_BEAT_SCHEDULE`. This is a breaking change change which should be documented in the changelog.
- These beat schedule entries will be replaced by the new taskadded in openwisp-users.

**Blockers**
- [x] https://github.com/openwisp/openwisp-users/issues/499

pandafy:
## Checklist

Closes #706

## Description of Changes

Moved expiration of user logic to openwisp-users

## Blockers 

- [ ] https://github.com/openwisp/openwisp-users/pull/517


tests/openwisp2/sample_users/migrations/0001_initial.py
```diff
@@ -34,6 +34,16 @@ class Migration(migrations.Migration):
                         blank=True, null=True, verbose_name="last login"
                     ),
                 ),
+                (
+                    "expiration_date",
+                    models.DateField(
+                        blank=True,
+                        help_text="Date on which the user account will expire.",
```
nemesifier: This still uses the old help text for `expiration_date`. With the related `openwisp-users` branch installed, `./run-qa-checks` detects this as a missing sample app migration (`0005_alter_user_expiration_date`) and fails. I think we should keep this migration aligned with the current `openwisp-users` field definition, including the new help text.

openwisp_radius/base/models.py
```diff
@@ -1060,6 +1060,8 @@ def get_or_create_user(self, row, users_list, password_length):
     def save_user(self, user):
         OrganizationUser = swapper.load_model("openwisp_users", "OrganizationUser")
         RegisteredUser = swapper.load_model("openwisp_radius", "RegisteredUser")
+        if self.expiration_date is not None:
+            user.expiration_date = self.expiration_date
```
nemesifier, line 1087: Here we set `user.expiration_date` after the user has already been validated, then call `save()` directly. This means a new batch can still create active users with an expiration date in the past, bypassing the validation added in `openwisp-users`. I think new batch creation should follow the same rule and reject past expiration dates, while historical data can still be handled by the migration.

openwisp_radius/tests/test_migrations.py
```diff
@@ -453,3 +453,94 @@ def test_convert_user_credentials_data_invalid_json(self):
...
+        user = self._create_user(
+            username="batch-copy-preserve-user",
+            expiration_date=original_expiration_date,
+        )
+        batch.users.add(user)
+        self._copy_batch_expiration_to_user()
+        user.refresh_from_db()
+        self.assertEqual(user.expiration_date, original_expiration_date)
+
+    def test_copy_batch_expiration_to_user_prefers_latest_batch(self):
```
nemesifier, line 521: This test makes sense because it treats the user's own `expiration_date` as the source of truth when a user belongs to multiple batches. The cleanup command does not follow the same rule though: it still deletes whole batches by `RadiusBatch.expiration_date`, and `RadiusBatch.delete()` deletes all attached users. This can delete a user whose account expiration was later extended or copied from a newer batch. I think `delete_old_radiusbatch_users` should take each user's own `expiration_date` into account and avoid deleting users whose expiration is empty or later than the cleanup threshold.

nemesifier (changes requested):
I think these issues need to be addressed before merging.

nemesifier:
@coderabbitai full review or resolve if no outstanding issues

nemesifier:
@coderabbitai resolve

nemesifier (dismissed):

nemesifier (approved):

nemesifier (approved):
Thanks @pandafy 👏

nemesifier (approved):

PR #735, opened 2026-06-24
[fix] Record monitoring metrics for sessions closed via bulk updates

### LINKED ISSUE #734: [bug] NAS reboot path bypasses post_save signal for stale accounting sessions
**Describe the bug**

When a NAS sends an Accounting-On packet, `AbstractRadiusAccounting._close_stale_sessions_on_nas_boot` closes stale sessions using a queryset `.update()`:

https://github.com/openwisp/openwisp-radius/blob/HEAD/openwisp_radius/base/models.py#L586-L588

Django's `.update()` bypasses model `save()` and therefore does not emit `post_save` signals. The `post_save_radiusaccounting` receiver (connected in `openwisp_radius/integrations/monitoring/apps.py:128`) is not triggered, so `tasks.post_save_radiusaccounting.delay()` is never dispatched. As a result, accounting metrics for sessions closed during a NAS reboot are not recorded in openwisp-monitoring.

Suggested fix: iterate over affected sessions and call `.save()` on each, or dispatch the monitoring task explicitly after the bulk update.

**Steps To Reproduce**

1. Create an active RADIUS accounting session.
2. Trigger NAS reboot handling by sending an Accounting-On packet from the NAS.
3. `_close_stale_sessions_on_nas_boot` updates the stale session(s) using `.update()` (sets `stop_time` and `terminate_cause`).
4. No `post_save` signal is emitted and the monitoring task is not queued.
5. Metrics for the closed session do not appear in openwisp-monitoring.

**Expected behavior**

When stale sessions are closed on NAS boot, the `post_save_radiusaccounting` task should be triggered so session metrics are recorded in openwisp-monitoring.

nemesifier:
## Checklist

Closes #734.

## Description of Changes

The RADIUS monitoring integration only writes traffic snapshots when Django's `post_save` signal is emitted. However, two common session closure paths use bulk updates which bypass this signal:

- NAS reboot handling (`Accounting-On`) closes stale sessions via `QuerySet.update()`.
- The `close_previous_radius_accounting_sessions` receiver uses `bulk_update()`.

As a result, traffic from sessions closed through these paths is present in the SQL accounting table but missing from the RADIUS traffic chart, leading to undercounted totals.

This change introduces a new `radius_accounting_closed` signal and a helper `emit_radius_accounting_closed()` that integrations can call after bulk updates. The monitoring integration now listens to both `post_save` and this new signal, ensuring every closed session is snapshotted.

Additionally, the monitoring metric timestamp is now aligned with the session's `stop_time` instead of the task execution time. This prevents traffic from being credited to the wrong time bucket when a session is closed or snapshotted later.

Regression tests are added to ensure these paths dispatch the monitoring task correctly and use the correct timestamp.

## Screenshot

Not applicable (no UI changes).

### DIFF
--- a/docs/developer/utils.rst
+++ b/docs/developer/utils.rst
@@ -27,6 +27,30 @@ completes successfully, just before the response is returned.
 The ``view`` argument can also be used to access the ``request`` object
 i.e. ``view.request``.
 
+``radius_accounting_closed``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Path**: ``openwisp_radius.signals.radius_accounting_closed``
+
+**Arguments**:
+
+- ``sender``: ``RadiusAccounting`` model class
+- ``instance``: closed ``RadiusAccounting`` instance
+
+This signal is emitted when a ``RadiusAccounting`` session is closed. For
+regular ``save()`` paths, it is emitted only when ``stop_time`` changes
+from ``None`` to any value, or when a closed session is created directly.
+Editing an already closed session does not emit this signal again.
+
+The signal is emitted after the database transaction is committed.
+
+Integrations which need to react to closed accounting sessions should
+listen to this signal.
+
+When closing multiple sessions with ``bulk_update()``, use
+``RadiusAccounting.emit_radius_accounting_closed()`` after the database
+update to emit this signal once for each closed session.
+
 .. _radius_captive_portal_mock_views:
 
 Captive portal mock views
--- a/docs/developer/utils.rst
+++ b/docs/developer/utils.rst
@@ -45,7 +45,10 @@ Editing an already closed session does not emit this signal again.
 The signal is emitted after the database transaction is committed.
 
 Integrations which need to react to closed accounting sessions should
-listen to this signal.
+listen to this signal. For example, the monitoring integration uses this
+signal to write RADIUS traffic snapshots for sessions closed by regular
+``post_save`` paths, ``Accounting-On`` packets and automatic stale-session
+cleanup.
 
 When closing multiple sessions with ``bulk_update()``, use
 ``RadiusAccounting.emit_radius_accounting_closed()`` after the database
--- a/openwisp_radius/base/models.py
+++ b/openwisp_radius/base/models.py
@@ -5,6 +5,7 @@ import os
 import string
 from datetime import timedelta
 from io import StringIO
+from typing import Iterable
 
 import django
 import phonenumbers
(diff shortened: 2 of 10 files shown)


docs/developer/utils.rst
```diff
@@ -27,6 +27,30 @@ completes successfully, just before the response is returned.
 The ``view`` argument can also be used to access the ``request`` object
 i.e. ``view.request``.
 
+``radius_accounting_closed``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
pandafy, line 31: Will it make more sense to always send this signal when the RadiusAccounting session is closed? It will help in cleaner implementation downstream (openwisp-monitoring integration). 

Right now, we will have to listen to two different signals, which is fine. but requires developers to be aware of the context.

openwisp_radius/base/models.py
```diff
@@ -583,9 +584,15 @@ def _close_stale_sessions_on_nas_boot(cls, called_station_id):
...
-        closed_count = stale_sessions.update(
-            stop_time=now(), terminate_cause="NAS-Reboot"
+        closed_sessions = list(stale_sessions)
+        stop_time = now()
+        for session in closed_sessions:
+            session.stop_time = stop_time
+            session.terminate_cause = "NAS-Reboot"
+        closed_count = cls.objects.bulk_update(
+            closed_sessions, fields=["stop_time", "terminate_cause"]
         )
```
pandafy: We typically use `iterator()` with `bulk_update` with a chuksize of 1000 (default in Django). Right now, the code will load all the objects in the memory which is not ideal. 

We can make the query efficient by only fetching the `id` in the queryset. We'll probably need to check that it does not increase the number of queries.

pandafy (changes requested):

openwisp_radius/receivers.py
```diff
@@ -126,6 +127,21 @@ def close_previous_radius_accounting_sessions(instance, created, **kwargs):
     RadiusAccounting.objects.bulk_update(
         closed_sessions, fields=["stop_time", "terminate_cause"]
     )
+    emit_radius_accounting_closed(closed_sessions)
+
+
+def emit_radius_accounting_closed_on_save(instance, created, *args, **kwargs):
```
nemesifier: This is too AI sloppy, I will improve tomorrow.

nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai full review

PR #757, opened 2026-07-28
[change!] Moved password reset to openwisp-users & prevent user enumeration #754

### LINKED ISSUE #754: [bug] Mark RADIUS SAML sessions as external for password expiration handling
⚠️ **core issue, maintainers only**.

**Describe the bug**

OpenWISP RADIUS has a deployed SAML integration based on `djangosaml2`. This integration predates allauth SAML support and is used by WiFi users authenticating through providers such as Azure AD, SPID, and university or municipal identity providers.

OpenWISP Users is changing password expiration handling so that it applies only to sessions authenticated with a local password. OAuth and SAML sessions must not be blocked because the user has an expired local password.

The RADIUS SAML integration does not use allauth's social-login flow, so it cannot be detected through allauth's `sociallogin` signal argument. Its successful login flow is handled by `djangosaml2`, which authenticates the user and calls Django `auth.login(request, user)` before the RADIUS post-login hook completes.

**Steps To Reproduce**

1. Enable password expiration in a deployment using the OpenWISP RADIUS SAML integration.
2. Use a user with a usable but expired local password.
3. Complete login through the RADIUS SAML integration.
4. Observe that the session can be treated as password-authenticated and redirected to the password-change page.

**Expected behavior**

- A user who completes login through the RADIUS SAML integration must have the Django session marked as externally authenticated.
- An expired local password must not block subsequent requests from that SAML-authenticated session.
- The existing RADIUS SAML login, organization provisioning, captive portal redirect, and token behavior must remain unchanged.
- Password-authenticated sessions must continue to be handled by OpenWISP Users' password expiration policy.

**Proposed implementation**

OpenWISP Users should expose a public helper for recording the completed authentication method in the Django session, for example:

```python
set_authentication_method(request, "external")
```

The RADIUS SAML post-login hook should call this helper after `djangosaml2` has successfully authenticated and logged in the user. The helper should use the same session marker used by OpenWISP Users for allauth OAuth and SAML logins.

This keeps the dependency direction correct: OpenWISP Users does not need to know about RADIUS, `djangosaml2`, or RADIUS-specific backend paths, while RADIUS explicitly identifies its SAML login as external authentication.

**Testing**

Add coverage for an existing user with a usable but expired local password completing the RADIUS SAML login flow. Verify that the resulting Django session contains the external authentication marker and that the password expiration middleware does not redirect the user to the password-change page.

Also verify that normal password login remains marked as password-authenticated and continues to trigger password expiration.

This issue should be implemented together with the related OpenWISP Users password-expiration issue: https://github.com/openwisp/openwisp-users/issues/511

**Screenshots**

Not applicable.

**System Informatioon:**

- OS: Linux
- Python Version: Python 3.12
- Django Version: Django 4.2 through 5.2 are supported by the project
- Browser and Browser Version (if applicable): Not applicable

pandafy:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).
- N/A I have updated the documentation.

Closes #754

## Description of Changes

- Password reset now always returns HTTP 200 to prevent user enumeration. Previously, unknown users returned HTTP 404.
- Password reset is now handled by openwisp-users. The local PasswordResetSerializer is deprecated and will be removed in 1.4.0.
- SAML sessions are now marked as externally authenticated, preventing PasswordExpirationMiddleware from redirecting SAML users to the password change page.

## Blockers 

- [ ] https://github.com/openwisp/openwisp-users/pull/544


nemesifier (commented):
Removed.

openwisp_radius/api/views.py
```diff
@@ -535,45 +544,52 @@ def post(self, request, *args, **kwargs):
...
+        Return active users matching ``identifier`` who belong to this organization.
+
+        Always returns a queryset and never reveals whether ``identifier`` exists
+        outside this organization. Both an unknown identifier and a user who is
+        not a member of this organization produce an empty queryset, allowing the
+        password reset endpoint to return the same success response in all cases
+        and preventing user enumeration.
+        """
+        app_label = User._meta.app_config.label
+        return (
```
nemesifier, line 575: Please preserve the existing superuser behavior here. The previous flow called validate_membership(), which allows superusers, but this relation filter excludes a superuser without an OrganizationUser record and silently sends no email. I pushed ace3c59c with a regression test. Include active superusers in this queryset and keep the membership filter for other users.

openwisp_radius/base/forms.py
```diff
@@ -52,28 +46,3 @@ def __init__(self, *args, **kwargs):
             help_text = f"Refer to the <b><u><a href='{docs_link}'>docs</a></u></b> \
                 for more details on importing users from a CSV"
             self.fields["csvfile"].help_text = help_text
-
-
-class PasswordResetForm(BasePasswordResetForm):
```
nemesifier, line 57: Please keep this import path as a deprecated alias or thin subclass of openwisp_users.base.forms.PasswordResetForm. Issue #511 requires preserving RADIUS extension points and import paths, but extensions importing or subclassing openwisp_radius.base.forms.PasswordResetForm will now fail immediately.

nemesifier (changes requested):
Please address the two compatibility regressions below. I pushed ace3c59c with a failing regression test for the superuser case.

pandafy:
@coderabbitai full review!

openwisp_radius/api/freeradius_views.py
```diff
@@ -435,7 +436,10 @@ def authenticate_user(self, request, user, password):
         return bool(
             getattr(request, "_mac_allowed", False)
             or (
-                not user.has_password_expired()
+                (
+                    not is_password_based_login(request, user=user)
```
nemesifier: I think this currently mixes the authentication method of two different credentials: the DRF user token and the RADIUS token.

For example, a user can log in through SAML, receive a DRF token and a RADIUS token, and then use the RADIUS token to access the network. That flow should continue to work even if the user's unused local password has expired.

However, the same user may later submit the actual local password. That password must still be rejected if it has expired. The latest SAML login must not make the local password valid again.

The current condition applies `password_based_token=False` to both checks:

```python
user.check_password(password)
or self.check_user_token(request, user, password)
```

As a result, after a SAML or social login sets the flag to `False`, an expired local password is accepted. `test_authorize_password_expired_with_saml_login` currently demonstrates this behavior because it submits the default local password, `tester`, rather than the RADIUS token.

I suggest separating the two credentials:

- If `user.check_password(password)` succeeds, always enforce the user's password-expiration policy.
- If the supplied value is a RADIUS token, enforce expiration according to how that specific RADIUS token was issued.
- Store this information on `RadiusToken`, or in an equivalent RADIUS-owned place, because it describes that credential rather than the user in general.

The token-issuing views should use the higher-level API provided by OpenWISP Users instead of calling `record_password_based_token(user, True/False)` directly. A possible flow would be:

```python
token = issue_auth_token(request, user, renew=renew)
password_based = is_password_based_token(token)
radius_token = get_or_create_radius_token(
    user,
    organization,
    password_based=password_based,
)
```

The exact API can differ, but the important part is that OpenWISP Users decides how its token was obtained, while OpenWISP RADIUS stores and enforces the authentication method of its own token.

Please update the test so it covers both cases explicitly: the expired local password must be rejected, while a RADIUS token issued after SAML or social authentication remains valid.

setup.py
```diff
@@ -42,7 +42,8 @@
         ),
         (
             "openwisp-users @"
-            "https://github.com/openwisp/openwisp-users/archive/refs/heads/1.3.tar.gz"
+            "https://github.com/openwisp/openwisp-users/archive/refs/heads/"
+            "issues/511-passowrd-expiration-sso.tar.gz"
```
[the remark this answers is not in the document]
nemesifier: Stale after the rebase. The temporary `issues/511-passowrd-expiration-sso` archive is no longer used; this dependency now uses the project-standard 1.3 compatibility branch URL.

openwisp_radius/tests/test_api/test_rest_token.py
```diff
@@ -303,6 +303,24 @@ def test_user_auth_token_password_expired(self):
         response = self._user_auth_token_helper("tester")
         self.assertEqual(response.data["password_expired"], True)
 
+    @mock.patch("openwisp_users.settings.USER_PASSWORD_EXPIRATION", 30)
+    def test_password_login_triggers_password_expiration(self):
```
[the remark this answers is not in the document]
nemesifier: Stale after the rebase. `test_password_login_triggers_password_expiration` is absent, and `test_user_auth_token_password_expired` is the sole endpoint-level password-expiration test.

nemesifier:
@coderabbitai full review

nemesifier:
@coderabbitai any outstanding issue?

nemesifier:
@coderabbitai review

nemesifier:
@coderabbitai resolve

nemesifier (approved):

PR #772, opened 2026-08-19
[feature] Updated monitoring integration to support new timeseries DBs

### LINKED ISSUE #769: [feature] Add InfluxDB 2 and Elasticsearch support to RADIUS monitoring
Depends on https://github.com/openwisp/openwisp-monitoring/issues/848.

`openwisp_radius.integrations.monitoring` supports only InfluxDB 1. Its six chart configurations use InfluxQL, and `rebuild_radius_accounting_metrics` issues a raw InfluxQL deletion query.

As a result, enabling the RADIUS monitoring integration with InfluxDB 2 or Elasticsearch prevents RADIUS registration, traffic, and unique-session charts from working.

Add InfluxDB 2 and Elasticsearch query definitions for all RADIUS monitoring charts while preserving their current behavior:

- User registrations, grouped by registration method.
- Total registered users, using the latest cumulative snapshot per time bucket.
- Device-specific and general RADIUS traffic, with upload and download values in GB.
- Device-specific and general unique RADIUS sessions, grouped by registration method.

Use the Monitoring time-series client API for rebuilding accounting metrics so the management command works with every supported backend.

Update the monitoring integration test configuration and CI to run the tagged RADIUS monitoring suite against InfluxDB 1, InfluxDB 2, and Elasticsearch.

**Describe alternatives you've considered**

Keeping the integration InfluxDB 1-only leaves RADIUS incompatible with supported Monitoring backends.

Adding backend-specific client calls throughout RADIUS would duplicate storage-layer behavior that belongs in OpenWISP Monitoring.

**Additional context**

InfluxDB 2 support was added to OpenWISP Monitoring in https://github.com/openwisp/openwisp-monitoring/pull/801.

Elasticsearch support was added to OpenWISP Monitoring in https://github.com/openwisp/openwisp-monitoring/pull/829.

pushpitkamboj:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).

Closes #769.

Elasticsearch support is added in a follow-up commit on this branch, once
openwisp/openwisp-monitoring#829 provides the missing client logic.

## Description of Changes

The RADIUS monitoring integration only defined InfluxQL queries, so its charts
did not work on the other timeseries backends supported by OpenWISP Monitoring.

- Added the InfluxDB 2 (Flux) chart and summary queries of the six RADIUS
  charts. A summary query is needed because, unlike InfluxDB 1, the influxdb2
  backend cannot derive the summary from the chart query.
- Moved the deletion performed by `rebuild_radius_accounting_metrics` from a raw
  InfluxQL query to `timeseries_db.delete_metric_data()`, so that the command
  works on every backend.
- The monitoring integration tests can now be run on any supported backend with
  `TIMESERIES_BACKEND`; CI runs them on both InfluxDB 1 and InfluxDB 2.

This PR requires openwisp/openwisp-monitoring#829, which adds what is missing
on the client side:

- `delete_metric_data()` accepts a `timestamp`, to delete a single point;
- the influxdb2 backend recognizes `last()` as an aggregate function, which the
  "Total Registered Users" chart relies on;

`requirements-test.txt` points to that branch in the meantime and has to be
changed back to `1.3` after it is merged.


.github/workflows/ci.yml
```diff
@@ -11,25 +11,33 @@ on:
 
 jobs:
   build:
-    name: Python==${{ matrix.python-version }} | ${{ matrix.django-version }}
+    name: >-
+      ${{ matrix.tsdb }} | Python==${{ matrix.python-version }} |
+      ${{ matrix.django-version }}
```
nemesifier, line 19: This workflow still inherits the repository's `write` default for `GITHUB_TOKEN`, although these jobs only need read access. The expanded matrix therefore runs third-party actions with more authority than necessary.

Please set workflow-level `permissions: contents: read` and grant any additional scope only to the specific job that requires it.

Severity: P2

.github/workflows/ci.yml
```diff
@@ -11,25 +11,33 @@ on:
...
         exclude:
           # Python 3.13 supported only in Django >=5.1.3
           - python-version: "3.13"
             django-version: django~=4.2.0
+        # The other timeseries backends are tested only on the latest Python
+        # and Django versions, and only on the monitoring integration.
+        include:
+          - python-version: "3.13"
+            django-version: django~=5.2.0
+            tsdb: influxdb2
```
nemesifier: This matrix adds only InfluxDB 2, but the PR declares `Closes #769`, which also requires Elasticsearch query definitions and integration coverage. The current settings reject `TIMESERIES_BACKEND=elasticsearch`, so merging this version would close the issue while leaving one of its supported backends broken.

Please either complete the Elasticsearch configuration, queries, documentation, and CI coverage before merging, or narrow this PR and keep the remaining work tracked by an open linked issue instead of closing `#769`.

Severity: P2

.github/workflows/pypi.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.10"
+          python-version: "3.11"
```
nemesifier: This change seems unrelated to this PR. Is it necessary? If not, remove.

Severity: P2

openwisp_radius/integrations/monitoring/configuration.py
```diff
@@ -5,6 +5,100 @@
...
+        return query + _flux_summary_time
+    query += _flux_method_group + (
+        " |> aggregateWindow(every: {window}, fn: " + function + ', timeSrc: "_start"'
+        "{window_timezone})"
+    )
+    if function == "last":
+        # Flux does not provide the linear fill of InfluxDB 1
+        return (
+            query
+            + ' |> fill(column: "_value", usePrevious: true)'
```
nemesifier, line 60: This `fill()` does not fill missing registration buckets. Flux selector functions such as `last()` remove empty buckets, so there is no value here for `fill(usePrevious: true)` to use. If the hourly task misses one run, the chart has a gap instead of showing the most recent total.

Please change the query so an empty bucket is returned and shows the most recent total rather than disappearing. Add an InfluxDB 2 regression test with a missing middle bucket that checks the timestamps and values.

Severity: P2

openwisp_radius/integrations/monitoring/configuration.py
```diff
@@ -5,6 +5,100 @@
 
 from openwisp_radius.registration import REGISTRATION_METHOD_CHOICES
 
+_flux_range = (
+    'import "date"\n{timezone_import}from(bucket: "{bucket}")'
+    " |> range(start: {time_start}{end_range})"
```
nemesifier, line 10: Flux `range()` excludes its `stop`, while the existing InfluxQL queries include points matching the requested end date with `time <= end_date`. Every new Flux chart and summary can therefore omit a point exactly on the selected upper boundary.

Please make the Flux queries include a point whose timestamp is exactly the requested end date, as the existing InfluxQL queries do. Add a regression test for that boundary case.

Severity: P3

nemesifier (changes requested):
Please rebase on the latest master, see my comments below

openwisp_radius/integrations/monitoring/configuration.py
pushpitkamboj: fixed in the influxdb2 client in openwisp/openwisp-monitoring#829, where `{end_range}` is generated: the range stop now includes the end date. Added a chart test here that fails without it.

openwisp_radius/integrations/monitoring/configuration.py
pushpitkamboj: I made `createEmpty: true` explicit so it can't regress silently, and added the regression test: it fails with None if the fill is removed, and with a missing bucket if createEmpty is false.

nemesifier:
@coderabbitai full review and check AGENTS.md rule adherence.

openwisp_radius/integrations/monitoring/configuration.py
```diff
@@ -5,6 +5,100 @@
...
+    query += _flux_method_group + (
+        " |> aggregateWindow(every: {window}, fn: " + function + ', timeSrc: "_start"'
+        ", createEmpty: true{window_timezone})"
+    )
+    if function == "last":
+        # Flux does not provide the linear fill of InfluxDB 1
+        return (
+            query
+            + ' |> fill(column: "_value", usePrevious: true)'
+            + _flux_method_projection
```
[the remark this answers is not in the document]
pushpitkamboj: Checked on InfluxDB 2.9.1. `aggregateWindow(fn: last, createEmpty: true)` does materialise the empty window as a null row (removing the `fill()` leaves the bucket present with a null value instead of dropping it), so `fill(usePrevious: true)` fills it on all three, the docs limitation applies to the `window() |> last() form` which this query doesn't use.

openwisp_radius/integrations/monitoring/management/commands/test_rebuild_radius_accounting_metrics.py
```diff
@@ -120,6 +127,46 @@ def test_rebuild_radius_accounting_metrics_commit(
             1,
         )
 
+    @patch("logging.Logger.warning")
+    def test_rebuild_radius_accounting_metrics_deletes_only_session_point(self, *args):
+        from ...tasks import post_save_radiusaccounting
```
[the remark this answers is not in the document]
pushpitkamboj: Can't move it up unfortunately as tasks.py calls `load_model()` at import time, and the monitoring apps only get installed when `MONITORING_INTEGRATION` is set, so a module level import makes test discovery blow up with `ImproperlyConfigured: Could not find monitoring.Metric` on all the jobs that aren't testing the integration (I tried it, same failure as 37eff1c).
But I found a nicer way that drops the local import altogether: I only needed a metric for an already accounted session, and the `post_save` receiver already writes that at `stop_time`. So the test just creates the session normally now and lets the signal handle it. o tasks import anywhere, and it covers the receiver too instead of copying its kwargs by hand.

PR #773, opened 2026-08-20
[feature] Add REST API list, detail, and delete endpoints for RADIUS batch user creation #771

### LINKED ISSUE #771: [feature] Add REST API endpoints for reviewing RADIUS batch user creation

The RADIUS batch REST API currently supports only creation with `POST /api/v1/radius/batch/`.

After a batch has been created, API clients cannot list batch operations, retrieve an operation's status and metadata, or delete an operation. This is especially limiting for asynchronously processed batches, whose completion status cannot be retrieved through the REST API.

Protected downloads already exist for prefix-generated credentials and uploaded CSV files, but clients cannot discover these resources through a batch list or detail API.

Add read and deletion endpoints while preserving the existing batch-creation API and response behavior:

- Keep `POST /api/v1/radius/batch/` unchanged.
- Keep `user_credentials` in the synchronous `201 Created` response for prefix-generated batches.
- Keep asynchronous creation responses unchanged: they return `202 Accepted`, do not return credentials, and expose the PDF link only after completion.
- Add `GET /api/v1/radius/batch/` to list batches belonging to organizations managed by the requesting user.
- Add `GET /api/v1/radius/batch/<uuid>/` to retrieve a batch.
- Add `DELETE /api/v1/radius/batch/<uuid>/` to delete a batch and its associated users, consistently with the existing `RadiusBatch.delete()` behavior.
- Require the appropriate Django model permissions and enforce organization isolation for every operation.
- Do not allow batch editing through these endpoints.
- Update `docs/user/rest-api.rst` to document the collection, detail, filtering, deletion, and protected-download behavior.

The collection endpoint must use pagination and support filters consistent with the Django admin batch changelist:

- Filter by organization.
- Filter by strategy.
- Search by batch name.

The list and detail representations must not expose `user_credentials`. The synchronous creation response remains the only JSON response which includes plaintext generated credentials.

Instead, list and detail responses must expose links to the existing protected download endpoints:

- Completed prefix batches expose `pdf_link`, pointing to the protected credentials PDF endpoint.
- CSV batches expose a CSV download link when a CSV file is available.

The read representation should expose safe review metadata, including the ID, organization, name, strategy, status, group, expiration date, notes, timestamps, and associated users without plaintext credentials or password hashes.

Deletion must be rejected while `status` is `processing`, matching the existing individual admin deletion behavior. The API must return a clear error message explaining that a batch cannot be deleted while it is being processed. This restriction must also apply to the custom admin bulk action, which currently bypasses the status check. Pending, completed, and failed batches remain deletable.

**Describe alternatives you've considered**

- Returning `user_credentials` from batch list and detail responses: rejected because it would repeatedly expose plaintext credentials and make accidental client-side logging or persistence more likely.
- Creating new password-download storage or endpoints: unnecessary because the protected PDF endpoint already provides this capability for prefix-generated batches.
- Leaving batch review limited to the Django admin: unsuitable for API clients and does not provide a REST status lookup for asynchronous jobs.

**Additional context**

Issue #102 intentionally stores generated credentials so the protected PDF can be generated on demand. This change should reuse that mechanism rather than introduce a second credential-delivery flow.

Issue #608 identified that REST API behavior must account for asynchronous batch processing, including preventing deletion while a batch is in progress. The asynchronous lifecycle was implemented, but batch list, detail, and deletion endpoints were not added.

BHARATH0153:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://github.com/openwisp/openwisp-radius/blob/master/CONTRIBUTING.md) and [OpenWISP Anti AI Spam Policy](https://github.com/openwisp/openwisp-radius/blob/master/anti-ai-spam-policy.md).

Closes #771

## Description of Changes

- Added GET endpoint to list batch user creation operations with filtering (organization, strategy), search (by name), and pagination.
- Added GET detail endpoint (`/api/v1/radius/batch/<uuid>/`) to retrieve a single batch by UUID.
- Added DELETE endpoint to remove a batch and its associated users. Returns 409 Conflict if the batch status is `processing`.
- Added `RadiusBatchReadSerializer` that excludes `user_credentials` from list/detail responses to avoid repeatedly exposing plaintext credentials.
- Added `RadiusBatchFilter` for filtering by organization and strategy.
- Updated admin `delete_selected_batches` action to skip batches with `processing` status with a warning message.
- Added 15 new tests covering list, detail, delete, permissions, filtering, search, cross-org access, and processing guard.
- Updated REST API documentation with GET/DELETE docs, filters table, batch detail section, and cross-references to importing/generating users docs.


openwisp_radius/api/views.py
```diff
@@ -196,6 +229,57 @@ def get(self, request, *args, **kwargs):
...
+    name="delete",
+    decorator=swagger_auto_schema(
+        operation_description="""
+        Deletes a batch user creation operation and its associated users.
+        Cannot delete a batch while it is being processed.
+        """,
+        responses={204: "No Content", 409: "Conflict"},
+    ),
+)
+class BatchDeleteView(
```
nemesifier: Why is this not bundled in BatchDetailView?

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,308 @@
...
+        self._get_admin()
+        staff = self._create_staff_user("deletestaff", org=self.default_org)
+        delete_perm = Permission.objects.get(codename="delete_radiusbatch")
+        with self.subTest("w/o login"):
+            batch = self._create_prefix_batch(name="batch-noauth")
+            response = self.client.delete(
+                reverse("radius:batch_delete", args=[batch.pk])
+            )
+            self.assertEqual(response.status_code, 401)
+        with self.subTest("superuser"):
```
nemesifier, line 408: Please read `AGENTS.md` rules about `subTest`.

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,308 @@
+import swapper
+from django.contrib.auth import get_user_model
+from django.contrib.auth.models import Permission
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.urls import reverse
+from rest_framework import status
```
nemesifier, line 7: Rename this file `test_batch.py`

openwisp_radius/tests/test_api/test_api_batch.py
```diff
@@ -0,0 +1,308 @@
...
+
+from ...utils import load_model
+from ..mixins import ApiTokenMixin, BaseTestCase
+
+User = get_user_model()
+RadiusBatch = load_model("RadiusBatch")
+OrganizationUser = swapper.load_model("openwisp_users", "OrganizationUser")
+
+
+class TestBatchListDetailDelete(ApiTokenMixin, BaseTestCase):
```
nemesifier: There's tests relating to the batch API in `test_api.py`, please move those here.

Rename this class for consistency:

```suggestion
class TestBatch(ApiTokenMixin, BaseTestCase):
```

openwisp_radius/api/views.py
```diff
@@ -196,6 +229,57 @@ def get(self, request, *args, **kwargs):
...
+    ProtectedAPIMixin, FilterByOrganizationManaged, RetrieveDestroyAPIView
+):
+    authentication_classes = (BearerAuthentication, SessionAuthentication)
+    permission_classes = (IsAdminUser, DjangoModelPermissions)
+    queryset = RadiusBatch.objects.all()
+    serializer_class = RadiusBatchReadSerializer
+
+    def delete(self, request, *args, **kwargs):
+        with transaction.atomic():
```
nemesifier: I suspect this could be done in a cleaner way, without overriding and rewriting the delete method. There should be a method to get the object, we can override that method and call `obj = super().get(...)` then check if the status is processing before returning, in which case we can raise a [DRF exception](https://www.django-rest-framework.org/api-guide/exceptions/#apiexception), probably need to define a new exception for conflict HTTP status because I don't think DRF has a built-in one (please double check).

The advantage is that we avoid redefining a standard DRF method, which increases maintainability.

nemesifier (changes requested):
The new test class is not being added to the tests of the sample app.

nemesifier:
@coderabbitai review AGENTS.md rules adherence.

BHARATH0153:
@coderabbitai now review AGENTS.md rules adherence.

openwisp_radius/api/views.py
```diff
@@ -257,27 +253,28 @@ class BatchDetailView(ProtectedAPIMixin, FilterByOrganizationManaged, RetrieveAP
...
-                    {"detail": _("Cannot delete a batch while it is being processed.")},
-                    status=status.HTTP_409_CONFLICT,
-                )
-            batch.delete()
-        return Response(status=status.HTTP_204_NO_CONTENT)
+    def get_object(self):
+        obj = super().get_object()
+        if self.request.method == "DELETE":
+            with transaction.atomic():
+                batch = RadiusBatch.objects.select_for_update().get(pk=obj.pk)
```
nemesifier: why are you fetching again the obj if we already got it in `obj = super().get_object()`?

openwisp_radius/api/views.py
```diff
@@ -257,27 +253,28 @@ class BatchDetailView(ProtectedAPIMixin, FilterByOrganizationManaged, RetrieveAP
...
+                if batch.status == RadiusBatch.PROCESSING:
+                    raise Conflict(
+                        _("Cannot delete a batch while it is being processed.")
```
nemesifier: Let's write:

"The radius batch object is currently being processed and cannot be deleted."

openwisp_radius/api/views.py
```diff
@@ -257,27 +253,28 @@ class BatchDetailView(ProtectedAPIMixin, FilterByOrganizationManaged, RetrieveAP
...
-                return Response(
-                    {"detail": _("Cannot delete a batch while it is being processed.")},
-                    status=status.HTTP_409_CONFLICT,
-                )
-            batch.delete()
-        return Response(status=status.HTTP_204_NO_CONTENT)
+    def get_object(self):
+        obj = super().get_object()
+        if self.request.method == "DELETE":
+            with transaction.atomic():
```
nemesifier: why the transaction? It's not immediately obvious.

openwisp_radius/api/serializers.py
```diff
@@ -562,6 +562,70 @@ class Meta:
...
+class RadiusBatchReadSerializer(serializers.ModelSerializer):
+    organization = serializers.PrimaryKeyRelatedField(read_only=True)
+    users = BatchUserSerializer(many=True, read_only=True)
+    pdf_link = serializers.SerializerMethodField(required=False, read_only=True)
+    csv_link = serializers.SerializerMethodField(required=False, read_only=True)
+    status = serializers.CharField(read_only=True)
+
+    def get_pdf_link(self, obj):
+        if (
+            isinstance(obj, RadiusBatch)
```
nemesifier: why, what else could it be?

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,311 @@
...
+            self.assertEqual(response.status_code, 401)
+        with self.subTest("superuser"):
+            batch = self._create_prefix_batch(name="batch-super")
+            header = self._get_auth_header()
+            response = self.client.delete(
+                reverse("radius:batch_detail", args=[batch.pk]),
+                HTTP_AUTHORIZATION=header,
+            )
+            self.assertEqual(response.status_code, 204)
+        with self.subTest("staff w/o delete permission"):
```
nemesifier: Leave one blank line before each `subTest`, applies to all other occurrences, except when `subTest` is inside a loop.

openwisp_radius/api/urls.py
```diff
@@ -83,6 +83,11 @@ def get_api_urls(api_views=None):
                 name="update_registered_user_registration_method",
             ),
             path("radius/batch/", api_views.batch, name="batch"),
+            path(
+                "radius/batch/<uuid:pk>/",
+                api_views.batch_detail,
+                name="batch_detail",
```
nemesifier:
```suggestion
                name="radius_batch_detail",
```

BHARATH0153:
@nemesifier once look into this please thanks!

nemesifier:
@coderabbitai full review

openwisp_radius/api/views.py
```diff
@@ -196,6 +233,56 @@ def get(self, request, *args, **kwargs):
...
+    authentication_classes = (BearerAuthentication, SessionAuthentication)
+    permission_classes = (IsAdminUser, DjangoModelPermissions)
+    queryset = RadiusBatch.objects.select_related("organization").prefetch_related(
+        "users"
+    )
+    serializer_class = RadiusBatchReadSerializer
+
+    def perform_destroy(self, instance):
+        with transaction.atomic():
+            batch = RadiusBatch.objects.select_for_update().get(pk=instance.pk)
```
nemesifier: isn't "instance" already the object we need to delete? Why the `transaction` and the `select_for_update`?

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,323 @@
...
+            name=name,
+            strategy="prefix",
+            prefix="test",
+            organization=organization,
+            status=RadiusBatch.COMPLETED,
+        )
+        batch.save()
+        return batch
+
+    def _create_staff_user(self, username="staffuser", org=None):
```
nemesifier: Why don't you use utility methods like `_create_operator` or `_create_administrator`?
There's a bunch of helpers like these in openwisp-users which are reused across openwisp, please use them rather than definining new ones.

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,323 @@
...
+class TestBatch(ApiTokenMixin, BaseTestCase):
+    def _get_auth_header(self, username="admin", password="tester"):
+        if username == "admin":
+            self._get_admin()
+        login_payload = {"username": username, "password": password}
+        login_url = reverse("radius:user_auth_token", args=[self.default_org.slug])
+        login_response = self.client.post(login_url, data=login_payload)
+        return f"Bearer {login_response.json()['key']}"
+
+    def _create_prefix_batch(self, name="test-prefix-batch", organization=None):
```
nemesifier: There's already a `_create_radius_batch`.

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,323 @@
...
+        header = self._get_auth_header()
+        url = reverse("radius:batch")
+        with self.subTest("filter prefix"):
+            response = self.client.get(
+                url, {"strategy": "prefix"}, HTTP_AUTHORIZATION=header
+            )
+            names = [b["name"] for b in response.json()["results"]]
+            self.assertIn("prefix-batch", names)
+            self.assertNotIn("csv-batch", names)
+        with self.subTest("filter csv"):
```
nemesifier, line 103: Leave a blank line when subTest are chained this way, it's in the `AGENTS.md` but somehow @coderabbitai is missing this..

nemesifier (changes requested):

openwisp_radius/tests/test_api/test_batch.py
```diff
@@ -0,0 +1,447 @@
...
+
+from ...utils import load_model
+from ..mixins import ApiTokenMixin, BaseTestCase
+
+User = get_user_model()
+RadiusBatch = load_model("RadiusBatch")
+
+
+class TestBatch(ApiTokenMixin, BaseTestCase):
+    def _get_auth_header(self, username="admin", password="tester"):
```
nemesifier: Why is this helper method needed? Isn't `ApiTokenMixin` already providing everything we need?

Why can't you verify these things on your own before sending bad code?

nemesifier (changes requested):

nemesifier:
@coderabbitai resume

openwisp_radius/admin.py
```diff
@@ -463,8 +464,7 @@ def save_model(self, request, obj, form, change):
         obj.schedule_processing(number_of_users=num_users)
 
     def delete_model(self, request, obj):
-        obj.users.all().delete()
-        super(RadiusBatchAdmin, self).delete_model(request, obj)
+        obj.delete_if_not_processing()
```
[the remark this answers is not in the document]
nemesifier: The POST permission check calls `can_delete()` before Django reaches `log_deletions()`, so a processing batch receives `PermissionDenied` and no deletion log is created. The proposed override would change the established permission behavior, so no change is needed.

nemesifier:
@coderabbitai are there any outstanding issues?

nemesifier:
@coderabbitai are there any outstanding issues?

nemesifier:
@coderabbitai resolve

tests/openwisp2/integrations/tests.py
```diff
@@ -29,6 +29,19 @@ def test_swagger_api_docs(self):
         self.client.force_login(admin)
         response = self.client.get(reverse("schema-swagger-ui"), {"format": "openapi"})
         self.assertEqual(response.status_code, 200)
+        schema = response.json()
```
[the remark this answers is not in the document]
nemesifier: Validated against the current sample project: the focused test passes and `response.json()` returns the OpenAPI document. The `format=openapi` parameter selects the schema renderer, so this endpoint is correct.

tests/openwisp2/integrations/tests.py
```diff
@@ -29,6 +29,19 @@ def test_swagger_api_docs(self):
...
+        response_schema = schema["paths"]["/radius/batch/{id}/"]["get"]["responses"][
```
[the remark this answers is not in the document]
nemesifier: Validated against the generated schema: the focused test passes with `/radius/batch/{id}/`, so the existing path lookup is correct.

openwisp_radius/api/swagger.py
```diff
@@ -60,3 +62,18 @@ class RegisterResponse(serializers.Serializer):
         read_only=True,
         help_text=("Only send when `REST_USE_JWT` is `True`."),
     )
+
+
+class BatchUserPaginationResponse(serializers.Serializer):
```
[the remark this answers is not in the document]
nemesifier: Validated against the generated schema: the focused test passes and emits `#/definitions/BatchUserPaginationResponse`. An explicit `ref_name` is not needed.

nemesifier:
@coderabbitai any outstanding issue?

nemesifier (approved):

